PR #32, opened 2022-03-18
Add AO Rendering Material and Example 

### LINKED ISSUE #28: Add "AO Mode" example that performs hemisphere sampling
See section 2.2 here:

http://kth.diva-portal.org/smash/get/diva2:1574351/FULLTEXT01.pdf

**Plan**
- [x] Perform randomized, accumulated hemisphere sampling with normals
- [x] Create a demo that uses a nice model and renders a "clay" type AO model
- [x] Not screen space -- must operate on the surface of the geometry and accumulate in the scene
  - render to float target
  - generalize the "accumulation" rendering?
  - render to screen with sRGB transform
- [x] ~Fix hemisphere sampling bias~
  - bias in the model
  - merge vertices and smooth
- [ ] Support normal maps?

### LINKED ISSUE #29: Add support for non indexed geometry to "mergeMeshes"
Just create an index if there isn't one.

gkjohnson:
Fix #28 
Fix #29 
Related to https://github.com/mrdoob/three.js/issues/14055

Add support for a raster material that renders an accumulated ambient occlusion effect on the mesh. In the future it can be repurposed to render to a UV map and saved / used as an AO map.

**TODO**
- [x] Add support for normal maps (with tangents only)
- [x] Comments
- [x] Lower resolution
- [x] Test normal maps
- [ ] Look into the slowness when using multiple materials?

**AO Rendering**

_radius = 2.0_

_radius = 0.75_

_radius = 0.25_

**With Normal Maps**

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,12 @@
+# These are supported funding model platforms
+
+github: gkjohnson
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+otechie: # Replace with a single Otechie username
+custom: # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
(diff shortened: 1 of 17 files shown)


gkjohnson:
cc @bhouston. I'd started working on this last week -- here's the PR for "AO mode" that includes options for "radius" and sample count. The primarily reusable bit a material that rasterizes the geometry and performs raytracing to determine the darkness of the pixel. It includes support tangent space normal maps, as well, but requires that tangents be provided as geometry attributes. Once this is merged I can send over a demo.

The example demonstrates how to render and accumulate samples into a buffer. To render an AO Map you really just need to use the UV coordinates as a vertex transformations instead of the positions. I made #33 to discuss what an API for generating these maps might look like.

This should provide a good start for someone to start looking into AO map generation. I think that should cover the features needed for AO at the moment but we can improve over time if needed.

The last thing I noticed is that when creating multiple materials to render with (ie different materials for multiple meshes) it causes the framerate to drop. I'll have to test a bit more and probably ask around on twitter to track down what exactly might be happening there.

gkjohnson:
**Live Demo**

https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/aoRender.html

Issue #10, opened 2022-03-02
Port CPU Path Tracer Implementation to Shader


gkjohnson:
Include support for:
- Metalness
- GGX specular / glossiness

Optional:
- Area Lights
- Multiple Importance Sampling
- Transmission

Notes:
- Just stub area lights
- need to compute probability of hitting a pixel on the env map (more complicated with PMREM)
- Just avoid MIS for first pass
- Simplify shader by removing ability for a separate gradient background for now.
  - Possibly add a separate "background" env map (can make gradient from that)
  - For translucent object we need to track a direct / refracted ray vs a bounced ray.

Files:
- Structs
- BSDF Sample Functions (PDF and Ray sample)
- Lighting sample functions
- Path tracing loop?
- NOTE: Export so others can implement

gkjohnson:
Closing in favor #46, #47

PR #38, opened 2022-03-23
Add more surface properties

gkjohnson:
Related to #10

**Next**
- [x] port all functions over to glsl
- [x] add structs for material sampling results
- [x] make sure functions compile
- [x] Add emissive gui property
- [x] integrate into path tracer
  - possibly just need `bsdfSample` function?

**TODO**
- ~double check ggx functions. Hard edges at side.~
- [x] ensure 100% metalness works as expected / looks correct.
- [x] understand incorrect reflection on right side of model
- [x] fix direction discard when on incorrect side of surface
- [x] add support for filtered roughness.
- [x] add dropdown for floor material
- [ ] ensure filtered roughness factor works (might have to wait until env map CDF support)
- [ ] fix transmission
- [ ] alpha map cutout


gkjohnson:
image

Issue #33, opened 2022-03-18
Add utility class for generating AO Map


gkjohnson:
Something like the following:

```js
const uvGenerator = new UVGenerator();
uvGenerator.generate( targetMeshes, { name: 'uv2' } );

const generator = new AOMapGenerator();
generator.samples = 1000;

const targetMeshes = [ ... ];
const texture = await generator.generate( renderer, targetMeshes, aoContext, { name: 'uv2' } );
```

donmccurdy:
/cc @robertlong – Hi! Hope it's OK to ask a couple questions here. I'm wondering about using xatlas-web for this. I think we'd ideally need two workflows supported:

- (A) 1 mesh - unwrap and pack
    - load 1 mesh, providing `position` and `normal` attributes
    - generate a new UV atlas
    - use `xref` vertex IDs to rewrite the original mesh since some vertices have been duplicated. Original mesh may have arbitrary other attributes not used in the atlasing process.

- (B) N meshes - repack only
    - load N meshes that already have UVs, providing `position`, `normal`, and `uv` attributes
    - generate a shared atlas, packing all UVs into the same [0,1] coordinate space
    - use `xref` vertex IDs to rewrite all original meshes, as above

At a glance I'm having trouble seeing if these workflows are supported by xatlas itself, the xatlas-web bindings, or neither... is that something you happen to know? 

Related: https://github.com/donmccurdy/glTF-Transform/issues/534

hybridherbst:
A nice workflow would also be (kind of an extension of B):
- C)
  - load N meshes that may or may not have UVs
  - generate UV2 if UV1 is overlapping or doesn't exist
  - atlas all UV2s into 1-M AO maps (depends on target texel count)
  - generate AO map(s) in UV2
  - write out meshes with new UV2 + AO, no need to repack existing textures

This is generally what game engines (Unity/Unreal at least) do for lightmaps.

gkjohnson:
@donmccurdy Looks like there's a new xatlas wrapper being developed -- maybe it alleviates some of your concerns?

https://twitter.com/repalash/status/1521165176086487040

donmccurdy:
Thanks to @toji's work on https://github.com/toji/watlas and https://github.com/donmccurdy/glTF-Transform/pull/1660, glTF Transform v4.2 includes official UV unwrapping features:

- https://gltf-transform.dev/modules/functions/functions/unwrap
[...]

PR #42, opened 2022-03-29
Example: Drag and drop glTF files

donmccurdy:
Totally unsure if this PR is something you actually want in the example, but was a fun way to try out `three-gpu-pathtracer` and let me kick the tires on [`@gltf-transform/view`](https://github.com/donmccurdy/glTF-Transform-View/) at the same time. ~~Adds support in the AO example for drag-and-drop of GLB files, and has an export button (working, minus the AO texture itself) as well.~~ Adds an example of drag-and-drop viewing using the path tracer.

~Related:~

- ~#33~

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,10 +29,10 @@
     "three-mesh-bvh": "^0.5.10"
   },
   "devDependencies": {
-    "@gltf-transform/core": "^2.0.5",
-    "@gltf-transform/extensions": "^2.0.5",
-    "@gltf-transform/functions": "^2.0.5",
-    "@gltf-transform/view": "^0.3.0",
+    "@gltf-transform/core": "^2.1.5",
+    "@gltf-transform/extensions": "^2.1.5",
+    "@gltf-transform/functions": "^2.1.5",
+    "@gltf-transform/view": "^0.4.4",
     "@placemarkio/flat-drop-files": "^1.0.1",
     "concurrently": "^6.3.0",
     "eslint": "^7.32.0",
@@ -62,5 +62,8 @@
   "bugs": {
     "url": "https://github.com/gkjohnson/three-gpu-pathtracer/issues"
   },
-  "homepage": "https://github.com/gkjohnson/three-gpu-pathtracer#readme"
+  "homepage": "https://github.com/gkjohnson/three-gpu-pathtracer#readme",
+  "alias": {
+    "buffer": false
+  }
 }
--- a/package.json
+++ b/package.json
@@ -29,6 +29,11 @@
     "three-mesh-bvh": "^0.5.10"
   },
   "devDependencies": {
+    "@gltf-transform/core": "^2.0.5",
+    "@gltf-transform/extensions": "^2.0.5",
+    "@gltf-transform/functions": "^2.0.5",
+    "@gltf-transform/view": "^0.3.0",
+    "@placemarkio/flat-drop-files": "^1.0.1",
     "concurrently": "^6.3.0",
     "eslint": "^7.32.0",
     "eslint-config-mdcs": "^5.0.0",
--- a/example/dragDrop.html
+++ b/example/dragDrop.html
@@ -1,0 +1,54 @@
+<html>
+    <head>
+        <title>Drag-and-Drop</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+
+        <style>
+            html, body {
+                margin: 0;
+                padding: 0;
+				background-color: #111;
+            }
+
+			#info {
+				position: absolute;
+				bottom: 0;
+				left: 0;
+				font-family: 'Courier New', Courier, monospace;
+				color: white;
+				pointer-events: none;
+			}
+
+			#samples, #credits {
+
+				opacity: 0.5;
+				background-color: rgba( 0.0, 0.0, 0.0, 0.5 );
+				padding: 5px;
+				display: inline-block;
+
+			}
+
+			#loading {
+				position: absolute;
+				left: 50%;
+				top: 50%;
+				transform: translate(-50%, -50%);
+				color: white;
+				font-family: 'Courier New', Courier, monospace;
+			}
+        </style>
+
+    </head>
+    <body>
+		<div id="loading">DRAG AND DROP .GLB or .GLTF</div>
+		<div id="info">
+			<div>
+				<div id="samples">--</div>
+			</div>
+			<div>
+				<pre id="credits">--</pre>
+			</div>
+		</div>
+        <script src="./dragDrop.js" type="module"></script>
+    </body>
+</html>
--- a/example/dragDrop.js
+++ b/example/dragDrop.js
@@ -191,8 +191,6 @@ async function updateModel( modelDocument ) {
 	material.materials.updateFrom( materials, textures );
 	material.setDefine( 'MATERIAL_LENGTH', materials.length );
 
-	generator.dispose();
-
 	loadingEl.style.visibility = 'hidden';
 
 	animate();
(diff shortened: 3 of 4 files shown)


gkjohnson:
Thanks Don! This is great. I was thinking about adding a drag and drop demo at some point -- it should make testing easier and provide a pretty good template for an ao map generator in the future. How do you feel about making a separate demo for drag and drop, though?

Also I haven't pulled your branch or looked at it in too much detail, yet, but the image you posted doesn't look quite right. Maybe the tangents are incorrect somehow? Here's what the "Antique Camera" model looks like in the current page with MikkTSpace tangents generated when no tangents exist:

image

donmccurdy:
> How do you feel about making a separate demo for drag and drop, though?

I'm good with whatever you prefer – did you have in mind a separate demo that only includes drag-and-drop and rendering? or that also includes AO baking?

Will look into the rendering difference. I wonder if it's related to differences in the unweld/weld part...

gkjohnson:
Hmm -- I haven't thought about this too much so any feedback is appreciated but regarding demos here's the demos I'm imagining in the longer term now:

1. A "beauty" demo that shows off the potential of the path tracer with multiple loadable models (maybe a couple -- lego, sketchfab models, etc).
2. A simpler path tracing demo that can be used as a boiler plate for using the path tracer in a project.
3. A simple demo showing how to bake ao maps / light maps.
4. A dual-material orb page for easy material adjustment and testing.
5. A drag and drop viewer that shows off path tracer capabilities for user models and enables ao baking / saving.

Though I hadn't considered a demo with ao map saving until you suggested it 😁 Granted it might be the type of thing that deserves it own dedicated project / viewer / editor at some point. Maybe even in the three.js editor. But I think a demo here would serve as a good starting point.

How does that sound?

donmccurdy:
> Will look into the rendering difference. I wonder if it's related to differences in the unweld/weld part...

Here's a modified copy of the original AntiqueCamera sample, with the unweld+mikktspace+weld steps applied beforehand:

[AntiqueCamera.mod.glb.zip](https://github.com/gkjohnson/three-gpu-pathtracer/files/8384336/AntiqueCamera.mod.glb.zip)

Results are identical to the (broken) original screenshot, so I don't think there's anything specifically in this PR that is wrong. I think it's more likely that something about my unweld+mikktspace+weld implementation that differs from what's now in three.js, more likely in the weld step because the other two should be identical.

I do notice that the glTF attached has 8 validation errors, related to 0-length tangents. The most visible differences are in these elongated triangles on the camera body, so I wonder if there's something problematic in the geometry here? All versions render the same in three.js as far as I can see.

Screen Shot 2022-03-30 at 5 40 34 PM

I may not spend more time on that unless this turns out to be a more wide-spread issue, since I haven't had any reports of trouble with unweld/weld otherwise.

donmccurdy:
The example suggestions sound great! I guess UV unwrapping and packing are the urgent needs right now then?

> Granted it might be the type of thing that deserves it own dedicated project / viewer / editor at some point. Maybe even in the three.js editor. But I think a demo here would serve as a good starting point.

I am definitely tempted to try to do something like that in gltf-transform CLI or https://gltf.report/ at some point... But for the CLI, the headless-gl dependency would be an obstacle — it introduces a major source of installation challenges for users who might not be using that feature at all — so maybe better as a separate CLI or some kind of plugin to gltf-tranform/cli. And for https://gltf.report/ i've got a lot of other yak-shaving that would need to happen first.

So anyway, yeah a demo here sounds like a great starting point to me. :)

gkjohnson:
> Here's a modified copy of the original AntiqueCamera sample, with the unweld+mikktspace+weld steps applied beforehand:
>
> [AntiqueCamera.mod.glb.zip](https://github.com/gkjohnson/three-gpu-pathtracer/files/8384336/AntiqueCamera.mod.glb.zip)

Interesting - when I load the above file directly the AO component looks correct. This is after I've deinterleaved the attributes to support geometry merging and worker bvh generation (https://github.com/mrdoob/three.js/pull/23814). Once this gets closer to being mergeable I can take a deeper look to see what might be going on.

> The example suggestions sound great! I guess UV unwrapping and packing are the urgent needs right now then?

In terms of the ability to generate AO maps, yup! As you've seen #33 includes some info for the next steps around ao map generation. Robert Long mentioned an interest in contributing to this, as well.

donmccurdy:
> This is after I've deinterleaved the attributes to support geometry merging and worker bvh generation ...

Ah ok! I should note the demo here (using `@gltf-transform/view`) will not create interleaved geometry though.

package.json
```diff
@@ -57,5 +62,8 @@
   "bugs": {
     "url": "https://github.com/gkjohnson/three-gpu-pathtracer/issues"
   },
-  "homepage": "https://github.com/gkjohnson/three-gpu-pathtracer#readme"
+  "homepage": "https://github.com/gkjohnson/three-gpu-pathtracer#readme",
+  "alias": {
+    "buffer": false
+  }
```
donmccurdy, line 68: Parcel tries to automatically install [buffer](https://www.npmjs.com/package/buffer), and that messes some things up here, because my packages contain some lines like this:

```js
if ( typeof Buffer !== 'undefined' ) {
  // ...
}
```

So this just tells Parcel to ignore it.

donmccurdy:
Made some updates here — this is now a new `dragDrop.html` demo without the AO-baking component for the time being. I had it all working on a previous version of three-gpu-pathtracer but it needs some updates for the latest release API, and I haven't gotten to that. A few screenshots in the meantime:

Screen Shot 2022-04-29 at 12 00 01 AM
Screen Shot 2022-04-29 at 12 00 22 AM
Screen Shot 2022-04-29 at 12 00 52 AM

gkjohnson:
Awesome! This looks so great. I'll have to give that last model a test to see why those trees aren't being cut out the way I expect them to, though 🤔 

image

gkjohnson:
Just pulled and tested - this is awesome! I just pushed a fix for the crashing on subsequent file upload. The generator was being reused but disposed of after the first use. Heh it should probably throw an error if using it after disposal 😅. I'll add an issue to three-mesh-bvh. Other than that here's my feedback so, though I imagine some of this is on your list already:

- The canvas should be hidden when a model is being part
- Looks "gltf" files are not supported, yet
- It might be nice if the loading text changed from "parsing" to "generating BVH" so theres a bit more indication of progress.

> it's crashing on certain models

Which other models is it crashing on?

donmccurdy:
> Looks "gltf" files are not supported, yet

Just added .gltf folder parsing - adds a little complexity to normalize file paths but not too bad. 

Here's an example that crashes now, but worked with the previous viewer api:

[HousePainted-v4.glb.zip](https://github.com/gkjohnson/three-gpu-pathtracer/files/8592373/HousePainted-v4.glb.zip)

donmccurdy:
Here's another that fails in a different way but worked previously:

[ShiftHappensForest-v3.glb.zip](https://github.com/gkjohnson/three-gpu-pathtracer/files/8592408/ShiftHappensForest-v3.glb.zip)

gkjohnson:
The model wasn't working due to a max uniforms error. I added a "material reducer" to reuse as many materials / textures across the model as possible to cut down on the number of material uniforms required. In the next release this max uniforms issue shouldn't be a problem anymore.

In terms of the three.js error that should be fixed in the next release by https://github.com/mrdoob/three.js/pull/23843

I'll take a look at why the alpha clip isn't working on that model 🤔

donmccurdy:
>  I added a "material reducer" to reuse as many materials / textures across the model as possible to cut down on the number of material uniforms required

Interesting! Is this looking for strictly identical materials and de-duplicating them? Or something else? Also an option:

```js
await modelDocument.transform(
  dedup( { propertyTypes: [ PropertyType.MATERIAL ] } )
);
```

gkjohnson:
> Interesting! Is this looking for strictly identical materials and de-duplicating them? 

It's looking at all material fields aside from things like "UUID" - I prefer the reducer approach in this case since users can rely on it regardless of whether or not gltf-transform is used, I think, but I don't have strong opinions.

And it looks like the alpha clip issue is caused by an overly small "RAY_OFFSET" value. So increasing that a bit to 1e-4 should fix it. I'll look into that on the main branch.

PR #144, opened 2022-05-12
Provide defaults for material's normalScale and emissive values.

joewhatkins:
Provide for defaults for these properties when parsing them in updateFrom.

Ideally this behaves more like it did at https://github.com/gkjohnson/three-gpu-pathtracer/blob/4fc62182d4fe03c834515de388d0af7a6fb8ea24/src/uniforms/MaterialStructUniform.js


gkjohnson:
Great - thank you!

Issue #156, opened 2022-05-23
Demos broken on pixel 6


gkjohnson:
Demos are broken on Pixel 6 specifically. The material ball seems unaffected. v0.0.2 seemed to work so some feature after #117 must have broken it. These are the PR commits to test:

- Matte [53bbd10](https://github.com/gkjohnson/three-gpu-pathtracer/pull/117/commits/53bbd1067a38852fb52695e80b951b58c32c43c3)
- Texture Materials [32a4595](https://github.com/gkjohnson/three-gpu-pathtracer/pull/118/commits/32a45958e2663a5ee77ea26aebd939932bf03820)
- HDR CDF [91ad32d](https://github.com/gkjohnson/three-gpu-pathtracer/pull/120/commits/91ad32df91d99580a55cb2103fc2f1dae980b394)
- Ray Jitter [c8c8490](https://github.com/gkjohnson/three-gpu-pathtracer/pull/128/commits/c8c84909cc2c9ba4713bda7b03a467e2a4ae5fdf)
- Ray Jitter 2 [84bb906](https://github.com/gkjohnson/three-gpu-pathtracer/pull/129/commits/84bb906c18d6a73da61ce6cf2d6dda18b231251c)
-  Blur Generator [008ce71](https://github.com/gkjohnson/three-gpu-pathtracer/pull/138/commits/008ce71d2b4130e6e330ecad4b47f32b813c6743)
- Transmissive MIS [202175a](https://github.com/gkjohnson/three-gpu-pathtracer/pull/139/commits/202175a8f6cf8d0e0006c279b988d0bbcfafbe82)
- Transparent BG [e80053f](https://github.com/gkjohnson/three-gpu-pathtracer/pull/143/commits/e80053fe222208a358c70dcf3e29cb20829309cd)
- Adjust alpha ray offset [6524a9b](https://github.com/gkjohnson/three-gpu-pathtracer/pull/149/commits/6524a9bb61bfa42bc8f74f053772dfcb9b21148e)

Related tweet:

https://twitter.com/garrettkjohnson/status/1526364743862128641

<img src="https://user-images.githubusercontent.com/734200/169912881-0cf46d4b-4776-42f1-87cd-92c26594408a.png" width="200px"/>

<img src="https://user-images.githubusercontent.com/734200/169912915-4b3495d5-eba4-4944-96c8-cf93382f2071.png" width="200px"/>

gkjohnson:
From @setpixel:

[...]

PR #157, opened 2022-05-24
Remove usage of floatBitsToInt to fix Android Bug

setpixel:
Changed material ints to floats to fix bug on android not being able to use floatBitsToInt()

We should submit a bug to the android chrome team to fix.


gkjohnson:
Related to #156

Issue #82, opened 2022-04-07
Selectively disable materials from casting shadows


gkjohnson:
So enclosed room internals can still be affected by an environment map.

The material / object would just be ignored on subsequent bounces.

gkjohnson:
@setpixel my working idea for an implementation here is to just ignore any materials that have shadow casting disabled only after the first bounce - effectively treating them as 100% transparent. But that means that reflections, emission, etc would also all be ignored. Is that what this feature does in Blender, as well?

gkjohnson:
From @setpixel:

Blender lets you disable an object from being visible to certain ray types. See docs [here](https://docs.blender.org/manual/en/latest/render/cycles/object_settings/object_data.html#ray-visibility):

![image](https://user-images.githubusercontent.com/734200/163895205-f71effc8-f75d-4b95-b1bc-aa08c83120ce.png)

In these images the room (walls + ceiling) are set to not cast shadows but still show up in the reflections of the sphere:

![image](https://user-images.githubusercontent.com/734200/163895200-6cfea5cd-f61f-41f4-a07c-8ef96ea65536.png)

![image](https://user-images.githubusercontent.com/734200/163895323-995eda42-53c9-4ec8-9b28-b4a1d3cf3323.png)

In this portion of the reflection it looks like the reflection mixes both the skybox lighting and the reflected lighting:

![image](https://user-images.githubusercontent.com/734200/163895424-894e20f7-b3f4-4495-8d09-48789cd07d95.png)

PR #160, opened 2022-05-28
Add support for disabling shadow casting per material

gkjohnson:
Related to #82
Related to #161

Adds a flag for disabling the casting of shadows from certain materials. Works by disabling hits for materials only when the direct light sampling is used which primarily contributes lighting to diffuse surfaces. How can we determine a specular ray from a diffuse ray? From the sampling? From the probability? From how far off from a perfectly specular reflection it is?

```js
ptMaterial.materials.setCastShadow( index, value );
```

**TODO**

- [x] Update Docs
- [x] Add demo flag
- [x] Get this working without MIS
  - determine whether a ray is "specular" without MIS
  - add surface skipping if not MIS ray
  - check specular ray probability and use to randomly pass through the surface or not (or use a fixed threshold?)
  - OR check how far off of the ideal specular direction the ray is.

Images of the orb in an entirely enclosed box lit by an environment map:

K6aCHFjO

7Ss5iNXK
 
cc @setpixel

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ ptRenderer.alpha = true;
 // init quad for rendering to the canvas
 const fsQuad = new FullScreenQuad( new THREE.MeshBasicMaterial( {
 	map: ptRenderer.target.texture,
-	
+
 	// if rendering transparent background
 	blending: THREE.CustomBlending,
 } ) );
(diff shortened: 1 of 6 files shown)


gkjohnson:
**MIS**
image

image

**No MIS**
image

image

PR #162, opened 2022-05-28
Orthographic camera support

### LINKED ISSUE #152: Orthographic camera support
Hey! An incredible project! Are you planning to add an orthographic camera in the near future? This would make it possible to create cool images like this one

164291812-ab100a57-8c62-4d50-b3c3-5a8cd2a74faa

It would also be great to see the fog effect. For example
 
164180642-2c65f372-6d16-4f77-9658-6d41b6a44075

I created this image with the support of Erich Loftis and his amazing project:
 https://erichlof.github.io/THREE.js-PathTracing-Renderer/Volumetric_Rendering.html

It will be very cool to have a react area light, a directional light and a spotlight

JustASquid:
Resolves #152 

Still TODO:

- [x] Fix jitter/sampling. Can we have a simplified/common jitter implementation that works for both orthographic and perspective cameras?
- [x] Should we add an ortho camera to any other examples?
- [x] Ensure DOF isn't broken

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -109,7 +109,7 @@ let envMap, envMapGenerator;
 let loadingModel = false;
 let delaySamples = 0;
 
-const orthoWidth = 3;
+const orthoWidth = 2;
 
 init();
 
--- a/example/index.js
+++ b/example/index.js
@@ -104,7 +104,7 @@ const params = {
 let creditEl, loadingEl, samplesEl;
 let floorPlane, gui, stats, sceneInfo;
 let renderer, orthoCamera, perspectiveCamera, activeCamera;
-let	ptRenderer, fsQuad, orthoControls, perspectiveControls, scene;
+let	ptRenderer, fsQuad, controls, scene;
 let envMap, envMapGenerator;
 let loadingModel = false;
 let delaySamples = 0;
@@ -148,13 +148,8 @@ async function init() {
 		blending: CustomBlending
 	} ) );
 
-	perspectiveControls = new OrbitControls( perspectiveCamera, renderer.domElement );
-	perspectiveControls.update();
-	perspectiveControls.addEventListener( 'change', resetRenderer );
-
-	orthoControls = new OrbitControls( orthoCamera, renderer.domElement );
-	orthoControls.update();
-	orthoControls.addEventListener( 'change', resetRenderer );
+	controls = new OrbitControls( perspectiveCamera, renderer.domElement );
+	controls.addEventListener( 'change', resetRenderer );
 
 	envMapGenerator = new BlurredEnvMapGenerator( renderer );
 
@@ -466,34 +461,34 @@ function updateEnvBlur() {
 
 }
 
-
-function switchControls( src, dst ) {
-
-	dst.object.position.copy( src.object.position );
-	dst.target.copy( dst.target );
-	dst.zoom = src.zoom;
-	dst.enabled = true;
-	src.enabled = false;
-	dst.update();
-
-}
-
-
 function updateCamera( cameraProjection ) {
 
 	if ( cameraProjection === "Perspective" ) {
 
+		if ( activeCamera ) {
+
+			perspectiveCamera.position.copy( activeCamera.position );
+
+		}
+
 		activeCamera = perspectiveCamera;
-		switchControls( orthoControls, perspectiveControls );
 
 	} else {
 
+		if ( activeCamera ) {
+
+			orthoCamera.position.copy( activeCamera.position );
+
+		}
+
 		activeCamera = orthoCamera;
-		switchControls( perspectiveControls, orthoControls );
 
 	}
 
+	controls.object = activeCamera;
 	ptRenderer.camera = activeCamera;
+
+	controls.update();
 
 	resetRenderer();
 
--- a/example/index.js
+++ b/example/index.js
@@ -14,6 +14,7 @@ import {
 	WebGLRenderer,
 	Scene,
 	PerspectiveCamera,
+	OrthographicCamera,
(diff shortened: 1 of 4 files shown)


JustASquid:
I've made a refactor to the ray generation function, which addresses an existing TODO as well as makes the code fairly uniform between perspective/ortho cameras. Interestingly, it appears that the function `ndcToCameraRay` is not needed, and so I've eliminated a fair few operations. Let me know if I've missed something though!

example/index.js
```diff
@@ -435,6 +466,39 @@ function updateEnvBlur() {
...
+
+	dst.object.position.copy( src.object.position );
+	dst.target.copy( dst.target );
+	dst.zoom = src.zoom;
+	dst.enabled = true;
+	src.enabled = false;
+	dst.update();
+
+}
+
```
gkjohnson, line 480: Is it not possible to swap out the camera on the same controls object? I think ideally there would just be a single OrbitControls.

gkjohnson (commented):
This is great! There's something fun about ortho renders. I added a few comments on some changes I think should be made, as well.

> Should we add an ortho camera to any other examples?

I've been using the "materialBall" demo as an all-encompassing test scene so it would be nice to have the option there, as well.

image

image

src/core/PathTracingRenderer.js
```diff
@@ -48,6 +48,9 @@ function* renderTask() {
 
 				material.cameraWorldMatrix.copy( camera.matrixWorld );
 				material.invProjectionMatrix.copy( camera.projectionMatrixInverse );
+				// An orthographic projection matrix will always have the bottom right element == 1
+				// And a perspective projection matrix will always have the bottom right element == 0
+				material.isOrthographicCamera = camera.projectionMatrix.elements[ 15 ] > 0;
```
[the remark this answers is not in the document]
JustASquid: We only have access to the inverted projection matrix inside the shader afaik. I don't know enough about the theory to be able to make any confident statements about the inverted matrix, but if you can point me to some reference I'd be quite happy to make the change.

example/index.js
JustASquid: First thing I tried, it didn't seem to work (OrbitControls is a bit of a weird beast in my experience). You're right though, it would be way better if there was just one, maybe it warrants some more hacking...

src/core/PathTracingRenderer.js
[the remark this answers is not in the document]
gkjohnson: The projection matrix is passed into the shader by three.js so we should be able to use it. The three.js line material example checks it like so:

https://github.com/mrdoob/three.js/blob/63958ace0b2db846381bd35c76724edfeed7a720/examples/jsm/lines/LineMaterial.js#L139

Technically we could also just call `inverse( projectMatrix )` in the shader to get the inv projection matrix, as well 🤷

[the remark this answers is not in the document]
gkjohnson: Strike what I said above -- that actually won't work. What you've got is good for now, I think.

JustASquid:
One interesting thing I noticed when adding an orthogonal camera to the materialBall example: The envmap background gets projected from a single point. This makes perfect sense if the background is assumed to exist at a distance of infinity. I'd have to imagine THREE.js handles it differently in the rasterizer though, I wonder how it's parameterized? 

I would suggest that fixing this is out of scope for this PR, but interesting nonetheless.

gkjohnson:
I haven't tested yet but what I imagine happening here is right (just a solid color) - it doesn't really make sense to display an env map as a background with an ortho camera. Three.js does not handle this gracefully as it is:

https://jsfiddle.net/8v9gw32b/

gkjohnson (commented):
Just a couple more comments - also I'm seeing in the "materialBall" demo that there's an error in the console when increasing the number of tiles to render with.

Other than that I think this looks great! And DoF seems to work with it, too, which is a great plus

example/materialBall.js
```diff
@@ -8,9 +8,13 @@ import { RGBELoader } from 'three/examples/jsm/loaders/RGBELoader.js';
...
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
 
-let renderer, controls, sceneInfo, ptRenderer, camera, fsQuad, materials;
+let renderer, controls, sceneInfo, ptRenderer, activeCamera, fsQuad, materials;
+let perspectiveCamera, orthoCamera;
 let envMap, envMapGenerator, scene;
 let samplesEl;
+
+const orthoWidth = 5;
+
```
gkjohnson, line 17: Last question and this otherwise looks great! What is "orthoWidth" in these examples and why it it different in each one?

JustASquid: Sets the size of the frame. I eyeballed it so that the image looks similar to the perspective camera (the equivalent value on the perspective camera is FOV, but we don't extract that to a variable because it isn't needed in the resize() callback, but this is.)

gkjohnson:
Thanks again! This is a great addition and with DoF working for it, too! Nice work!

cc @bhouston

vinkovsky:
Hello! I found some problems with the ortho camera

if I switch the camera to ortho, the background appears to be a solid color rather than the environment of the image. Also if I move the camera around the object it shows the env cube

Снимок экрана 2022-06-14 в 15 49 34

perspective view


Снимок экрана 2022-06-14 в 15 50 50

orthographic view




Снимок экрана 2022-06-14 в 15 50 23


orthographic while moving

gkjohnson:
See comment [here](https://github.com/gkjohnson/three-gpu-pathtracer/pull/162#issuecomment-1140376638). That behavior is expected when using an orthographic camera.

PR #163, opened 2022-05-28
Add `alphaMap` support to `PhysicalPathTracingMaterial`

FarazzShaikh:
This PR adds `alphaMap` support to the renderer

image

The puddles in the above demo are via an alpha map'd circle.

### DIFF
--- a/src/materials/PhysicalPathTracingMaterial.js
+++ b/src/materials/PhysicalPathTracingMaterial.js
@@ -187,6 +187,11 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 
 							}
 
+							// alphaMap
+							if ( material.alphaMap != -1 ) {
+								albedo.a *= texture2D( textures, vec3( uv, material.alphaMap ) ).x;
+							}
+
 							// transmission
 							float transmission = material.transmission;
 							if ( material.transmissionMap != - 1 ) {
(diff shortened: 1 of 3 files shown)


src/materials/PhysicalPathTracingMaterial.js
```diff
@@ -187,6 +187,11 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 
 							}
 
+							// alphaMap
+							if ( material.alphaMap != -1 ) {
+								albedo.a *= texture2D( textures, vec3( uv, material.alphaMap ) ).x;
+							}
```
gkjohnson, line 193: nit: can we add extra newlines in the if block and space after the minus in `- 1` to match the others?

gkjohnson (commented):
This is great! Thanks for adding this!

src/materials/PhysicalPathTracingMaterial.js
FarazzShaikh: Surely! Done!

BTW we should add a `.prettierrc` file to the repo because else everyones different Prettier settings will cuase unnecessary large changes

gkjohnson: There's an eslintrc file but I still need to update a few rules. Either way this shader code is in a string so a linter won't catch it :(

gkjohnson:
Thanks again!

PR #170, opened 2022-06-05
Spherical camera projection

### LINKED ISSUE #166: Spherical Camera support
**Is your feature request related to a problem? Please describe.**

The renderer currently supports both the Perspective and Orthographic camera.  Many production renderers, although not games engines, support so called 360 or Spherical cameras.  These can be incredibly useful for rendering panoramic views.

It would be nice to add this to the renderer.

This would be nice for rendering background scenes into envmaps or creating panoramic viewers of complex scenes that otherwise could not be realized in interactive 3D.

**Describe the solution you'd like**

I am unsure how you would specify a spherical camera in Three.js.  Maybe specify it as a CubeCamera?

The output from this renderer should be a latlong arranged image.  So it should varying from -180 degrees to +180 degrees longitude on the x axis.  And from -90 degrees to +90 degrees latitude on the y axis.

Basically the output should look like this:

kitchenb_3

JustASquid:
Resolves #166 

Note: Unlike the orthographic camera, the spherical camera has no direct analogue that works in the rasterizer. So I've only added the new camera to the MaterialBall example for the moment. Happy to discuss though!


gkjohnson:
Thanks! Code looks good! But in terms of API I think I'm more leaning towards an `FEATURE_EQUIRECT_RENDER` #define instead of something like a "SphericalCamera" class that doesn't / can't actually provide a whole lot. What do you think?

Example looks good - I was thinking that adding the option to the "interior scene" demo might make a good amount of sense, too, since that's more the kind of scenario where you might use this kind of render.

image

JustASquid:
@gkjohnson Happy to defer to your opinion of course, but I think having a special camera feels more intuitive to the user. E.g. let's say we have a flag for equirect render, and they are using a Physical camera with DOF settings - it may not be obvious why the DOF settings stop working once they enable the equirect flag.

Adding to the interior scene is a good idea, I'll do that.

bhouston:
I think a Spherical Camera class is also more natural.  But it is awkward that there isn't an equivalent in Three.js.  Although there is CubeCamera in Three.js.  Now it is a spherical camera, but it also has sub-cameras to represent the various faces of the cube.  I wonder if it makes sense to re-interpret CubeCamera as Spherical within this renderer?  It sort of does to me.  But... I am not sure.

bhouston:
> How do people feel about calling the camera an EquirectCamera since that's the spherical projection it's using?

Makes sense.

V-Ray has a single Camera object, but then it has a drop down on that Camera to specify: Perspective, Orthogonal, Spherical Panorama (equirectangular), Box (e.g. CubeMap), FishEye, etc....  https://docs.chaos.com/display/VMAX/Camera

 But then the camera has a lot of additional properties because it is sort of the merger of all of these camera types into one.

JustASquid:
Made the suggested changes and added the new camera(s) to the interior scene. I had to move the camera position a bit for the best effect.

example/materialBall.js
```diff
@@ -7,9 +7,10 @@ import { PathTracingSceneWorker } from '../src/workers/PathTracingSceneWorker.js
 import { RGBELoader } from 'three/examples/jsm/loaders/RGBELoader.js';
 import { MeshoptDecoder } from 'three/examples/jsm/libs/meshopt_decoder.module.js';
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
+import { EquirectCamera } from "../src/core/EquirectCamera";
```
gkjohnson, line 10: nit: double -> single quotes

example/interior.js
```diff
@@ -7,8 +7,10 @@ import { PathTracingSceneWorker } from '../src/workers/PathTracingSceneWorker.js
 import { RGBELoader } from 'three/examples/jsm/loaders/RGBELoader.js';
 import { MeshoptDecoder } from 'three/examples/jsm/libs/meshopt_decoder.module.js';
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
+import { EquirectCamera } from "../src/core/EquirectCamera";
```
gkjohnson, line 10: nit: double -> single quotes

gkjohnson (commented):
Code looks great, nice work again!

My only other comment is that it's not quite clear what the camera control behavior is when equirect is enabled in the examples. For the material ball example the camera seems to rotate around the ball instead of the camera center (which is what I'd expect) and in the interior scene it seems to rotate around the camera origin.

It might just be okay to disable camera controls when enabling the equirect camera and snapping to a fixed location. Then we can start the camera off in a good location for perspective rendering and still pick a more ideal spot for the equirect rendering. What do you think?

JustASquid:
Hey @gkjohnson The reason the camera controls feel funny in the interior scene is that the camera target is very close to the camera origin, this is intended to give the ability to look around the room without the camera needing to leave the room. We could disable the controls but I don't know if that's worthwhile? We don't really want to imply that the equirect camera can't move/rotate. What do you think?

gkjohnson:
I think I just want the interaction to be less confusing in the example. When the user is orbit-rotating the camera with the equirect render it's really not clear what's happening especially with the material orb staying in one spot. Realistically I don't expect anyone to be looking at the equirect render the way it's showing in the demo - it's really just for debugging and to show it's possible. If there winds up being confusion around the capability of the equirect camera then we can change it but I think a static camera should be suitable enough. If it is going to rotate in that I think I expect it to rotate around the camera origin.

So I think those are my two preferred alternatives:
- static camera a hand-picked spot with no rotation
- static camera at a custom hand-picked spot with rotation only around the camera origin

That's my angle, at least. Happy to hear other options!

JustASquid:
For rotation around the camera origin, maybe it would be more natural to use [PointerLockControls](https://threejs.org/docs/#examples/en/controls/PointerLockControls)?

gkjohnson:
I think the method you used with an OrbitControls and short radius worked pretty well. The simplest approach might just be two sets of OrbitControls - one for the equirect camera that sets the min / max distances to the same really small value.

src/materials/PhysicalPathTracingMaterial.js
```diff
@@ -50,7 +50,10 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 				textures: { value: new RenderTarget2DArray().texture },
 				cameraWorldMatrix: { value: new Matrix4() },
 				invProjectionMatrix: { value: new Matrix4() },
-				isOrthographicCamera: { value: true },
+				// 0 = Perspective
+				// 1 = Orthographic
+				// 2 = Equirectangular
+				cameraType: { value: 0 },
```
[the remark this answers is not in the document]
JustASquid: No problem @gkjohnson I'll merge and look into that.

gkjohnson:
Excellent work again! Thank you!

Issue #161, opened 2022-05-28
Add branch for perfectly specular reflection / smoother surface specularity


gkjohnson:
The GGX function does not behave well at at small tight values and incurs floating point error artifacts meaning we have to keep a small value to avoid them which blurs our final result. Instead if a roughness value is below a certain threshold we could interpolate towards the perfect vector to avoid the artifacts and get perfect specularity.

See progressively blurring reflections here with 0 roughness:
![170802001-5f5b9123-1122-435c-8898-cc85f95af141](https://user-images.githubusercontent.com/734200/170802951-4662e709-fd55-465d-90bb-9639d00cc366.jpg)

gkjohnson:
Just setting the ray is not enough - the PDF function must be adjusted, as well. It's possible the PDF function is where the precision issues are cropping up

bhouston:
I can put a bounty on an accepted PR that fixes this issue of $250 USD sponsored by @Threekit.

bhouston:
Be aware that there is now a pending PR for this!  https://github.com/gkjohnson/three-gpu-pathtracer/pull/171

PR #171, opened 2022-06-09
Fix enabling perfect specular reflections.

richardassar:
Fix for https://github.com/gkjohnson/three-gpu-pathtracer/issues/161

- shaderUtils.js/acosApprox(...): added small-order polynomial approximation of `acos`, removing banding artifacts
- shaderGGXFunctions.js/ggxDistribution(...): `theta` calculated with `acos` approximation
- shaderMaterialSampling.js/specularDirection(...): switched from `filteredRoughness` to `rougness`

Reproduced the issue (having scene.js files for a repro in such cases would be helpful). The issue is caused by `filteredRoughness` being used in the `specularDirection` function.

Screenshot from 2022-06-09 00-37-34

Switching to `roughness` reveals the precision issues of the PDF function.

Screenshot from 2022-06-09 00-40-09

Replacing `theta` in `ggxDistribution` with an approximation resolves this. Shown here is a mirror-box with no emissive component and a higher number of bounces demonstrates perfect specular reflections.

Screenshot from 2022-06-09 01-59-08

I took a look at the other example scenes and they appeared unchanged.

### DIFF
--- a/src/shader/shaderUtils.js
+++ b/src/shader/shaderUtils.js
@@ -226,6 +226,9 @@ export const shaderUtils = /* glsl */`
 
 	}
 
+	// Fast arccos approximation used to remove banding artifacts caused by numerical errors in acos.
+	// This is a cubic Lagrange interpolating polynomial for x = [-1, -1/2, 0, 1/2, 1].	
+	// Source: https://stackoverflow.com/questions/3380628/fast-arc-cos-algorithm/answer#3380723	
 	float acosApprox(float x) {
 		return ( -0.69813170079773212 * x * x - 0.87266462599716477 ) * x + 1.5707963267948966;
 	}
--- a/src/shader/shaderUtils.js
+++ b/src/shader/shaderUtils.js
@@ -225,4 +225,8 @@ export const shaderUtils = /* glsl */`
 		return vec3( sinPhi * cos( theta ), cos( phi ), sinPhi * sin( theta ) );
 
 	}
+
+	float acosApprox(float x) {
+		return ( -0.69813170079773212 * x * x - 0.87266462599716477 ) * x + 1.5707963267948966;
+	}
 `;
--- a/src/materials/PhysicalPathTracingMaterial.js
+++ b/src/materials/PhysicalPathTracingMaterial.js
@@ -568,16 +568,11 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 						// then we can just always assume this is a front face.
 						surfaceRec.frontFace = side == 1.0 || transmission == 0.0;
 
-						// Compute the filtered roughness value to use during specular reflection computations. A minimum
-						// value of 1e-6 is needed because the GGX functions do not work with a roughness value of 0 and
-						// the accumulated roughness value is scaled by a user setting and a "magic value" of 5.0.
+						// Compute the filtered roughness value to use during specular reflection computations.
+						// The accumulated roughness value is scaled by a user setting and a "magic value" of 5.0.
 						// If we're exiting something transmissive then scale the factor down significantly so we can retain
 						// sharp internal reflections
-						surfaceRec.filteredRoughness = clamp(
-							max( surfaceRec.roughness, accumulatedRoughness * filterGlossyFactor * 5.0 ),
-							1e-3,
-							1.0
-						);
+						surfaceRec.filteredRoughness = clamp( max( surfaceRec.roughness, accumulatedRoughness * filterGlossyFactor * 5.0 ), 0.0, 1.0 );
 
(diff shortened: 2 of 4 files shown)


richardassar:
This plot shows the `acos` approximation used https://www.desmos.com/calculator/zj2pfxbyoe

richardassar:
Mirrors in the `interior.html` scene are now sharp as a result.

Before:
Screenshot from 2022-06-09 02-59-49_


After:
Screenshot from 2022-06-09 02-56-07_
Screenshot from 2022-06-09 02-56-07_crop

richardassar:
Using `filteredRoughness` in `specularDirection` instead requires widening the clamp to [0, 1], and adjusting `accumulatedRoughness` to also use the `acos` approximation. An less-than epsilon check is necessary to avoid division by values close to zero in `ggxDistribution`.

The `acos` approximation can be applied elsewhere but I've opted to restrict its use only to the sources of the precision issues.

Screenshot from 2022-06-09 08-53-19

richardassar:
Screenshot from 2022-06-09 09-25-52

bhouston:
Really nice work!

richardassar:
Thanks!

src/shader/shaderUtils.js
```diff
@@ -225,4 +225,8 @@ export const shaderUtils = /* glsl */`
 		return vec3( sinPhi * cos( theta ), cos( phi ), sinPhi * sin( theta ) );
 
 	}
+
+	float acosApprox(float x) {
+		return ( -0.69813170079773212 * x * x - 0.87266462599716477 ) * x + 1.5707963267948966;
+	}
```
gkjohnson, line 231: Is there a reference for this approximation? Can we add a comment with some information on it?

gkjohnson (commented):
Great work, this is excellent! Thank you - I hadn't thought the built in acos would cause these kinds of issues. Do you know why that is? Is it just known to be an imprecise call?

src/shader/shaderUtils.js
richardassar: > Great work, this is excellent! Thank you - I hadn't thought the built in acos would cause these kinds of issues. Do you know why that is? Is it just known to be an imprecise call?

`acos` on the GPU is implemented with something like this https://developer.download.nvidia.com/cg/acos.html - the number of operations and the use of `sqrt` all contribute to numerical precision issues, this is further worsened when denormals are flushed to zero. 

The reported absolute error <= 6.7e-5 detailed in the above page should be fine in most cases but when interacting with other functions in this iterative setting this gives rise to pathological fixed-points.

richardassar: > Is there a reference for this approximation? Can we add a comment with some information on it?

This is a Lagrange interpolating polynomial for the points x∈{-1, -½, 0, ½, 1} e.g. https://mathworld.wolfram.com/LagrangeInterpolatingPolynomial.html 

The values themselves come from https://stackoverflow.com/questions/3380628/fast-arc-cos-algorithm/answer#3380723

The cubic polynomial factored as it is saves a few operations, which is a small extra win.

richardassar:
> Great work, this is excellent! Thank you - I hadn't thought the built in acos would cause these kinds of issues. Do you know why that is? Is it just known to be an imprecise call?

`acos` on the GPU is implemented with something like this https://developer.download.nvidia.com/cg/acos.html - the number of operations and the use of `sqrt` all contribute to numerical precision issues, this is further worsened when denormals are flushed to zero. 

The reported absolute error <= 6.7e-5 detailed in the above page should be fine in most cases but when interacting with other functions in this iterative setting this gives rise to pathological fixed-points.

src/shader/shaderUtils.js
richardassar: I'll add a comment and update the PR.

gkjohnson:
Awesome, thanks! This is all really great information and would be good to include in a comment for `acosApprox`.

richardassar:
You're welcome. Updated with a comment, let me know if you'd like more elaboration or not there.

gkjohnson:
Thanks again!

richardassar:
Just thought I'd add this Mathematica example computing the Lagrange polynomial approximation as it provides more context than that stackoverflow post. 

You can use this replacing `ArcCos` with other functions if in future they also become sources of similar numerical issues.

Screenshot from 2022-06-10 12-37-41

gkjohnson:
@richardassar It looks like this might have introduced some artifacts. Any thoughts on how they could be addressed? This is with default settings on the `materialBall` example:

**BEFORE**
image

**AFTER**
image

richardassar:
Resolved in https://github.com/gkjohnson/three-gpu-pathtracer/pull/176

Issue #113, opened 2022-04-26
Support for spotlights


DanielDoesDesign:
I tried to implement an example of light shining through a transmissive surface only to find the raycaster does not recognize light sources. 

https://i.imgur.com/VnmmqaV.png

https://i.imgur.com/Dz23nMY.jpeg

The effect I am trying to replicate is that similar of the "material ball" example. With a small radial point being emissive with a wide 120 degree angle.

Ben-Mack:
Yes, +1 for basic lights support like SpotLight and DirectionalLight

gkjohnson:
Area lights and volume lights will probably come long before some of these other more fake lights.

gkjohnson:
Sounds like @richardassar is looking into spot light support

richardassar:
Working on adding basic support just now and as a stretch goal I might look into allowing the user to apply a preset texture like this:

![Spot light presets](https://i0.wp.com/odederell3d.blog/wp-content/uploads/2018/09/cgl_spotlight_presets_pack_1-0_previews.jpg?resize=648%2C534&ssl=1)

richardassar:
Sneak peek at the spot lights feature WIP.

![Screenshot 2022-06-13 at 00-06-03 Material Orb Path Tracing](https://user-images.githubusercontent.com/868396/173257271-1ece8895-2613-4df3-968a-0a438895d053.png)

Builds on top of the area lights work by introducing a light type field to the light struct and a PDF tailored to spot lights, all that remains is to add support for all of the THREE.js spot light parameters such as penumbra and decay.

bhouston:
[richardassar](https://github.com/richardassar) Here is the IES light parser we used in Clara.io: https://gist.github.com/bhouston/f2d1935aeabe2630d20c10390ce17cb5

What we did is just create a texture out of it and then use that to modulate the light intensity out of the light.

I think that IES support could be a separate issue and PR with a separate bounty of $250 USD.  I think it is important to keep things bit size.

bhouston:
@richardassar Here is the IES light parser we used in Clara.io: https://gist.github.com/bhouston/f2d1935aeabe2630d20c10390ce17cb5

What we did is just create a texture out of it and then use that to modulate the light intensity out of the light.

I think that IES support could be a separate issue and PR with a separate bounty of $250 USD. I think it is important to keep things bit size.

bhouston:
[...]

PR #176, opened 2022-06-11
Fix for artifacts introduced by Perfect Specular Reflections feature

richardassar:
Problem reported in https://github.com/gkjohnson/three-gpu-pathtracer/pull/171#issuecomment-1152805532

`ggxDistribution` now uses `acosApprox` only when `roughness` is zero, otherwise uses the introduced `acosSafe` which is clamped to the input range [-1, 1]. Lowered the epsilon check for `denom`, returning Pi as a suitably large value in this case.

**Before the fix:**
Screenshot 2022-06-11 at 08-57-21 Material Orb Path Tracing

**After the fix:**
Screenshot 2022-06-11 at 08-59-05 Material Orb Path Tracing

**Banding artifacts remain solved in the roughness = 0 case:**
Screenshot 2022-06-11 at 09-08-10 Material Orb Path Tracing

**The mirror box scene still shows perfect reflections for each bounce:**
Screenshot 2022-06-11 at 09-01-41 Material Orb Path Tracing

**No noticeable artifacts in the other example scenes:**
Screenshot 2022-06-11 at 09-06-30 PBR Path Tracing
Screenshot 2022-06-11 at 09-04-15 Interior Scene Path Tracing

### DIFF
--- a/src/shader/shaderGGXFunctions.js
+++ b/src/shader/shaderGGXFunctions.js
@@ -72,17 +72,18 @@ float ggxDistribution( vec3 halfVector, float roughness ) {
 
 	// See equation (33) from reference [0]
 	float a2 = roughness * roughness;
+	a2 = max( EPSILON, a2 );
 	float cosTheta = halfVector.z;
 	float cosTheta4 = pow( cosTheta, 4.0 );
 
 	if ( cosTheta == 0.0 ) return 0.0;
 
-	float theta = ( roughness == 0.0 ) ? acosApprox( halfVector.z ) : acosSafe( halfVector.z );
+	float theta = acosSafe( halfVector.z );
 	float tanTheta = tan( theta );
 	float tanTheta2 = pow( tanTheta, 2.0 );
 
 	float denom = PI * cosTheta4 * pow( a2 + tanTheta2, 2.0 );
-	return ( denom < 1e-12 ) ? PI : ( a2 / denom );
+	return ( a2 / denom );
 
 	// See equation (1) from reference [2]
 	// const { x, y, z } = halfVector;
--- a/src/shader/shaderGGXFunctions.js
+++ b/src/shader/shaderGGXFunctions.js
@@ -77,12 +77,12 @@ float ggxDistribution( vec3 halfVector, float roughness ) {
 
 	if ( cosTheta == 0.0 ) return 0.0;
 
-	float theta = acosApprox( halfVector.z );
+	float theta = ( roughness == 0.0 ) ? acosApprox( halfVector.z ) : acosSafe( halfVector.z );
 	float tanTheta = tan( theta );
 	float tanTheta2 = pow( tanTheta, 2.0 );
 
 	float denom = PI * cosTheta4 * pow( a2 + tanTheta2, 2.0 );
-	return denom < 1e-6 ? 1.0 : a2 / denom;
+	return ( denom < 1e-12 ) ? PI : ( a2 / denom );
 
 	// See equation (1) from reference [2]
 	// const { x, y, z } = halfVector;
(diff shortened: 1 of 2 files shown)


richardassar:
Note this also impacts the rectangular area lights in https://github.com/gkjohnson/three-gpu-pathtracer/pull/174 which I'm using as the basis for adding the spot light feature request in https://github.com/gkjohnson/three-gpu-pathtracer/issues/113

**Before (MIS on):**
Screenshot 2022-06-11 at 17-36-51 Area Light Path Tracing

**After (MIS on, reflection improved):**
Screenshot 2022-06-11 at 17-37-21 Area Light Path Tracing

**Before (MIS off, fireflies present):**
Screenshot 2022-06-11 at 17-36-13 Area Light Path Tracing

**After (MIS off, fireflies suppressed):**
Screenshot 2022-06-11 at 17-36-20 Area Light Path Tracing

gkjohnson:
I'm still seeing some artifacts when the roughness === 0.0. Notice the halo around the reflection of the ceiling light in the floor that should be perfectly reflective. It averages away after a number of frames but it still seems like there's an issue:

image

Interestingly I noticed that changing this line:

```js
float theta = ( roughness == 0.0 ) ? acosApprox( halfVector.z ) : acosSafe( halfVector.z );
```

to this

```js
float theta = acosSafe( halfVector.z );
```

seems to alleviate the artifacts as far as I can tell. Changing to back to regular `acos` seems to reintroduce the circular artifacts we were seeing before. Perhaps just `acosSafe` is enough to solve the issues?

richardassar:
I'd noticed these also, then reverted to before any of my changes and still saw them. 

However, I have noticed that after a branch change I need to kill and restart `npm run start` so I might have missed this.

I'll check again and will update the PR accordingly.

richardassar:
Checked your suggestion, it works well and is improved by clamping `a2` to EPSILON instead of the `denom` ternary. 

The `acosApprox` in `accumulatedRoughness` is still required when bounces between zero-roughness surfaces is present, otherwise the blurring and precision issues come back.

I've also noticed that the issues you've reported, and some that I'm seeing in developing the spot light go away with importance sampling turned off, so there may be an issue there or by some interaction. I'm investigating further

**MIS off:**
Screenshot from 2022-06-11 21-50-28

src/shader/shaderGGXFunctions.js
```diff
@@ -72,17 +72,18 @@ float ggxDistribution( vec3 halfVector, float roughness ) {
...
 	if ( cosTheta == 0.0 ) return 0.0;
 
-	float theta = acosApprox( halfVector.z );
+	float theta = acosSafe( halfVector.z );
 	float tanTheta = tan( theta );
 	float tanTheta2 = pow( tanTheta, 2.0 );
 
 	float denom = PI * cosTheta4 * pow( a2 + tanTheta2, 2.0 );
-	return denom < 1e-6 ? 1.0 : a2 / denom;
+	return ( a2 / denom );
```
gkjohnson, line 86: With this change (and `max( EPSILON, a2 )`) I'm still seeing halos:

image image

This line you had here previously seemed to behave better, though I'm not sure I follow exactly where it came from:

```
return ( denom < 1e-12 ) ? PI : ( a2 / denom );
```

richardassar: Try removing `envColor` from: 

`gl_FragColor.rgb += attenuatedColor * environmentIntensity * envColor * throughputColor * sampleColor * misWeight / envPdf;`

in `PhysicalPathTracingMaterial.js`

This seems to solve the halo issue but I wanted to examine more closely what is causing a correct looking line to be a factor here, perhaps some quantity is being included twice. 

The above is in an MIS-only `#ifdef` and with MIS-off the halos are still present, this might be informative.

I think `max( EPSILON, a2 )` and `acosSafe` are more sensible changes here though and this is revealing an issue elsewhere, I'll try to track it down.

gkjohnson: I'll merge this for now since I do think it's a big improvement - I made #179 to track any other thoughts on what might be causing this. Thanks again!

Issue #181, opened 2022-06-13
SpotLight IES profile support


bhouston:
**Is your feature request related to a problem? Please describe.**

It would be nice to support IES profiles when rendering spot lights.

Here is a library of IES profiles: [https://ieslibrary.com/en/home](https://t.co/N2yIJ0geJd)

Here is example code for parsing an IES profile: https://gist.github.com/bhouston/f2d1935aeabe2630d20c10390ce17cb5

I believe that for the majority of cases a simple 1D texture can represent IES profile data.

**Describe the solution you'd like**

It would be nice to be able to load an IES profile into a texture and then use that texture as a modulator on SpotLights.  I think it is important to separate the IES profile loading from the texture setting on the spot lights to keep everything simple.

CC: @richardassar

bhouston:
@Threekit is pleased to offer a $250 USD bounty for an accepted PR that implements the feature described in this issue.

richardassar:
In UE you can add them to point and rectangular area lights also, as shown [here](https://docs.unrealengine.com/5.0/en-US/using-ies-light-profiles-in-unreal-engine/).

It seems like something that should be applied to all lights, optionally, with spotlights just masking one half of the profile.

gkjohnson:
Thanks @bhouston! I'm assuming it's okay if we include your IESLoader in this project?

bhouston:
yes. :)

richardassar:
Integrated the IESLoader from @bhouston. 

Required only minimal changes such as changing the texture type from `UnsignedByteType` to `FloatType`, as the precision is necessary, and changing the texture format from `LuminanceFormat` to `RedFormat` to avoid WebGL warnings/errors.

The textures are packed by the newly introduced `IESProfilesTexture`, a `WebGLArrayRenderTarget`, which is stored as a `PhysicalPathTracingMaterial` uniform. Spot lights may optionally use the profiles, falling back to the default behavior when no profile is selected. There's an extra `radius` parameter which can be used to influence shadow softness.

Just need to tidy the code, make a nicer demo scene and  then submit the PR.

![Screenshot 2022-06-14 at 08-33-55 Material Orb Path Tracing](https://user-images.githubusercontent.com/868396/173520092-c5da8b73-2b51-4ad8-a4e7-212b72d8f516.png)

[...]

Issue #96, opened 2022-04-11
Add support for iridescence


gkjohnson:

bhouston:
Iridescence is a very important property in many situations - such as oil slicks, or on leather furniture as well as some exotic car paints or ceramics.

We should follow the glTF iridescence model specified here:

https://github.com/KhronosGroup/glTF/blob/a9cd1cb657587b590b6f95c8654038a4e9d66596/extensions/2.0/Khronos/KHR_materials_iridescence/README.md

It has these properties:

|                                 | Type                                                                                                 | Description                                                       | Required            |
| ------------------------------- | ---------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------| ------------------- |
| **iridescenceFactor**           | `number`                                                                                             | The iridescence intensity factor.                                 | No, default:`0.0`   |
| **iridescenceTexture**          | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | The iridescence intensity texture.                                | No                  |
| **iridescenceIor**              | `number`                                                                                             | The index of refraction of the dielectric thin-film layer.        | No, default:`1.3`   |
| **iridescenceThicknessMinimum** | `number`                                                                                             | The minimum thickness of the thin-film layer given in nanometers. | No, default:`100.0` |
| **iridescenceThicknessMaximum** | `number`                                                                                             | The maximum thickness of the thin-film layer given in nanometers. | No, default:`400.0` |
| **iridescenceThicknessTexture** | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | The thickness texture of the thin-film layer.                     | No                  |

bhouston:
I am pleased to be able to put a bounty of $500 USD on an accepted PR that implements the official glTF iridescence extension in this path tracer.

richardassar:
Have gathered a few resources and started taking a look at this. :microscope:

richardassar:
Here are the results after a quick first pass at this, just showing the dielectric component for now. 

I need to cross-reference with the other glTF extensions to ensure everything is integrated correctly with the existing material code, but here's how it looks at present.

![Screenshot 2022-06-21 at 14-40-30 Material Orb Path Tracing](https://user-images.githubusercontent.com/868396/174819149-d3251a49-21b1-4f5f-995c-81670a5a0342.png)

[...]

PR #184, opened 2022-06-14
Support for spot lights with IES profiles

richardassar:
Implements the features requested in https://github.com/gkjohnson/three-gpu-pathtracer/issues/113 and https://github.com/gkjohnson/three-gpu-pathtracer/issues/181

Screenshot 2022-06-15 at 00-10-53 Spot Lights Path Tracing

Builds on top of the area lights feature added by @mrxz. Merge after https://github.com/gkjohnson/three-gpu-pathtracer/pull/174 has been merged.

- Adds a `type` field to lights, allowing the use of `RectAreaLight` and `SpotLight`
- The non-IES spot light `angle`, `distance`, `decay` and `penumbra` parameters match the behaviour of THREE.js spotlights in physically accurate mode
- When using IES profiles `angle` and `penumbra` are ignored
- The `radius` parameter offsets the spotlight's position to match that of the cone angle and provides a means for soft shadows
- The `lampIntensityScale` parameter, when not 0, displays a circular area light with an intensity that is scaled by the spot light's main `intensity` parameter, otherwise this is invisible (light counts used in computations elsewhere might need to consider this but it seems fine without)

In `getPhotometricAttenuation` I've left some commented lines for if/when the azimuth angle is to be used, as `IESLoader` provides a 360 x 180 texture. If we don't need this then a 1 x 180 or 180 x 1 texture array can be used instead for the main `IESProfilesTexture`.

**Outstanding issues**

This change also further highlights some issues between the MIS and non-MIS branches, with the spot lights not working at all in the non-MIS branch. 

The final `FEATURE_MIS` block in `PhysicalPathTracingMaterial.js` seems to be lacking a non-MIS counterpart but I could be wrong in thinking this is the case. 

When many lights are present there's also a "warm up" time where pixels affected by the IES lights seem to receive their contributions slowly. This doesn't affect the "lamp" or any rectangular area lights, only the contributions from `randomLightSample`. (edit: this seems to only happen on load, and goes away after any `reset()`, looking into it)

I'd like to add some more IES profiles and improve the demo scene, will submit this for review now and get to that separately.

🔦

### DIFF
--- a/example/spotLights.js
+++ b/example/spotLights.js
@@ -9,7 +9,7 @@ import { MeshoptDecoder } from 'three/examples/jsm/libs/meshopt_decoder.module.j
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
 import { TransformControls } from 'three/examples/jsm/controls/TransformControls.js';
 
-let renderer, controls, transformControlsScene, spotLight1, lights, spotLightHelpers, sceneInfo, ptRenderer, activeCamera, fsQuad, materials;
+let renderer, controls, transformControlsScene, spotLight1, lights, spotLights, spotLightHelpers, sceneInfo, ptRenderer, activeCamera, fsQuad, materials;
 let perspectiveCamera, orthoCamera;
 let envMap, envMapGenerator, scene;
 let samplesEl;
@@ -272,6 +272,7 @@ async function init() {
 			// spot lights
 			spotLightHelpers = [ ];
 			lights = [ ];
+			spotLights = [ ];
 
 			const decays = [ 0, 1.5, 0, 0.25, 0 ];
 			for ( let i = 0; i < 5; ++ i ) {
@@ -283,8 +284,7 @@ async function init() {
 				spotLight.penumbra = 0.0;
 				spotLight.decay = decays[ i ];
 				spotLight.distance = 0.0;
-				spotLight.intensity = 50.0;
-				spotLight.lampIntensityScale = 0.0;
+				spotLight.intensity = 150.0;
 				spotLight.radius = 0.5;
 				spotLight.iesProfile = - 1 + i;
 
@@ -295,6 +295,7 @@ async function init() {
 				spotLight.shadow.focus = 1.0;
 				spotLight.castShadow = true;
 
+				spotLights.push( spotLight );
 				lights.push( spotLight );
 
 				const targetObject = new THREE.Object3D();
@@ -373,7 +374,7 @@ async function init() {
 
 			scene.add( result.scene );
 
-			const { bvh, textures, materials, lights } = result;
+			const { bvh, textures, materials, lights, spotLights } = result;
 			const geometry = bvh.geometry;
 			const material = ptRenderer.material;
 
@@ -387,6 +388,8 @@ async function init() {
 			material.iesProfiles.updateFrom( renderer, iesProfileURLs );
 			material.lights.updateFrom( lights );
 			material.lightCount = lights.length;
+			material.spotLights.updateFrom( spotLights );
+			material.spotLightCount = spotLights.length;
 
 			generator.dispose();
 
@@ -550,7 +553,7 @@ async function init() {
 	matFolder3.add( params.material3, 'castShadow' ).onChange( reset );
 	matFolder3.close();
 
-	const matFolder4 = gui.addFolder( 'Floor Material' );
+	const matFolder4 = gui.addFolder( 'Wall Material' );
 	matFolder4.addColor( params.material4, 'color' ).onChange( reset );
 	matFolder4.add( params.material4, 'roughness', 0, 1 ).onChange( reset );
 	matFolder4.add( params.material4, 'metalness', 0, 1 ).onChange( reset );
@@ -570,7 +573,6 @@ async function init() {
 	matFolder5.add( spotLight1, 'distance', 0.0, 20.0 ).onChange( reset );
 	matFolder5.add( spotLight1, 'angle', 0.0, Math.PI / 2.0 ).onChange( reset );
 	matFolder5.add( spotLight1, 'penumbra', 0.0, 1.0 ).onChange( reset );
-	matFolder5.add( spotLight1, 'lampIntensityScale', 0.0, 1.0 ).onChange( reset );
 	matFolder5.add( spotLight1, 'iesProfile', - 1, iesProfileURLs.length - 1, 1 ).onChange( reset );
 
 	animate();
@@ -706,6 +708,7 @@ function animate() {
 	ptRenderer.material.physicalCamera.updateFrom( activeCamera );
 
 	ptRenderer.material.lights.updateFrom( lights );
+	ptRenderer.material.spotLights.updateFrom( spotLights );
 
 	activeCamera.updateMatrixWorld();
 
--- a/example/spotLights.js
+++ b/example/spotLights.js
@@ -315,7 +315,7 @@ async function init() {
 					spotLight.add( spotLightHelper );
 					transformControlsScene.add( spotLightHelper );
 
-					spotLightHelpers.push ( spotLightHelper );
+					spotLightHelpers.push( spotLightHelper );
 
 					spotLight1 = spotLight;
 
--- a/example/spotLights.js
+++ b/example/spotLights.js
@@ -1,0 +1,753 @@
+import * as THREE from 'three';
+import { FullScreenQuad } from 'three/examples/jsm/postprocessing/Pass.js';
+import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
+import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
+import { PathTracingRenderer, PhysicalPathTracingMaterial, PhysicalCamera, BlurredEnvMapGenerator } from '../src/index.js';
+import { PathTracingSceneWorker } from '../src/workers/PathTracingSceneWorker.js';
+import { RGBELoader } from 'three/examples/jsm/loaders/RGBELoader.js';
+import { MeshoptDecoder } from 'three/examples/jsm/libs/meshopt_decoder.module.js';
+import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
+import { TransformControls } from 'three/examples/jsm/controls/TransformControls.js';
+
+let renderer, controls, transformControlsScene, spotLight1, lights, spotLightHelpers, sceneInfo, ptRenderer, activeCamera, fsQuad, materials;
+let perspectiveCamera, orthoCamera;
+let envMap, envMapGenerator, scene;
+let samplesEl;
+
+const orthoWidth = 5;
+
+const iesProfileURLs = [
+	'https://gist.githubusercontent.com/richardassar/0241a4c03091fc61d7d727d61480f1ed/raw/780409421bf8efe6cc828314d456aa2c8803976c/108b32f07d6d38a7a6528a6d307440df.ies',
+	'https://gist.githubusercontent.com/richardassar/80a5be46f3bf2e91c98cba093f60a443/raw/8650c0c019bfdb120ecc07917c97de9e86c0438d/1aec5958092c236d005093ca27ebe378.ies',
+	'https://gist.githubusercontent.com/richardassar/a5f4362935b59bb8c5ea822f17b178bc/raw/bd03ff9716a0217577d5b0f91c9ee06b363b87d0/02a7562c650498ebb301153dbbf59207.ies',
+	'https://gist.githubusercontent.com/richardassar/8a712f70b99da706a13f4e902b5508e9/raw/d0d469e128b5cdec23cfe1dd10854846f8a6a1c4/1a936937a49c63374e6d4fbed9252b29.ies',
+	'https://gist.githubusercontent.com/richardassar/6cbb7abf80f73a2ca10d39f7d7587556/raw/a459770128719d0af76eac8076376dd4b708e834/00c6ce79e1d2cdf3a1fb491aaaa47ae0.ies'
+];
+
+const params = {
+
+	material1: {
+		color: '#ffc766',
+		emissive: '#FFFFFF',
+		emissiveIntensity: 0.0,
+		roughness: 0.2,
+		metalness: 1.0,
+		ior: 1.495,
+		transmission: 0.0,
+		opacity: 1.0,
+		matte: false,
+		castShadow: true,
+	},
+	material2: {
+		color: '#db7157',
+		emissive: '#FFFFFF',
+		emissiveIntensity: 0,
+		roughness: 0.0,
+		metalness: 0.8,
+		transmission: 0.0,
+		ior: 1.495,
+		opacity: 1.0,
+		matte: false,
+		castShadow: true,
+	},
+	material3: {
+		color: '#3465a4',
+		roughness: 0.4,
+		metalness: 0.4,
+		matte: false,
+		castShadow: false,
+		receiveShadow: true
+	},
+	material4: {
+		color: '#FFFFFF',
+		emissive: '#FFFFFF',
+		emissiveIntensity: 0.0,
+		roughness: 0.4,
+		metalness: 0.1,
+		matte: false,
+		castShadow: false,
+		transmission: 0.0,
+		opacity: 1.0,
+	},
+
+	multipleImportanceSampling: true,
+	stableNoise: false,
+	environmentIntensity: 0.5,
+	environmentRotation: 0,
+	environmentBlur: 0.0,
+	backgroundBlur: 0.05,
+	bounces: 3,
+	samplesPerFrame: 1,
+	acesToneMapping: true,
+	resolutionScale: 1 / window.devicePixelRatio, // TODO: remove before commit
+	transparentTraversals: 20,
+	filterGlossyFactor: 0.5,
+	tiles: 1,
+	backgroundAlpha: 1,
+	checkerboardTransparency: true,
+	showTransformControls: true,
+	cameraProjection: 'Perspective',
+};
+
+if ( window.location.hash.includes( 'transmission' ) ) {
+
+	params.material1.metalness = 0.0;
+	params.material1.roughness = 0.05;
+	params.material1.transmission = 1.0;
+	params.material1.color = '#ffffff';
+	params.bounces = 10;
+
+}
+
+// adjust performance parameters for mobile
+const aspectRatio = window.innerWidth / window.innerHeight;
+if ( aspectRatio < 0.65 ) {
+
+	params.bounces = Math.max( params.bounces, 6 );
+	params.resolutionScale *= 0.5;
+	params.tiles = 2;
+	params.multipleImportanceSampling = false;
+	params.environmentBlur = 0.35;
+
+}
+
+init();
+
+async function init() {
+
+	renderer = new THREE.WebGLRenderer( { antialias: true } );
+	renderer.toneMapping = THREE.ACESFilmicToneMapping;
+	renderer.setClearColor( 0, 0 );
+	renderer.shadowMap.enabled = true;
+	//renderer.shadowMap.type = THREE.PCFSoftShadowMap;
+	document.body.appendChild( renderer.domElement );
+
+	const aspect = window.innerWidth / window.innerHeight;
+	perspectiveCamera = new PhysicalCamera( 75, aspect, 0.025, 500 );
+	perspectiveCamera.position.set( 16, 2, 15 );
+
+	const orthoHeight = orthoWidth / aspect;
+	orthoCamera = new THREE.OrthographicCamera( orthoWidth / - 2, orthoWidth / 2, orthoHeight / 2, orthoHeight / - 2, 0, 100 );
+	orthoCamera.position.set( - 4, 2, 3 );
+
+	ptRenderer = new PathTracingRenderer( renderer );
+	ptRenderer.alpha = true;
+	ptRenderer.material = new PhysicalPathTracingMaterial();
+	ptRenderer.material.setDefine( 'TRANSPARENT_TRAVERSALS', params.transparentTraversals );
+	ptRenderer.material.setDefine( 'FEATURE_MIS', Number( params.multipleImportanceSampling ) );
+	ptRenderer.tiles.set( params.tiles, params.tiles );
+
+	fsQuad = new FullScreenQuad( new THREE.MeshBasicMaterial( {
+		map: ptRenderer.target.texture,
+		blending: THREE.CustomBlending,
+	} ) );
+
+	controls = new OrbitControls( perspectiveCamera, renderer.domElement );
+	controls.target.set( 16, 1, 1 );
+	controls.addEventListener( 'change', () => {
+
+		ptRenderer.reset();
+
+	} );
+
+	scene = new THREE.Scene();
+
+	samplesEl = document.getElementById( 'samples' );
+
+	envMapGenerator = new BlurredEnvMapGenerator( renderer );
+
+	const envMapPromise = new Promise( resolve => {
+
+		new RGBELoader()
+			.load( 'https://raw.githubusercontent.com/mrdoob/three.js/dev/examples/textures/equirectangular/royal_esplanade_1k.hdr', texture => {
+
+				envMap = texture;
+
+				updateEnvBlur();
+				resolve();
+
+			} );
+
+	} );
+
+	transformControlsScene = new THREE.Scene();
+
+	const generator = new PathTracingSceneWorker();
+	const gltfPromise = new GLTFLoader()
+		.setMeshoptDecoder( MeshoptDecoder )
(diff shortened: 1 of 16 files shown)


richardassar:
I have also yet to check the status of (from the original issue):

> I tried to implement an example of light shining through a transmissive surface only to find the raycaster does not recognize light sources.

I checked against perfect reflective / mirror surfaces and don't see the expected behaviour, this should be filed as a separate issue.

**Pointing at floor**
Screenshot 2022-06-15 at 01-12-15 Spot Lights Path Tracing

**Pointing at the mirror**
Screenshot 2022-06-15 at 01-13-33 Spot Lights Path Tracing

I'd expect to see the spot light appear in both the mirror image and the floor, and no illumination of the material ball closest to the camera.

There was also a comment:

> Yes, +1 for basic lights support like SpotLight and DirectionalLight

DirectionalLight would be trivial to add on top of these changes, perhaps as a separate issue.

gkjohnson:
From https://github.com/gkjohnson/three-gpu-pathtracer/issues/96#issuecomment-1163522242 - just wanted to get ahead of this before you put too much time into it:

> I still have to resolve the issue of the spot lights failing to render in non-MIS mode.

I actually don't expect spot lights to function in non-MIS mode and I think it's okay if they don't considering they require a ray hitting a specific point at a pretty specific angle (at least that's my understanding without having had a chance to dig through the code, yet).

Correct me if I'm wrong but any type of lighting that requires hitting really narrow hot spots I expect to be really firefly-ridden without lighting importance sampling. In terms of use case I consider non-MIS mode to be a fallback for lower power cases where you're okay provide a blurry / bright ambient environment map or large light sources.

But let me know if I'm wrong and there are approaches that afford them easily working as expected in without MIS.

example/spotLights.js
```diff
@@ -0,0 +1,753 @@
...
+
+const orthoWidth = 5;
+
+const iesProfileURLs = [
+	'https://gist.githubusercontent.com/richardassar/0241a4c03091fc61d7d727d61480f1ed/raw/780409421bf8efe6cc828314d456aa2c8803976c/108b32f07d6d38a7a6528a6d307440df.ies',
+	'https://gist.githubusercontent.com/richardassar/80a5be46f3bf2e91c98cba093f60a443/raw/8650c0c019bfdb120ecc07917c97de9e86c0438d/1aec5958092c236d005093ca27ebe378.ies',
+	'https://gist.githubusercontent.com/richardassar/a5f4362935b59bb8c5ea822f17b178bc/raw/bd03ff9716a0217577d5b0f91c9ee06b363b87d0/02a7562c650498ebb301153dbbf59207.ies',
+	'https://gist.githubusercontent.com/richardassar/8a712f70b99da706a13f4e902b5508e9/raw/d0d469e128b5cdec23cfe1dd10854846f8a6a1c4/1a936937a49c63374e6d4fbed9252b29.ies',
+	'https://gist.githubusercontent.com/richardassar/6cbb7abf80f73a2ca10d39f7d7587556/raw/a459770128719d0af76eac8076376dd4b708e834/00c6ce79e1d2cdf3a1fb491aaaa47ae0.ies'
+];
```
gkjohnson, line 25: Where did these profiles come from? I can also add them to my "3d-demo-data" repo where the other 3d models are stored so all assets are colocated. Do they have more human readable names?

src/shader/shaderLightSampling.js
```diff
@@ -1,11 +1,51 @@
...
+		// based upon Frostbite 3 Moving to Physically-based Rendering
+		// page 32, equation 26: E[window1]
+		// https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf
+		float distanceFalloff = 1.0 / max( pow( lightDistance, decayExponent ), EPSILON );
+
+		if ( cutoffDistance > 0.0 ) {
+			distanceFalloff *= pow2( saturate( 1.0 - pow4( lightDistance / cutoffDistance ) ) );
+		}
+
+		return distanceFalloff;
```
gkjohnson, line 20: Looks like some off indentation here

src/shader/shaderLightSampling.js
```diff
@@ -1,11 +1,51 @@
...
+}
+
+// float atan2(float y, float x) {
+
+//     bool s = ( abs( x ) > abs( y ) );
+//     return mix( PI / 2.0 - atan( x, y ), atan( y, x ), s );
+
+// }
```
gkjohnson, line 28: Is this commented code still needed?

src/shader/shaderLightSampling.js
```diff
@@ -1,11 +1,51 @@
...
+float getPhotometricAttenuation(int iesProfile, vec3 posToLight, vec3 lightDir, vec3 u, vec3 v) {
+    float cosTheta = dot(-posToLight, lightDir);
+    float angle = acos(cosTheta) * (1.0 / PI);
+
+	//float y = dot( posToLight, u ); // TODO: use azimuth if any IES profiles require this
+	//float x = dot( posToLight, v );
+	//float a2 = atan2( y, x );
+
+    return texture2D( iesProfiles, vec3( 0.0, angle, iesProfile ) ).r;
+}
```
gkjohnson, line 39: Similar to `readMaterialInfo` can we pass the iesTexture into this function? If possible I'd like to keep these utility functions in other files from reading uniforms directly so they can more easily be reused if needed.

src/shader/shaderLightSampling.js
```diff
@@ -27,20 +67,50 @@ LightSampleRec lightsClosestHit( sampler2D lights, uint lightCount, vec3 rayOrig
...
-		u *= 1.0 / dot(u, u);
-		v *= 1.0 / dot(v, v);
+		u *= 1.0 / dot( u, u );
+		v *= 1.0 / dot( v, v );
 
 		float dist;
-		if ( intersectsRectangle( light.position, normal, u, v, rayOrigin, rayDirection, dist ) ) {
 
-			if ( dist < lightSampleRec.dist || !lightSampleRec.hit ) {
+		if( light.type == 0 ) {
```
gkjohnson, line 75: Can we add a comment denoting which light is associated with each of these conditions?

src/shader/shaderLightSampling.js
```diff
@@ -27,20 +67,50 @@ LightSampleRec lightsClosestHit( sampler2D lights, uint lightCount, vec3 rayOrig
 			continue;
 		}
 
-		u *= 1.0 / dot(u, u);
-		v *= 1.0 / dot(v, v);
+		u *= 1.0 / dot( u, u );
+		v *= 1.0 / dot( v, v );
```
gkjohnson, line 71: Are these variables and the normal backface check above still required for spot lights? Or should they be moved into the "Area Light" condition?

src/shader/shaderLightSampling.js
```diff
@@ -27,20 +67,50 @@ LightSampleRec lightsClosestHit( sampler2D lights, uint lightCount, vec3 rayOrig
...
+		} else {
+
+			SpotLight spotLight = readSpotLightInfo( lights, l );
+
+			float angle = acos( spotLight.coneCos );
+			float angleTan = tan( angle );
+			float startDistance = spotLight.radius / angleTan;
+
+			vec3 lightPosition = light.position - normal * startDistance;
+			if ( spotLight.lampIntensityScale > 0.0 && spotLight.radius > 0.0 && intersectsCircle( lightPosition, normal, u, v, spotLight.radius * 2.0, rayOrigin, rayDirection, dist ) ) {
```
gkjohnson, line 101: Can you describe what's happening here? Is the spotlight being modeled as a disk here?

src/shader/shaderLightSampling.js
```diff
@@ -73,14 +143,67 @@ LightSampleRec randomRectAreaLightSample( Light light, vec3 rayOrigin ) {
...
 	Light light = readLightInfo( lights, l );
 
 	// sample the light
-	return randomRectAreaLightSample( light, rayOrigin );
+	if( light.type == 0 )
+
+		// rectangular area light
+		return randomRectAreaLightSample( light, rayOrigin );
+
+	else {
```
gkjohnson, line 200: nit: Can we include braces here

src/uniforms/IESProfilesTexture.js
```diff
@@ -0,0 +1,127 @@
...
+					.load( iesProfileURL, texture => {
+
+						textures.push( texture );
+						resolve();
+
+					} );
+
+			} ) );
+
+		} );
```
gkjohnson, line 62: Similar to [RenderTarget2DArray](https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/src/uniforms/RenderTarget2DArray.js) I think this class should keep the loading of IES profile textures separate from the initializing the texture itself. It might actually be cleanest for this class to extend `RenderTarget2DArray` since it looks like there's a lot of similar code.

example/spotLights.js
```diff
@@ -0,0 +1,753 @@
...
+
+				spotLight.position.set( i * 8, 7.0, 0.005 );
+				spotLight.angle = Math.PI / 8.0;
+				spotLight.penumbra = 0.0;
+				spotLight.decay = decays[ i ];
+				spotLight.distance = 0.0;
+				spotLight.intensity = 50.0;
+				spotLight.lampIntensityScale = 0.0;
+				spotLight.radius = 0.5;
+				spotLight.iesProfile = - 1 + i;
```
gkjohnson, line 289: Similar to the `PhysicalCamera` implementation that includes DoF parameters I'm wondering if it makes sense to include a `PhysicalSpotlight` class that includes IES and radius parameters.

src/shader/shaderLightSampling.js
```diff
@@ -27,20 +67,50 @@ LightSampleRec lightsClosestHit( sampler2D lights, uint lightCount, vec3 rayOrig
...
+			vec3 lightPosition = light.position - normal * startDistance;
+			if ( spotLight.lampIntensityScale > 0.0 && spotLight.radius > 0.0 && intersectsCircle( lightPosition, normal, u, v, spotLight.radius * 2.0, rayOrigin, rayDirection, dist ) ) {
+
+				if ( dist < lightSampleRec.dist || !lightSampleRec.hit ) {
+
+					lightSampleRec.hit = true;
+					lightSampleRec.dist = dist;
+					float cosTheta = dot( rayDirection, normal );
+					lightSampleRec.pdf = ( dist * dist ) / ( light.area * cosTheta );
+					lightSampleRec.emission = light.color * light.intensity * spotLight.lampIntensityScale;
```
gkjohnson, line 109: I think it's okay if we skip the `lampIntensityScale` factor for now - with this disabled the lighting does not work in the non MIS case or the primary path tracing. I think we'd want a more general approach that works for disabling both area lights and spot light surfaces -- I'll make an issue for it.

gkjohnson (commented):
Beautiful stuff! I see now how non-MIS sampling _should_ work but only if the spotlight "radius" is larger than 0. I've made a few comments and I think the last thing here is updating the README with info on the new lighting support!

example/spotLights.js
richardassar: I picked the ones that I thought looked nice from [ieslibrary](https://ieslibrary.com/en/home). The file itself has some fields which describe the manufacturer and product name, which we could expose via the `IESLoader` so that drop-downs can be populated.

As part of making a scene demonstrating spotlights I could download a new set and note down their names from the site (they annoyingly come with those hashed filenames).

src/shader/shaderLightSampling.js
richardassar: Got it, it's a shame we can't lint these also. If they were loaded as files (with some mechanism for includes / templating) that could prevent these from making it past the build tests. Not sure if it's worth it.

src/shader/shaderLightSampling.js
richardassar: It is if we want to use the other dimension IESLoader provides, most profiles don't end up using it but one (I forget which) from the set in the `iesProfiles` array does. 

The profiles are 360x180 and repeat the rows anyway, so we can use this and just pay the extra cost. It's a case of deciding whether it's worth it, if we decide to stick with 1D profiles then I don't need to use a texture array.

src/shader/shaderLightSampling.js
richardassar: The lamp itself is being modeled as a circular area light.

src/shader/shaderLightSampling.js
richardassar: Yes, I think defining them as `LIGHT_TYPE_AREA`, `LIGHT_TYPE_SPOTLIGHT` and setting these as defines would be nice. Similarly I think the same treatment for texture strides (such as `MATERIAL_PIXELS` should also be passed in this way, avoiding the need to edit it in two places).

example/spotLights.js
richardassar: That sounds good.

src/shader/shaderLightSampling.js
richardassar: I didn't like this concept of "lamp" here but I put it in to be more physically accurate, this probably goes against how most scenes are built hence why I opted to make this feature an optional / convenience thing. In actuality people probably would use a textured and emissive surface, say for a theatre lamp.

src/uniforms/IESProfilesTexture.js
richardassar: I think an issue there was that `RenderTarget2DArray` defaults to byte type and does some creation in the constructor, but I can always replace the constructor. That's still not very nice but is better than what I have here.

example/spotLights.js
gkjohnson: Oh that would be a good link to include the docs / comment. If you have a list of good ones you remember I can upload them otherwise I can pick a few and modify the example with some later.

src/shader/shaderLightSampling.js
gkjohnson: Ideally there'd be some way to lint content in these template tags - I'll have to look around for something like that. Worst cast we _could_ write a script to pull out the content from these strings, write it to a file, and lint it there but there'd be a lot of indirection in the results...

src/shader/shaderLightSampling.js
gkjohnson: Can you elaborate? I might be misunderstanding. Can't we adjust the function signature to the following?

```glsl
float getPhotometricAttenuation( sampler2DArray iesTexture, int iesProfile, vec3 posToLight, vec3 lightDir, vec3 u, vec3 v ) {
```

src/uniforms/IESProfilesTexture.js
gkjohnson: It should be okay to override the defaults again for `format`, `type`, etc in constructor of `IESProfilesTexture`.

src/shader/shaderLightSampling.js
gkjohnson: I think one of the disconnects here is that `lampIntensityScale` is actually attenuating the sample intensity, as well, whereas to get the effect you're talking about you really just want it to scale the intensity if the first ray (or maybe specular ray) hits it directly if I'm understanding right.

See the difference in effect here with Non MIS rendering:

| 1.0 lampIntensityScale | 0.1 lampIntensityScale  |
|---|---|
| image | image |

I agree with you that there's value in being able to hide the light surface but I think it should probably just be a boolean and work for all lights so I think it's okay if that comes later. I added #193 to track it.

src/shader/shaderLightSampling.js
richardassar: Apologies, I wasn't looking at the code.

> Can't we adjust the function signature to the following? ...

That's how I'd interpreted your initial point.  

We can thread the `iesTexture` through by passing it to `randomLightSample`, in some applications you can apply them to any light type so it isn't inappropriate to pass it to this function.

src/shader/shaderLightSampling.js
mrxz: Is it desirable to have spotlights being modelled as a circular area? I think having them be punctual lights is easier and that is also in line with both glTF and Three.JS. The biggest downside of course is that they won't work in the non MIS case.

gkjohnson: By default the spot lights are modeled as a single point but optionally you can adjust the "radius" of the spotlight surface (default being 0) to model it as a disc which provides the option for softer shadows. So everything should still match three.js with the option of higher fidelity options that three.js can't provide. In the future this might be a nice way to support point lights / sphere lights with a single implementation.

| radius = 0.0 | radius = 0.5 |
|---|---|
| image | image |

gkjohnson:
It actually does look like there's a difference in how MIS and non MIS rendering converges which points to an issue in how rays hit the surface of the light are handled and impact the way MIS specular rays re handled, as well:

| MIS | non MIS |
|---|---|
| image | image |

I think what's missing is the emission factors (ie profile, penumbra, cone, etc) being accounted for in [this code in the lightsClosestHit function](https://github.com/gkjohnson/three-gpu-pathtracer/blob/a50c9ca43d74f8778d4931b0ada4c76fb493fb6e/src/shader/shaderLightSampling.js#L93-L115)

richardassar:
This was my conclusion also, the current approach is taking energy from the spotlight itself. I'll take another look at this.

Any thoughts on https://github.com/gkjohnson/three-gpu-pathtracer/pull/184#issuecomment-1155833291 ? Sometimes similar things can occur with NEE (next event estimation).

gkjohnson:
> Any thoughts on https://github.com/gkjohnson/three-gpu-pathtracer/pull/184#issuecomment-1155833291 ? Sometimes similar things can occur with NEE (next event estimation).

I've been thinking about this and I think the lack reflected mirror light is from two reasons:

1. A non zero `filterGlossyFactor` is causing the subsequent reflected mirror rays to behave as diffuse rays rather than specular ones. The factor is supposed to alleviate cases where you have a rare high contribution ray that hits a diffuse surface, then specular, then bright light but this would be a case where it causes issues.

2. When the spot light is modeled as a point it's nearly impossible to randomly find a ray that scatters from the diffuse floor surface and hits the mirror such that it perfectly reflects at the angle required to get spotlight contributions. Effectively this mirror case uses the non-MIS sampling path so once https://github.com/gkjohnson/three-gpu-pathtracer/pull/184#issuecomment-1166329200 is addressed we should see reflected light on the ground when the light radius is non zero.

I'm not sure how to get reflected light to work with MIS sampling, though. Effectively you need to be able to choose floor diffuse rays that are directed toward the mirror in a way that they'll hit the spot light point. Perhaps there are other ways to handle this? @erlichof I'm not sure if you have any insight here?

bhouston:
Even if this doesn't yet support mirrors in MIS mode, maybe we should merge it anyways?  And allow others to also contribute?  I am just worried that leaving this in PR mode like this for a while can wreck momentum.  I believe lights mostly work otherwise right?

mikelyndon:
I'd love to see this merged. So many existing glTF scenes have spotlights in them that even an implementation with caveats would be more useful than not.

richardassar:
I'll bring this up to date with main and see if I can improve the implementation.

richardassar:
I've updated this PR. Separated the spot light sampling from the env and area light sampling, added circular area lights but in the area lights code path. The MIS and non-MIS case for spot lights no longer show the discrepancy (which was no spot lights in non-MIS mode) any more.

**MIS**
Screenshot from 2022-07-25 12-12-13
(Low resolution scale so that I could collect enough samples in non-MIS mode)

**Non-MIS**
Screenshot from 2022-07-25 12-12-03
(Low resolution scale so that I could collect enough samples in non-MIS mode)

**Circular Area Lights**
Screenshot from 2022-07-25 13-33-54

I can squash this PR's commits so that these changes are collected into a single commit if that's preferred, just let me know.

gkjohnson:
Oh man this is such a great set of additions! Really looking forward to this one. I'll take a deeper look again this week and get this merged. Thanks again for putting in all the effort for this!

richardassar:
You're welcome, no rush, I'll continue working on the ODIN denoiser now and respond to any feedback on this.

gkjohnson:
This isn't an issue with this PR but it's something that came up on my M1 Mac while testing - I get the following error when running examples in this branch:

> `Program Info Log: FRAGMENT shader texture image units count exceeds MAX_TEXTURE_IMAGE_UNITS(16)`

Here's the textures I can see when using Spector.js:

```
1. normalAttribute
2. tangentAttribute
3. uvAttribute
4. materialIndexAttribute
5. bvh.index
6. bvh.position
7. bvh.bvhBounds
8. bvh.bvhContents
9. materials
10. lights
11. spotlights
12. envMapInfo.marginalWeights
13. envMapInfo.conditionalWeights
14. envMapInfo.map
15. envMapInfo.totalSum
16. textures
17. iesTextures
```

So we're now one over our limit. I'll try to figure out an easy way around this in the short term but here are a couple things we could consider longer term. At this point I don't expect more data textures to be required, though, so I don't think this will be a longer standing issue:

1. Simplify the environment map info struct - specifically the need for a `envMapInfo.totalSum` texture to work around struct precision issues.
1. Consolidate area light and spotlight textures into a single data texture
1. Merge `iesTextures` into the `textures` sampler array (if possible)
1. Pack floating point vertex attribute data into a common sampler array
1. Pack floating point bvh data into a common sampler array / sampler

The first point seems like the most workable in the short term. I'll take some time to think it through. I'm not sure if you have any other thoughts on this. I've made #223 to track it in the long term.

gkjohnson:
I made a quick fix as a PR into this branch here:

https://github.com/richardassar/three-gpu-pathtracer/pull/1

gkjohnson:
And something else I noticed that seemingly only happens when "isCircular" is enabled - the lighting shows through the bottom of the floor:

**Below the Floor**
image

**Above the Floor**
image

Sorry for rapid fire comments - I'll keep an eye out for what could cause this when reviewing.

src/shader/shaderLightSampling.js
```diff
@@ -73,14 +110,103 @@ LightSampleRec randomRectAreaLightSample( Light light, vec3 rayOrigin ) {
...
 }
 
+LightSampleRec randomCircularAreaLightSample( Light light, vec3 rayOrigin ) {
+
+	LightSampleRec lightSampleRec;
+	lightSampleRec.hit = true;
+
+	lightSampleRec.emission = light.color * light.intensity;
+
+	float r = 2.0 * sqrt( rand() );
```
gkjohnson, line 120: I think this should be `0.5 * sqrt( rand() )` since theta is sampled from the full `2 * PI` range and the width / height represent the full width and full height of the light, not just the radius size.

src/shader/shaderLightSampling.js
```diff
@@ -73,14 +110,103 @@ LightSampleRec randomRectAreaLightSample( Light light, vec3 rayOrigin ) {
...
+	return lightSampleRec;
+
+}
+
+LightSampleRec randomSpotLightSample( SpotLight spotLight, sampler2DArray iesProfiles, vec3 rayOrigin ) {
+
+	LightSampleRec lightSampleRec;
+	lightSampleRec.hit = true;
+
+	float r = 2.0 * spotLight.radius * sqrt( rand() );
```
gkjohnson, line 145: I haven't tried to look for any issues with this visually but is it correct that this should be multiplied by 2.0, as well? Assuming that the radius of the spot lights directly represents the radius of a disc then shouldn't this be multiplied by 1.0 instead of 2.0 since theta is sampled from [0, 2 PI]?

src/shader/shaderLightSampling.js
richardassar: Checked this, it resolves both the light bleed and MIS/non-MIS discrepancy. Indeed the width/height representation requires we are using [-0.5, 0.5].

src/shader/shaderLightSampling.js
richardassar: This slipped through when experimenting, removing in the next push.

richardassar:
I've merged your texture fix PR and fixed those incorrect circle sampling functions.

I just merged `randomRectAreaLightSample` and `randomCircularAreaLightSample` as they shared a lot of common code, the area light type branches in `randomAreaLightSample` to select the computation of `randomPos`.

Also fixed a divide by zero when the spot light angle goes to zero (which was crashing my tab in Chromium).

Ready for another review pass.

gkjohnson:
Excellent! Looks great. Thank you again!

gkjohnson:
I made issues #226, #227, #228 with some small related follow-on tasks for anyone to take on

richardassar:
You're welcome :+1: 

One thing I wanted to reinstate was the azimuth angle in `getPhotometricAttenuation` but since it is used so rarely and I didn't have a good test case I decided against doing so. 

I did find a 2D IES profile but without any reference imagery there was no ground truth to compare against. Perhaps https://cndl.io/ or similar tools could help here.

This means that the array texture is overkill and we can just use a single texture where each row/column is a different 1D profile, perhaps separating 2D profiles into their own array texture or using the texture atlas with correct clamping/wrapping.

gkjohnson:
Great - I made issue #233 to track it for the future in case anyone wants to help out!

gkjohnson:
I also moved the IES profiles to my demo data repo with https://github.com/gkjohnson/three-gpu-pathtracer/commit/5181077e632219562a1deb136c65359aca77c86f just FYI.

PR #189, opened 2022-06-18
Update arealight demo

gkjohnson:
Updates to the area light demo with more controls and loaded model.

174458617-8e5d5e37-80cf-47df-a1c9-9d65aec21322

174458720-9fd0d21d-d03d-452b-b360-e7baf1f9adff

cc @mrxz in case you're interested!


mrxz:
That looks really nice! :+1:

Issue #85, opened 2022-04-08
Add denoiser


gkjohnson:
https://github.com/BrutPitt/glslSmartDeNoise

DennisSmolek:
The other day I was doing something trying to learn more about simulating clouds and I saw a promotion/release from Nvidia about their raytracing and it showed how they used denoise to reduce the sampling by at least half for even huge complex stuff. 

And that form most workflows, it was like the last node in the pipeline for super smooth results. 

I love your raytracer but I'm on a potato.. I can get to about 1000 passes but it still looks grainy... So I had shelved being able to use it personally any time soon.

I tried to see if there was a standard few sets of algorithms like there is for GENERATING noise, and it seems like theres some old school fundamentals but theres also a lot of crazy innovative things coming out all the time.. I tried limiting to things published within the last 3 years...

Of course found a big 0 for the WebGL space, but Nvidia and others are pushing each other. 

This paper came up a few times in discussions: https://cs.dartmouth.edu/wjarosz/publications/mara17towards.pdf

But when my eyes hit the math on page two:
<img width="825" alt="Screen Shot 2022-04-09 at 06 31 04" src="https://user-images.githubusercontent.com/1397052/162571045-4c19072f-9bfe-4e47-9745-f9e211066ee9.png">

I figured it reach out first haha..

Right before posting I found: https://alain.xyz/blog/ray-tracing-denoising

Which I didn't want to dive into if you had a solution in mind or strong opinions or something... 

After our tweets I'm going to go over that article, I pulled some others too. 

Dont search "Denoise" on shader tree.. .It repeatedly crashed my mac 100%, not just the browser..

There are 20-30 there too, that hopefully are implementations I can try

gkjohnson:
> I love your raytracer but I'm on a potato.. I can get to about 1000 passes but it still looks grainy...

Ha some day I'd like this to work well even on a potato but it's still early. There's still a lot of work to do to improve the speed of convergence including multiple importance sampling, direct light sampling, and denoising. But I'd like help getting all of it going so I appreciate you taking a look into this a bit!

> Which I didn't want to dive into if you had a solution in mind or strong opinions or something...

[...]

PR #194, opened 2022-06-25
Added denoiser https://github.com/BrutPitt/glslSmartDeNoise #85

richardassar:
Solves issue https://github.com/gkjohnson/three-gpu-pathtracer/issues/85

Added support for denoising with glslSmartDeNoise. Tuned the parameters a bit to find a good balance without too much blur.

Initially I tried adding `DenoiseMaterial` to the main `fsQuad` but due to how THREE.js handles tone mapping this broke tone mapping, so it has to go into `PathTracingRenderer`.

When additional denoisers are added these could be set by changing the `ptRenderer.denoiser` material.

I've added this only to `materialBall.js` and it defaults to enabled, with the slider defaulting to enabled also, we can turn this off before merging the PR.

### DIFF
--- a/example/materialBall.js
+++ b/example/materialBall.js
@@ -51,6 +51,12 @@ const params = {
 
 	multipleImportanceSampling: true,
 	stableNoise: false,
+	denoiseEnabled: true,
+	denoiseSliderEnabled: true,
+	denoiseSlider: 0.0,
+	denoiseSigma: 2.5,
+	denoiseThreshold: 0.1,
+	denoiseKSigma: 1.0,
 	environmentIntensity: 1,
 	environmentRotation: 0,
 	environmentBlur: 0.0,
@@ -116,6 +122,8 @@ async function init() {
 	ptRenderer.material.setDefine( 'TRANSPARENT_TRAVERSALS', params.transparentTraversals );
 	ptRenderer.material.setDefine( 'FEATURE_MIS', Number( params.multipleImportanceSampling ) );
 	ptRenderer.tiles.set( params.tiles, params.tiles );
+	ptRenderer.denoise = params.denoiseEnabled;
+	ptRenderer.denoiser.setDefine( 'USE_SLIDER', Number( params.denoiseSliderEnabled ) );
 
 	fsQuad = new FullScreenQuad( new THREE.MeshBasicMaterial( {
 		map: ptRenderer.target.texture,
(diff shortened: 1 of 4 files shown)


richardassar:
Alternatively we could apply tonemapping ourselves, which would mean we can avoid touching the `PathTracingRenderer` here, but that doesn't seem like a lesser of two evils.

gkjohnson:
Awesome! Looks like it works best on low frequency noise but it's a good start. I think we can embed to tonemapping in the denoising shader to avoid touching the PathTracingRenderer which seems like a nice separation to me. We can add [these lines](https://github.com/mrdoob/three.js/blob/e9b9783c743688ded2c0987751b420b83581c7f7/src/renderers/shaders/ShaderLib/meshbasic.glsl.js#L109-L110) from MeshBasicMaterial to the end of the shader main function to apply them:

```glsl
	#include <tonemapping_fragment>
	#include <encodings_fragment>
```

Also I'm not sure how easy this is to address but the shader seems to corrupt some of the background transparency effects:

| no denoise | denoise |
|---|---|
| image | image |

It looks like maybe the alpha channel isn't being accounted for in the denoiser and the dark bits are becoming extra blurry? To handle alpha cleanly I also had to use an alpha weighted color blend (see [here](https://github.com/gkjohnson/three-gpu-pathtracer/blob/2852e9e8dacb311d6288197abf1114275db65d53/src/materials/BlendMaterial.js#L50-L51))  to avoid artifacts. Not sure if something like that would be needed here?

richardassar:
> Awesome! Looks like it works best on low frequency noise but it's a good start. I think we can embed to tonemapping in the denoising shader to avoid touching the PathTracingRenderer which seems like a nice separation to me. We can add [these lines](https://github.com/mrdoob/three.js/blob/e9b9783c743688ded2c0987751b420b83581c7f7/src/renderers/shaders/ShaderLib/meshbasic.glsl.js#L109-L110) from MeshBasicMaterial to the end of the shader main function to apply them:
> 
> ```glsl
> 	#include <tonemapping_fragment>
> 	#include <encodings_fragment>
> ```
Thanks, I'll give this a shot.

> Also I'm not sure how easy this is to address but the shader seems to corrupt some of the background transparency effects:
> ...
> It looks like maybe the alpha channel isn't being accounted for in the denoiser and the dark bits are becoming extra blurry? To handle alpha cleanly I also had to use an alpha weighted color blend (see [here](https://github.com/gkjohnson/three-gpu-pathtracer/blob/2852e9e8dacb311d6288197abf1114275db65d53/src/materials/BlendMaterial.js#L50-L51)) to avoid artifacts. Not sure if something like that would be needed here?

Indeed, [here](https://github.com/gkjohnson/three-gpu-pathtracer/pull/194/commits/15126b22183a4d73338712624f40f571685e9010#diff-05b2931c1b18442b3e6e54f2b2ff5ae4019a73a630619d96b498cb444e4c7673R114) is just referencing `rgb` only, I'll see if modifying this to include the alpha channel fixes the issue.

As a next phase to this it could be interesting to detect depth discontinuities and use that to avoid blurring out edges too much, of course that doesn't handle everything but it might help balance things a little better. Just an idea.

richardassar:
I adjusted `DenoiseMaterial` to include the alpha channel when computing the exponent term of the Gaussian kernel, this seems remove the bias that caused the difference in blur you reported [here](https://github.com/gkjohnson/three-gpu-pathtracer/pull/194#issuecomment-1166684586).

I now also pre-multiply the color channels with the alpha channel which solved the issue shown below. Without pre-multiplying in the bottom right you see some colors bleeding in from the background.

| Without alpha pre-multiplied | With alpha pre-multiplied |
| ----------- | ----------- |
| Screenshot 2022-06-27 at 08-34-02 Material Orb Path Tracing | Screenshot 2022-06-27 at 08-34-09 Material Orb Path Tracing |

I tried only modifying the exponent, only pre-mulitplying and with both, it seems that both are required for the best results.

gkjohnson:
Great! Looks much better.

> As a next phase to this it could be interesting to detect depth discontinuities and use that to avoid blurring out edges too much, of course that doesn't handle everything but it might help balance things a little better. Just an idea.

It looks like the [SVGF](https://teamwisp.github.io/research/svfg.html) denoiser uses normals and depth to help guide the denoiser. I've used normals / depth buffers to help detect discontinuities in surfaces for screen space effects before but typically those work best with non-AA buffers. I haven't taken a deep dive into SVGF but I wonder how it works with anti aliasing?

Issue #195, opened 2022-06-25
Normal maps should use the diffuse texture.repeat


richardassar:
**Describe the bug**

We currently set the `normalMapTransform` from the `normalMap`, which results in incorrect rendering.

As described on this [discourse.threejs.org](https://discourse.threejs.org/t/different-repeats-for-map-and-bumpmap-textures/9885/2) post THREE.js normal maps should repeat with the diffuse map, their `texture.repeat` is ignored.

**Expected behavior**

The `normalMap` UVs should be scaled with the `map`'s `texture.repeat` and the `normalScale` value only.

**Screenshots and Repro Model**

three-gpu-pathtracer shows this:
![Screenshot from 2022-06-25 09-33-12](https://user-images.githubusercontent.com/868396/175766795-8a47ef9d-62fb-4b7a-ab37-b2e54cb842f4.png)

THREE.js shows this:
![Screenshot from 2022-06-25 09-34-25](https://user-images.githubusercontent.com/868396/175766714-299df4bb-0598-4fc0-a9d2-b12d687e6bac.png)

I confirmed the behavior by showing the underlying THREE.js render and setting:
```
diffuse.repeat.x = 20;
diffuse.repeat.y = 20;
```
and
```
normalMap.repeat.x = 1.0/20;
normalMap.repeat.y = 1.0/20;
```
which resulted in this, showing that `normalMap.repeat` is in fact ignored:

![Screenshot from 2022-06-25 10-35-19](https://user-images.githubusercontent.com/868396/175767764-caf25839-24f2-419e-84ef-0ac5f1a407f2.png)

Setting `normalScale` on the material results in the expected scaling.

I have a fix ready for this which I am about to push and make a PR.

**Platform:**

 - Device: Any/All
 - OS: Any/All
 - GPU: Any/All
 - Browser: Any/All
 - Three.js version: Any/All
 - Library version: Any/All

richardassar:
Created an issue so that there's an easily searchable record of this.

gkjohnson:
[...]

PR #196, opened 2022-06-25
Fix normapMap texture.repeat

richardassar:
Resolves https://github.com/gkjohnson/three-gpu-pathtracer/issues/195

- Transform `normalMap` by the `mapTransform` and apply `normalScale`

Behaviour is confirmed on [discourse.threejs.org](https://discourse.threejs.org/t/different-repeats-for-map-and-bumpmap-textures/9885/2) posts, from test and from consulting the THREE.js source code.

### DIFF
--- a/src/materials/PhysicalPathTracingMaterial.js
+++ b/src/materials/PhysicalPathTracingMaterial.js
@@ -610,7 +610,7 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 								vec3 bitangent = normalize( cross( normal, tangent ) * tangentSample.w );
 								mat3 vTBN = mat3( tangent, bitangent, normal );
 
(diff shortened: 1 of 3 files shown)


bhouston:
Very nice!  We should get this merged. :)

gkjohnson:
@bhouston see https://github.com/gkjohnson/three-gpu-pathtracer/issues/195#issuecomment-1166337295. If we want to add the ability for texture transforms to match the three.js behavior I think we can add a flag to the `MaterialsTexture` that triggers all textures to use the same offset. Something like `MaterialsTexture.shareTextureOffsets` that defaults to false.

bhouston:
This makes sense.  Or even MaterialsTexture.threeJsCompatibilityMode = true. and document that it is about limitations ThreeJS has with regards to texture channels.

Issue #168, opened 2022-06-03
Clear Coat material property


bhouston:
**Is your feature request related to a problem? Please describe.**

In adhering to the goal of reproducing the full glTF PBR Next model, one of the currently missing material properties is clear coat.  It is primarily used for lacquers, varnishes as well as the clear coat on cars or on ceramics.

**Describe the solution you'd like**

![khr-materials-clearcoat](https://user-images.githubusercontent.com/588541/174065969-5df8f8fe-b226-4821-a6bc-c3ddb13fc73a.png)

Parameters and behavior should follow glTF KHR_materials_clearcoat extension: 

https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_clearcoat 

The following parameters are contributed by the `KHR_materials_clearcoat` extension:

|                                  | Type                                                                            | Description                            | Required             |
|----------------------------------|---------------------------------------------------------------------------------|----------------------------------------|----------------------|
|**clearcoatFactor**               | `number`                                                                        | The clearcoat layer intensity.         | No, default: `0.0`   |
|**clearcoatTexture**              | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo)             | The clearcoat layer intensity texture. | No                   |
|**clearcoatRoughnessFactor**      | `number`                                                                        | The clearcoat layer roughness.         | No, default: `0.0`   |
|**clearcoatRoughnessTexture**     | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo)             | The clearcoat layer roughness texture. | No                   |
|**clearcoatNormalTexture**        | [`normalTextureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-material-normaltextureinfo) | The clearcoat normal map texture.      | No                   |

bhouston:
I am pleased to be able to put a bounty of $250 USD on an accepted PR that implements the official glTF clear coat extension in this path tracer.

richardassar:
Taking a look at this also, gathered some resources, existing implementations etc.

richardassar:
Here's an update on this. 

[...]

PR #203, opened 2022-07-08
Implemented clearcoat material property

richardassar:
Solves https://github.com/gkjohnson/three-gpu-pathtracer/issues/168

Screenshot from 2022-07-08 09-56-17


gkjohnson:
This looks really great - nice work!

And thanks for splitting the new material properties out separately. It makes it a lot easier to follow.

Issue #167, opened 2022-06-03
Sheen material property


bhouston:
**Is your feature request related to a problem? Please describe.**

In adhering to the goal of reproducing the full glTF PBR Next model, one of the currently missing material properties is sheen.  It is primarily used for cloth.

![cushion](https://user-images.githubusercontent.com/588541/171766562-79820feb-1d2d-45b5-9f8e-d898dbf994c6.png)

**Describe the solution you'd like**

Parameters and behavior should follow glTF KHR_materials_sheen extension: 

https://github.com/KhronosGroup/glTF/tree/a9cd1cb657587b590b6f95c8654038a4e9d66596/extensions/2.0/Khronos/KHR_materials_sheen 

The properties required to support are:

|                                  | Type                                                                            | Description                            | Required                       |
|----------------------------------|---------------------------------------------------------------------------------|----------------------------------------|--------------------------------|
|**sheenColorFactor**                   | `array`                                                                         | The sheen color in linear space        | No, default: `[0.0, 0.0, 0.0]` |
|**sheenColorTexture**         | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The sheen color (RGB).<br> The sheen color is in sRGB transfer function | No               |
|**sheenRoughnessFactor**               | `number`                                                                        | The sheen roughness.                   | No, default: `0.0`             |
|**sheenRoughnessTexture**         | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The sheen roughness (Alpha) texture. | No               |

bhouston:
I am pleased to be able to put a bounty of $500 USD on an accepted PR that implements the official glTF sheen extension in this path tracer.

richardassar:
Taking a look at this, I recall the paper [A Practical Microcylinder Appearance Model for Cloth Rendering](http://graphics.ucsd.edu/~henrik/papers/practical_microcylinder_appearance_model_for_cloth_rendering.pdf) which may or may not be useful here.

richardassar:
Results from a quick first pass at this. 

**Highlights at grazing angles:**
![Screenshot 2022-06-23 at 00-01-46 Material Orb Path Tracing](https://user-images.githubusercontent.com/868396/175168821-8e4fde49-dc2d-4696-b045-eb2c7ac6689c.png)

**Custom sheen color:**
[...]

PR #205, opened 2022-07-10
Implemented sheen material property

richardassar:
Solves https://github.com/gkjohnson/three-gpu-pathtracer/issues/167

Screenshot from 2022-07-10 16-03-13

### DIFF
--- a/src/shader/shaderMaterialSampling.js
+++ b/src/shader/shaderMaterialSampling.js
@@ -1,4 +1,5 @@
 import { shaderGGXFunctions } from './shaderGGXFunctions.js';
+import { shaderSheenFunctions } from './shaderSheenFunctions.js';
 
 export const shaderMaterialSampling = /* glsl */`
 
@@ -16,6 +17,8 @@ struct SurfaceRec {
 	float clearcoat;
 	float clearcoatRoughness;
 	float filteredClearcoatRoughness;
+	vec3 sheenColor;
+	float sheenRoughness;
 };
 
 struct SampleRec {
@@ -27,6 +30,7 @@ struct SampleRec {
 };
 
 ${ shaderGGXFunctions }
+${ shaderSheenFunctions }
 
 // diffuse
 float diffusePDF( vec3 wo, vec3 wi, SurfaceRec surf ) {
@@ -272,6 +276,28 @@ void clearcoatColor( inout vec3 color, vec3 wo, vec3 wi, SurfaceRec surf ) {
 
 }
 
+// sheen
+vec3 sheenColor( vec3 wo, vec3 wi, SurfaceRec surf ) {
+
+	vec3 halfVector = getHalfVector( wo, wi );
+
+	float cosThetaO = saturateCos( wo.z );
+	float cosThetaI = saturateCos( wi.z );
+	float cosThetaH = halfVector.z;
+
+	float D = velvetD( cosThetaH, surf.sheenRoughness );
+	float G = velvetG( cosThetaO, cosThetaI, surf.sheenRoughness );
+
+	// See equation (1) in http://www.aconty.com/pdf/s2017_pbs_imageworks_sheen.pdf
+	vec3 color = surf.sheenColor;
+	color *= D * G / ( 4.0 * abs( cosThetaO * cosThetaI ) );
+	color *= wi.z;
+
+	return color;
+
+}
+
+// bsdf
 void getLobeWeights( vec3 wo, vec3 clearcoatWo, SurfaceRec surf, out float diffuseWeight, out float specularWeight, out float transmissionWeight, out float clearcoatWeight ) {
 
 	float ior = surf.ior;
@@ -399,6 +425,9 @@ vec3 bsdfColor( vec3 wo, vec3 clearcoatWo, vec3 wi, vec3 clearcoatWi, SurfaceRec
 			color += specularColor( wo, wi, surf );
 
 		}
+
+		color *= sheenAlbedoScaling( wo, wi, surf );
+		color += sheenColor( wo, wi, surf );
 
 	}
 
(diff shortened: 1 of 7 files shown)


gkjohnson:
Awesome! I'll take a closer look at the code when I get a chance this next week. Here are a couple screenshots from the new "viewerTest" demo with some better glTF sample models. It looks like there's kind of error on the edge of the cloth in the toy car scene:

**khronos-SheenChair**

image

**khronos-toycar**

image

richardassar:
Thanks, I'll take a look at that. I noticed an issue when sheen and clear coat are looked at together so will resolve that and let you know.

richardassar:
Pushed fixes for this. Please take a look and let me know if you see anything else that needs addressing.

**Before**
Screenshot from 2022-07-11 13-16-42

**After**
Screenshot from 2022-07-11 13-17-11

src/shader/shaderMaterialSampling.js
```diff
@@ -272,6 +276,27 @@ void clearcoatColor( inout vec3 color, vec3 wo, vec3 wi, SurfaceRec surf ) {
...
+	float D = velvetD( cosThetaH, surf.sheenRoughness );
+	float G = velvetG( cosThetaO, cosThetaI, surf.sheenRoughness );
+
+	vec3 color = surf.sheenColor;
+	color *= D * G / ( 4.0 * abs( cosThetaO * cosThetaI ) );
+	color *= wi.z;
+
+	return color;
+
+}
```
gkjohnson, line 297: Is this function from a paper or anything? Is there a reference comment we could add?

src/shader/shaderSheenFunctions.js
```diff
@@ -0,0 +1,97 @@
...
+
+float sheenAlbedoScaling( vec3 wo, SurfaceRec surf ) {
+
+    float alpha = max( surf.sheenRoughness, 0.07 );
+    alpha = alpha * alpha;
+
+	float maxSheenColor = max( max( surf.sheenColor.r, surf.sheenColor.g ), surf.sheenColor.b );
+
+	float eWo = directionalAlbedoSheen( saturateCos( wo.z ), alpha );
+
```
gkjohnson, line 92: indentation issues 😞 Looks like this file is mostly using spaces for indentation rather than tabs like the other files.

src/shader/shaderSheenFunctions.js
```diff
@@ -0,0 +1,97 @@
...
+float velvetParamsInterpolate( int i, float oneMinusAlphaSquared ) {
+
+    const float p0[5] = float[5]( 25.3245, 3.32435, 0.16801, -1.27393, -4.85967 );
+    const float p1[5] = float[5]( 21.5473, 3.82987, 0.19823, -1.97760, -4.32054 );
+
+    return mix( p1[i], p0[i], oneMinusAlphaSquared );
+
+}
+
+float velvetL( float x, float alpha ) {
```
gkjohnson, line 26: Are there references for all these other sheen functions, as well, that we could add a comment for?

src/shader/shaderUtils.js
```diff
@@ -7,6 +7,14 @@ export const shaderUtils = /* glsl */`
 
 	}
 
+	float schlickFresnel( float u ) {
+
+		float m = clamp( 1.0 - u, 0.0, 1.0 );
+		float m2 = m * m;
+		return m2 * m2 * m;
+
+	}
```
gkjohnson, line 16: What's the difference between this function and the one above? Is this the same as calling the above function with `f0 === 1.0`?

gkjohnson (commented):
Looks great! Just a few comments on linting and it would be nice to document more information on the source of some of the functions.

src/shader/shaderMaterialSampling.js
richardassar: It is the `sheenColor * FS * GS * DS / (4 * abs(dot(N, L)) * abs(dot(N, V)))` from KHR_materials_sheen with the suggestion `The Fresnel term may be omitted, i.e., F = 1.` applied.

You can also check `sheen_layer` function in `sheen.glsl` in the `dspbr-pt` project which uses a similar formulation, just split out differently.

src/shader/shaderSheenFunctions.js
richardassar: I'd hoped vscode was smarter at picking up the project indentation style but it seems that it is not, I'll fix this.

src/shader/shaderSheenFunctions.js
richardassar: In `KHR_materials_sheen` the function `l` is implemented here with `velvetL` and `velvetParamsInterpolate` and `sheenDistribution` from `KHR_materials_sheen` is implemented in `velvetD`.

src/shader/shaderUtils.js
richardassar: It's the same as calling the above function with `f0 = 0.0` or calling `schlickFresnelFromIor` with an `iorRatio` of 1. This isn't actually used and I forgot to strip it out. Such a function is not necessary if constant propagation is applied during compilation.

src/shader/shaderSheenFunctions.js
richardassar: I'll search through the references on `KHR_materials_sheen` and annotate the functions with them with links / page number etc.

src/shader/shaderSheenFunctions.js
gkjohnson: I'm not sure if you have editorConfig installed but I _think_ that will also auto-configure indentation for template strings, as well.

richardassar:
I've fixed up the indentation and added references, please let me know if there's anything else that needs added or changed.

PR #206, opened 2022-07-12
Added support for iridescence

richardassar:
Solves https://github.com/gkjohnson/three-gpu-pathtracer/issues/96

Screenshot from 2022-07-12 16-19-31

### DIFF
--- a/src/shader/shaderIridescenceFunctions.js
+++ b/src/shader/shaderIridescenceFunctions.js
@@ -1,0 +1,130 @@
+export const shaderIridescenceFunctions = /* glsl */`
+
+// XYZ to sRGB color space
+const mat3 XYZ_TO_REC709 = mat3(
+	 3.2404542, -0.9692660,  0.0556434,
+	-1.5371385,  1.8760108, -0.2040259,
+	-0.4985314,  0.0415560,  1.0572252
+);
+
+vec3 fresnel0ToIor( vec3 fresnel0 ) {
+
+	vec3 sqrtF0 = sqrt( fresnel0 );
+	return ( vec3( 1.0 ) + sqrtF0 ) / ( vec3( 1.0 ) - sqrtF0 );
+
+}
+
+// Conversion FO/IOR
+vec3 iorToFresnel0( vec3 transmittedIor, float incidentIor ) {
+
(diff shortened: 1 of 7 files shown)


richardassar:
I've updated this PR to remove conflicts with current main. Please let me know if there's anything you'd like me to add or change.

Below is a quick test showing Clear Coat, Sheen and Iridescence material properties in use simultaneously.

Screenshot from 2022-07-20 10-03-11

src/shader/shaderIridescenceFunctions.js
```diff
@@ -0,0 +1,129 @@
...
+
+	// Force iridescenceIor -> outsideIOR when thinFilmThickness -> 0.0
+	float iridescenceIor = mix( outsideIOR, eta2, smoothstep( 0.0, 0.03, thinFilmThickness ) );
+
+	// Evaluate the cosTheta on the base layer (Snell law)
+	float sinTheta2Sq = square( outsideIOR / iridescenceIor ) * ( 1.0 - square( cosTheta1 ) );
+
+	// Handle TIR:
+	float cosTheta2Sq = 1.0 - sinTheta2Sq;
+	if ( cosTheta2Sq < 0.0 ) {
```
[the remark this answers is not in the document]
richardassar: Alright, I'll add those now.

richardassar:
Apologies, I'd pushed with a malformed comment, fixed that, tested and pushed again.

src/shader/shaderUtils.js
```diff
@@ -290,4 +296,28 @@ export const shaderUtils = /* glsl */`
...
+
+		return t * t;
+
+	}
+
+	vec4 square( vec4 t ) {
+
+		return t * t;
+
+	}
```
gkjohnson, line 321: Out of curiosity - is this faster or better than using `pow( t, 2.0 )`?

gkjohnson:
This is great, thanks!

src/shader/shaderUtils.js
richardassar: I believe it will be faster unless the compiler could detect the constant on the RHS and specialises this; however, `pow( x, y )` is undefined for `x < 0` which would make this specialisation invalid.

richardassar:
You're welcome, on to the next one.

Issue #204, opened 2022-07-09
Allow the pathtracer to filter certain objects before rendering the scene


joewhatkins:
Different from #99, as I'd like to just exclude certain meshes from the rendering instead of certain areas.

We're working on a use case where users can switch between conventional rendering for editing a scene and the pathtracer for nice visualizations of the scene.

There are a ton of edit gizmos and widgets we want to include in the edit view but exclude from the pathtracing view.

The easiest way to implement this would probably be to pass a method to (PathTracingSceneGenerator|DynamicPathTracingSceneGenerator) that would filter each of the scene's meshes before building the BVH and generating the scene. This seems to work well implementing it locally.

This sound good? If so I can have a PR out pretty quickly.

gkjohnson:
I'm thinking it might be simplest to just afford passing in an array of Object3Ds _or_ a single Scene in order to support this use case. Then the end user can perform whatever filtering strategy they want and pass in the result. Then each object is included recursively in the processing. That would mean changing the APIs to the following:

```js
const dynamicGenerator = new DynamicPathTracingSceneGenerator( Object3D | Array<Object3D> );

const generator = new PathTracingSceneGenerator();
generator.generate( Object3D | Array<Object3D> );
```

How does that sound?

joewhatkins:
SGTM - to be clear, we still want users to be able to pass in a scene after this change, right? Just also support single Objects and arrays?

joewhatkins:
Fat fingered and accidentally closed this, reopening

gkjohnson:
There would be no change for single objects. A `Scene` is just an instance of an `Object3D` which is what's [currently supported by the generation functions](https://github.com/gkjohnson/three-gpu-pathtracer#generate).

PR #207, opened 2022-07-13
Allow users to pass an array of Object3Ds to the pathtracer instead of an entire scene

joewhatkins:
See #204 

This allows users to perform their own filtering of a scene's objects prior to passing it to the pathtracer, in case there are any elements of the scene they'd like to omit from the path-traced rendering.

### DIFF
--- a/src/core/DynamicPathTracingSceneGenerator.js
+++ b/src/core/DynamicPathTracingSceneGenerator.js
@@ -12,7 +12,7 @@ export class DynamicPathTracingSceneGenerator {
 
 	constructor( scene ) {
 
-		this.scene = scene;
+		this.objects = Array.isArray( scene ) ? scene : [ scene ];
 		this.bvh = null;
 		this.geometry = new BufferGeometry();
 		this.materials = null;
@@ -28,7 +28,7 @@ export class DynamicPathTracingSceneGenerator {
 		this.geometry = new BufferGeometry();
 		this.materials = null;
 		this.textures = null;
-		this.staticGeometryGenerator = new StaticGeometryGenerator( this.scene );
+		this.staticGeometryGenerator = new StaticGeometryGenerator( this.objects );
 
 	}
 
@@ -36,7 +36,7 @@ export class DynamicPathTracingSceneGenerator {
 
 	generate() {
 
-		const { scene, staticGeometryGenerator, geometry } = this;
+		const { objects, staticGeometryGenerator, geometry } = this;
 		if ( this.bvh === null ) {
 
 			const attributes = [ 'position', 'normal', 'tangent', 'uv' ];
@@ -45,22 +45,22 @@ export class DynamicPathTracingSceneGenerator {
 
 				if ( c.isMesh ) {
 
-					const normalMapRequired = ! ! c.material.normalMap;
-					setCommonAttributes( c.geometry, { attributes, normalMapRequired } );
+					c.traverse( ( cTrav ) => {
+
+						if ( cTrav.isMesh ) {
+
+							const normalMapRequired = ! ! c.material.normalMap;
+							setCommonAttributes( c.geometry, { attributes, normalMapRequired } );
+
+						}
+
+					} );
 
 				}
 
 			}
 
-			if ( scene instanceof Object3D ) {
-
-				scene.traverse( checkObject );
-
-			} else if ( scene instanceof Array ) {
-
-				scene.forEach( checkObject );
-
-			}
+			objects.forEach( checkObject );
 
 			const textureSet = new Set();
 			const materials = staticGeometryGenerator.getMaterials();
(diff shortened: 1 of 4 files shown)


src/core/DynamicPathTracingSceneGenerator.js
```diff
@@ -49,7 +50,17 @@ export class DynamicPathTracingSceneGenerator {
...
+
+			if ( scene instanceof Object3D ) {
+
+				scene.traverse( checkObject );
+
+			} else if ( scene instanceof Array ) {
+
+				scene.forEach( checkObject );
+
+			}
```
gkjohnson, line 64: For the DynamicPathtracingSceneGenerator I think it's simplest if we just ensure that the `.scene` field is an array in the constructor which will simplify this "generate" function:

```js
this.scene = Array.isArray( scene ) ? scene : [ scene ];
```

We can also rename `this.scene` to `this.objects` so it's more semantically correct.

src/core/PathTracingSceneGenerator.js
```diff
@@ -33,7 +34,17 @@ export class PathTracingSceneGenerator {
...
+
+		if ( scene instanceof Object3D ) {
+
+			scene.traverse( checkObject );
+
+		} else if ( scene instanceof Array ) {
+
+			scene.forEach( checkObject );
+
+		}
```
gkjohnson, line 47: And here I think it would be best if we just ensure the passed object is an array at the beginning of the function:

```js
prepScene( scene ) {

  scene = Array.isArray( scene ) ? scene : [ scene ];

  // ...
```

This should also simplify the function contents

gkjohnson (commented):
Great thanks! Just a couple comments on the approach here. And aside from that we should update the docs!

src/core/PathTracingSceneGenerator.js
joewhatkins: For both this and the previous comment, I think we might want to keep it the way it is in the PR.

We want different behavior when passing it an array vs a single scene object. For arrays, we want to filter each object individually, ignoring their children. For scenes, we want to traverse through the tree and filter all of the children.

If we cast everything to an array at the beginning of generation, we'd need to traverse through each object inside of checkObject to replicate the current behavior when passing a scene. Assuming we treat prefiltered arrays of objects the same way, we'll include all of their children in the output scene, even if they were filtered before being passed to the method.

In our case, we have characters with several control widgets/gizmos and geometries as their descendants. We don't want to display the gizmo geometries in the pathtracer. We filter them out before passing them to the pathtracer:

`
let filteredObjects = []
scene.traverse((c) => {
  if (!isGizmoGeometry(c)) {
    filteredObjects.push(c)
  }
})

const results = generator.generate(filteredObjects)
`

If we implement things the way you're suggesting, we'd still end up rendering those gizmos since their parents (the characters) are still in filteredObjects. You could suggest we make copies of the characters and remove their children before passing them to .generate, but at that point we might as well just copy the scene entirely and heavily modify it before passing it to the current version of .generate

gkjohnson: I think I feel strongly that passing a single scene in and an array of one scene in should behave the same. I think it would be confusing if these function calls behaved differently:

```js
generator.generate( scene );

generator.generate( [ scene ] );
```

Can you explain the structure of your hierarchy? Do your _meshes_ have children? Are you not using groups to store meshes and gizmos? Another option is to set the gizmos to be `visible = false` before updating the geometry and resetting them back to visible afterward.

joewhatkins: Ignore my prior comment - I think I misunderstood why some elements were making it pass my filtering function. Your way should work fine for our use case. PR will be updated shortly.

src/core/DynamicPathTracingSceneGenerator.js
```diff
@@ -28,28 +28,39 @@ export class DynamicPathTracingSceneGenerator {
...
 	generate() {
 
-		const { scene, staticGeometryGenerator, geometry } = this;
+		const { objects, staticGeometryGenerator, geometry } = this;
 		if ( this.bvh === null ) {
 
 			const attributes = [ 'position', 'normal', 'tangent', 'uv' ];
-			scene.traverse( c => {
+
+			function checkObject( c ) {
```
gkjohnson, line 44: Great, thanks! With the change I think this `checkObject` function is no longer needed and can instead just be moved inside of a for loop.

src/core/DynamicPathTracingSceneGenerator.js
```diff
@@ -28,28 +28,39 @@ export class DynamicPathTracingSceneGenerator {
...
+						}
+
+					} );
 
 				}
 
-			} );
+			}
+
+			objects.forEach( checkObject );
```
gkjohnson, line 63: I prefer to use a for loop rather than forEach generally since they're faster:

```js
for ( let i = 0, l = objects.length; i < l; i ++ ) {

  const c = objects[ i ];

  // ...

}
```

gkjohnson:
Great thank you!

PR #208, opened 2022-07-13
Fix shadow, side, matte params

gkjohnson:
Related to #203 

cc @richardassar - looks like these offsets got missed in the clearcoat PR, just fyi! I missed them when testing, too. Found the issue while looking through the iridescence PR.


richardassar:
Thanks for the heads up and for fixing that. I don't know how that one slipped by, I check diffs pretty meticulously but will diff against main and/or check github's PR page more closely in future.

PR #216, opened 2022-07-21
Add lights to Dynamic Scene Generator.

mikelyndon:
I think this is relatively straight forward. The DynamicPathTracingSceneGenerator would also check for and add existing area lights.
This supports some work I'm doing on the react-three-gpu-pathtracer to update to 0.0.5 and give access to lights.


gkjohnson:
Looks great, thanks!

Issue #215, opened 2022-07-21
White Artifacts on Glancing Edges in v0.0.5


joewhatkins:
**Describe the bug**

Release 0.0.5 introduced an issue where we get strange bright segments on glancing angles on our meshes.

![image](https://user-images.githubusercontent.com/85718573/180146525-da7e8f9e-329e-44ae-8ed9-57a6b72c43b6.png)

On inspection of some of the examples in this repo, we noticed that there are similar artifacts on the skinnedMesh example.

![image](https://user-images.githubusercontent.com/85718573/180147449-ec2649ff-2e44-4270-a8c7-96e30fa256e7.png)

I ran the same examples after downgrading to 0.0.4 and there weren't any artifacts.

![image](https://user-images.githubusercontent.com/85718573/180147610-e10f7b8e-777e-4963-9530-ad769ceff1e0.png)

![image](https://user-images.githubusercontent.com/85718573/180147984-3fde1122-6343-4118-9d44-d72b76b5dc06.png)

Running a bisect identified [e4e275746877aea2579bc6a29792a0137da5bb70](https://github.com/gkjohnson/three-gpu-pathtracer/commit/e4e275746877aea2579bc6a29792a0137da5bb70) as the breaking commit - specifically the change to passing wo to ggxPDF instead of wi in specularPDF. This was originally implemented as part of #174 to add support for area lights and fix issues with a divergence between MIS and non-MIS rendering. 

I tried reading the referenced materials but it seems like the implementaitons in three-gpu-pathtracer diverge from them in a few different ways. I looked at other sources explaining specular PDF methods and found a few other implementations - one of which set the specular PDF to zero if the product of the normals of the outgoing direction and incident direction is less than zero. I'm not exactly an expert on PBR so please let me know if there's a more principled solution to this.

It's too late here for me to parse out exactly what this means, but implementing it locally seemed to (mostly) resolve the issue while still maintaining the expected area light behavior.

**0.0.4 (working as expected)**
![image](https://user-images.githubusercontent.com/85718573/180150389-c34cac36-0262-45ee-a08e-e8eb349a3ad5.png)

**0.0.5 without fix**
![image](https://user-images.githubusercontent.com/85718573/180150085-5c400349-21d2-4f8e-9ca4-8304dbf6b3f5.png)

**0.0.5 with fix**
![image](https://user-images.githubusercontent.com/85718573/180150658-0e572da2-9459-49a6-8190-78b0ae648203.png)

[...]

PR #221, opened 2022-07-24
Fix specular edge artifacts

joewhatkins:
Update the specularPDF method so that it takes the absolute value of several parameters - see https://www.pbr-book.org/3ed-2018/Light_Transport_I_Surface_Reflection/Sampling_Reflection_Functions for an in-depth explanation.

This change fixes the issues with white highlights on glancing edges described in #215 

**Before**
image

**After**
image

### DIFF
--- a/src/shader/shaderMaterialSampling.js
+++ b/src/shader/shaderMaterialSampling.js
@@ -73,9 +73,6 @@ float specularPDF( vec3 wo, vec3 wi, SurfaceRec surf ) {
 	// See 14.1.1 Microfacet BxDFs in https://www.pbr-book.org/
 	float filteredRoughness = surf.filteredRoughness;
 	vec3 halfVector = getHalfVector( wi, wo );
-	if ( wi.z * wo.z < 0.0 ) {
-		return 0.0;
-	}
 
 	float incidentTheta = acos( wo.z );
 	float D = ggxDistribution( halfVector, filteredRoughness );
--- a/src/shader/shaderMaterialSampling.js
+++ b/src/shader/shaderMaterialSampling.js
@@ -70,10 +70,18 @@ vec3 diffuseColor( vec3 wo, vec3 wi, SurfaceRec surf ) {
 // specular
 float specularPDF( vec3 wo, vec3 wi, SurfaceRec surf ) {
 
-	// See equation (27) in http://jcgt.org/published/0003/02/03/
+	// See 14.1.1 Microfacet BxDFs in https://www.pbr-book.org/
 	float filteredRoughness = surf.filteredRoughness;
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks! Can you talk a bit about the change from the equations in the `jcgt` paper to the ones in pbr book? Should we change the direction sampling to align with the PDF function now, as well?

joewhatkins:
Actually, I couldn't parse out how equation 27 in http://jcgt.org/published/0003/02/03/ mapped to the formula I corrected from #174 - nor could I parse out how equation 17 did (the cited equation before the change in #174).

In the process of trying to figure all of this out, I read through http://jcgt.org/published/0007/04/01/, a different paper hosted at the same site, that is cited in shaderGGXFunctions.js. Equation 17 in http://jcgt.org/published/0007/04/01/ does seem to correspond to specularPDF - so I think it's possible that someone meant to cite http://jcgt.org/published/0007/04/01/ but cited http://jcgt.org/published/0003/02/03/ instead.

As far as I can tell, [e4e2757](https://github.com/gkjohnson/three-gpu-pathtracer/pull/174/commits/e4e275746877aea2579bc6a29792a0137da5bb70) did fix a legitimate error - the specular PDF should be computed from the outgoing direction - but we should be using wo in the denominator instead of wi.

The equations I highlighted in the PBR book are very similar to the ones outlined in http://jcgt.org/published/0007/04/01/ but include some pseudocode with some differences that fixed the issues with highlights when I incorporated them.

The other changes involved taking the absolute value of some of the normal and dot product terms in the ggxPDF method. I broke this out from the ggxPDF method in shaderGGXFunctions.js since I figured that might affect things other than these specular problems. That might be a thing to incorporate into a bigger PR. Either way this seems to fix the issue.

All of that said, I'd love to hear more about [e4e2757](https://github.com/gkjohnson/three-gpu-pathtracer/pull/174/commits/e4e275746877aea2579bc6a29792a0137da5bb70) from @mrxz to make sure my understanding is correct.

gkjohnson:
Okay great, thanks for the explanation! It would have been me that messed up the comment references. It's on my list to revisit the GGX and sampling functions to address some other issues, too, so I'll take another pass at the comments, as well. This is definitely an improvement, though, so I'll merge this before the next release!

mrxz:
Since we compute the `halfVector` from `wi` and `wo` the dot product of `wi` and `halfVector` is the same as the dot product of `wo` and `halfVector`. So it should be fine to use either in the denominator. 

The thing I don't see yet is which change actually solved the issue and why. I would expect `dot( wo, halfVector )` to be always positive in the specularPdf case. Otherwise, shouldn't we also take the abs of it in the denominator?

Issue #60, opened 2022-04-01
Investigate TRAA for retaining samples on change


gkjohnson:

bhouston:
Ah, this is a duplicate of this: https://github.com/gkjohnson/three-gpu-pathtracer/issues/191

bhouston:
**Is your feature request related to a problem? Please describe.**

Temporal reproduction anti-aliasing is a method of reusing samples from previous frames on the current frame but warping them based on the scene/camera movement that occurred.  While this renderer currently does accumulate samples across frames, it can not handle motion in those frames.  Implementing TRAA will enable it to handle motion.

**Describe the solution you'd like**

Something akin to this:
https://www.youtube.com/watch?v=2XXS5UyNjjU

I believe naive approaches do not handle transparent backgrounds because those are not required in games.  This is okay for the first implementation, but a further improvement would be to handle transparent backgrounds.

richardassar:
I remember this talk and I'd been thinking about this in recent days, for static scenes an idea came to mind which I'm not sure has been tried before: output from the path tracer can be used to train a NeRF network, baking the scene into the network weights and hash table.

[Online demo of NeRF networks running in the browser](https://kwea123.github.io/nerf_pl/)

Although that's not what this issue requests I think it would be interesting to try. There's no need for camera rays to be coherent also since they're independent, so you can just sample random rays within a given volume and get good scene coverage.

This approach is somewhat related to the much more complex [Real-time Neural Radiance Caching for Path Tracing](https://research.nvidia.com/publication/2021-06_real-time-neural-radiance-caching-path-tracing) which makes use of Q-learning, probably a step too far for WebGL.

For retargeting I think having an approximate differentiable renderer could be an interesting approach, as you can quickly reject samples that fall within some bound of imperceptibility given the (possibly approximate) partial derivative of the final output with respect to all scene parameters. 

[Differentiable Monte Carlo Ray Tracing through Edge Sampling](https://people.csail.mit.edu/tzumao/diffrt/)

[Differentiable Path Tracing on the GPU/TPU](https://blog.evjang.com/2019/11/jaxpt.html)

[Path-Space Differentiable Rendering](https://shuangz.com/projects/psdr-sg20/psdr-sg20.pdf)

Differentiable renderers aren't very fast but there may be ways to approximate this alongside the main monte carlo integration.

[...]

PR #241, opened 2022-07-28
Feature: Temporal Resolving

0beqz:
### About

This PR implements temporal resolving to preserve samples and reduce noise on camera movement. It does so by calculating the velocity of each pixel each frame and stores it in a velocity buffer. In the `TemporalResolvePass` it's using the velocity to find out where a pixel was in the last frame and projects that pixel onto the new pixel if possible. It's using the velocity buffer, the current frame's depth buffer and the last frame's depth buffer to reduce artifacts by checking which pixels can't have the last frame reprojected onto them (e.g. because they were occluded in the last frame).

Related issue: https://github.com/gkjohnson/three-gpu-pathtracer/issues/60

### Comparisons

Temporal Resolving first enabled and then disabled:

[Temporal Resolve](https://user-images.githubusercontent.com/49001571/181601922-9e0dbab7-0495-4e53-ad2e-7fea7ca2aa7b.webm)

How good TR looks can depend on the scene, so scenes with complex geometry and intense specular reflections are more sophisticating for TR and won't usually look as smooth as scenes with simple geometry and rough materials.
So the scene from the video works really well for TR for example.

### Usage
The API would look like this:
```javascript
import { TemporalResolve } from 'three-gpu-pathtracer'

const temporalResolve = new TemporalResolve( ptRenderer, scene, camera );

// the in the render loop
renderer.autoClear = false;
temporalResolve.update();
fsQuad.material.map = temporalResolve.target.texture;
renderer.autoClear = true;
```

The parameters to tweak the `TemporalResolve` class are documented in the [readme](https://github.com/0beqz/three-gpu-pathtracer/tree/temporal-resolve#temporalresolve).

### Demos

[Online Demo](https://three-gpu-pathtracer-temporal-resolve.vercel.app/materialBall.html)

Currently, the two demos `example/index.js` and `example/materialBall.js` include Temporal Resolving.

### References
https://media.contentapi.ea.com/content/dam/ea/seed/presentations/dd18-seed-raytracing-in-hybrid-real-time-rendering.pdf (Raytracing in hybrid real-time rendering)
https://github.com/0beqz/screen-space-reflections#temporal-reprojection (mainly articles about TRAA)

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,56 @@ _extends THREE.Camera_
 
 A class indicating that the path tracer should render an equirectangular view. Does not work with three.js raster rendering.
 
+## TemporalResolve
+
+A class that implements temporal filtering to preserve samples when the camera is moving. This helps reduce noise on camera movement by reprojecting the last frame's samples into the current one.
+
+### .temporalResolveMix
+
+```javascript
+temporalResolveMix = 0.75 : Number
+```
+
+How much the last frame should be blended into the current one. Higher values will result in a less noisy look at the cost of more smearing.
+
+### .clampRing
+
+```javascript
+clampRing = 1 : Number
+```
+
+An integer to set the radius of pixels to be used for neighborhood clamping. Higher values will result in a less noisy look at the cost of more blurring.
+
+### .newSamplesSmoothing
+
+```javascript
+newSamplesSmoothing = 0.5 : Number
+```
+
+To reduce noise for pixels that appeared recently, the average of multiple adjacent pixels can be used instead of a pixel itself. This factor determines the influence of the averaged pixel. Higher values will result in less noise but also less sharpness.
+
+### .newSamplesCorrection
+
+```javascript
+newSamplesCorrection = 0.75 : Number
+```
+
+Higher values will make pixels that appeared recently have a greater influence on the output. This will result in more noise but less smearing.
+
+### .constructor
+
+```javascript
+constructor( ptRenderer : PathTracingRenderer, scene : Object3D, camera : Camera )
+```
+
+### .update
+
+```javascript
+update() : void
+```
+
+Updates the temporal resolve pass for the current frame.
+
 ## PhysicalSpotLight
 
 _extends THREE.SpotLight_
--- a/README.md
+++ b/README.md
@@ -45,14 +45,13 @@ _More features and capabilities in progress!_
 
 [Ambient Occlusion Material](https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/aoRender.html)
 
-
 ## Running examples locally
 
-To run and modify the examples locally, make sure you have Node and NPM installed.  Check the supported versions in [the test configuration](./.github/workflows/node.js.yml).
+To run and modify the examples locally, make sure you have Node and NPM installed. Check the supported versions in [the test configuration](./.github/workflows/node.js.yml).
 
 In order to install dependencies, you will need `make` and a C++ compiler available.
 
-On Debian or Ubuntu, run `sudo apt install build-essential`.  It should just work on MacOS.
+On Debian or Ubuntu, run `sudo apt install build-essential`. It should just work on MacOS.
 
 - To install dependencies, run `npm install`
 - To start the demos run `npm start`
@@ -63,13 +62,13 @@ On Debian or Ubuntu, run `sudo apt install build-essential`.  It should just wor
 **Basic Renderer**
 
 ```js
-import * as THREE from 'three';
-import { FullScreenQuad } from 'three/examples/jsm/postprocessing/Pass.js';
+import * as THREE from "three";
+import { FullScreenQuad } from "three/examples/jsm/postprocessing/Pass.js";
 import {
 	PathTracingSceneGenerator,
 	PathTracingRenderer,
 	PhysicalPathTracingMaterial,
-} from 'three-gpu-pathtracer';
+} from "three-gpu-pathtracer";
 
 // init scene, renderer, camera, controls, etc
 
@@ -78,8 +77,8 @@ renderer.toneMapping = THREE.ACESFilmicToneMapping;
 
 // initialize the path tracing material and renderer
 const ptMaterial = new PhysicalPathTracingMaterial();
-const ptRenderer = new PathTracingRenderer( renderer );
-ptRenderer.setSize( window.innerWidth, window.innerHeight );
+const ptRenderer = new PathTracingRenderer(renderer);
+ptRenderer.setSize(window.innerWidth, window.innerHeight);
 ptRenderer.camera = camera;
 ptRenderer.material = ptMaterial;
 
@@ -87,45 +86,46 @@ ptRenderer.material = ptMaterial;
 ptRenderer.alpha = true;
 
 // init quad for rendering to the canvas
-const fsQuad = new FullScreenQuad( new THREE.MeshBasicMaterial( {
-	map: ptRenderer.target.texture,
+const fsQuad = new FullScreenQuad(
+	new THREE.MeshBasicMaterial({
+		map: ptRenderer.target.texture,
 
-	// if rendering transparent background
-	blending: THREE.CustomBlending,
-} ) );
+		// if rendering transparent background
+		blending: THREE.CustomBlending,
+	})
+);
 
 // ensure scene matrices are up to date
 scene.updateMatrixWorld();
 
 // initialize the scene and update the material properties with the bvh, materials, etc
 const generator = new PathTracingSceneGenerator();
-const { bvh, textures, materials, lights } = generator.generate( scene );
+const { bvh, textures, materials, lights } = generator.generate(scene);
 
 // update bvh and geometry attribute textures
-ptMaterial.bvh.updateFrom( bvh );
-ptMaterial.normalAttribute.updateFrom( geometry.attributes.normal );
-ptMaterial.tangentAttribute.updateFrom( geometry.attributes.tangent );
-ptMaterial.uvAttribute.updateFrom( geometry.attributes.uv );
+ptMaterial.bvh.updateFrom(bvh);
+ptMaterial.normalAttribute.updateFrom(geometry.attributes.normal);
+ptMaterial.tangentAttribute.updateFrom(geometry.attributes.tangent);
+ptMaterial.uvAttribute.updateFrom(geometry.attributes.uv);
 
 // update materials and texture arrays
-ptMaterial.materialIndexAttribute.updateFrom( geometry.attributes.materialIndex );
-ptMaterial.textures.setTextures( renderer, 2048, 2048, textures );
-ptMaterial.materials.updateFrom( materials, textures );
+ptMaterial.materialIndexAttribute.updateFrom(geometry.attributes.materialIndex);
+ptMaterial.textures.setTextures(renderer, 2048, 2048, textures);
+ptMaterial.materials.updateFrom(materials, textures);
 
 // update the lights
-ptMaterial.lights.updateFrom( lights );
+ptMaterial.lights.updateFrom(lights);
 ptMaterial.lightCount = lights.length;
 
 // set the environment map
-const texture = await new RGBELoader().loadAsync( envMapUrl );
-ptRenderer.material.envMapInfo.updateFrom( texture );
+const texture = await new RGBELoader().loadAsync(envMapUrl);
+ptRenderer.material.envMapInfo.updateFrom(texture);
 
 animate();
 
 // ...
 
 function animate() {
-
 	// if the camera position changes call "ptRenderer.reset()"
 
 	// update the camera and render one sample
@@ -137,8 +137,7 @@ function animate() {
 	fsQuad.material.map = ptRenderer.target.texture;
 
 	// copy the current state of the path tracer to canvas to display
-	fsQuad.render( renderer );
-
+	fsQuad.render(renderer);
 }
 ```
 
@@ -147,16 +146,15 @@ function animate() {
 Using a pre blurred envioronment map can help improve frame convergence time at the cost of sharp environment reflections. If performance is concern then multiple importance sampling can be disabled and blurred environment map used.
 
 ```js
-import { BlurredEnvMapGenerator } from 'three-gpu-pathtracer';
+import { BlurredEnvMapGenerator } from "three-gpu-pathtracer";
 
 // ...
 
-const envMap = await new RGBELoader().loadAsync( envMapUrl );
-const generator = new BlurredEnvMapGenerator( renderer );
-const blurredEnvMap = generator.generate( envMap, 0.35 );
+const envMap = await new RGBELoader().loadAsync(envMapUrl);
+const generator = new BlurredEnvMapGenerator(renderer);
+const blurredEnvMap = generator.generate(envMap, 0.35);
 
 // render!
-
 ```
 
 ## Dynamic Scenes
@@ -164,12 +162,12 @@ const blurredEnvMap = generator.generate( envMap, 0.35 );
 Using the dynamic scene generator the same, frequently updated scene can be converted into a single reusable geometry multiple times and BVH refit which greatly improves subsequent scene updates. See `DynamicPathTracingSceneGenerator` docs for more info.
 
 ```js
-import { DynamicPathTracingSceneGenerator } from 'three-gpu-pathtracer';
+import { DynamicPathTracingSceneGenerator } from "three-gpu-pathtracer";
 
 // ... initialize scene etc
 
-const generator = new DynamicPathTracingSceneGenerator( scene );
-const { bvh, textures, materials } = generator.generate( scene );
+const generator = new DynamicPathTracingSceneGenerator(scene);
+const { bvh, textures, materials } = generator.generate(scene);
 
 // ... update path tracer and render
 ```
@@ -179,13 +177,13 @@ const { bvh, textures, materials } = generator.generate( scene );
 _NOTE WebWorker syntax is inconsistently supported across bundlers and sometimes not supported at all so the PathTracingSceneWorker class is not exported from the package root. If needed the code from src/worker can be copied and modified to accomodate a particular build process._
 
 ```js
-import { PathTracingSceneWorker } from 'three-gpu-pathtracer/src/workers/PathTracingSceneWorker.js';
+import { PathTracingSceneWorker } from "three-gpu-pathtracer/src/workers/PathTracingSceneWorker.js";
 
 // ...
 
 // initialize the scene and update the material properties with the bvh, materials, etc
 const generator = new PathTracingSceneWorker();
-const { bvh, textures, materials, lights } = await generator.generate( scene );
+const { bvh, textures, materials, lights } = await generator.generate(scene);
 
 // ...
 ```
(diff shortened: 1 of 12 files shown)


bhouston:
This is amazing work @0beqz!  So exciting to see it actually work!

BTW I have a similar bounty over at Three.js, which may be easier for you to do now - https://github.com/mrdoob/three.js/issues/14050

gkjohnson:
Agreed! This is really great.

For formatting if you run `npm run lint -- --fix` it should fix most of the linter issues automatically

0beqz:
Glad to hear!

Looks like I formatted all files with the wrong config by accident in a certain commit. The formatting should be correct now in the recent commit.

gkjohnson:
I still need to dig into the code more but I have a couple thoughts on the tiling behavior. At the moment when there's more than one "tile" that renders the TRAA only applies to the first tile during camera changes. I think one of the really nice capabilities of this would be that even when moving the camera with tiled rendering that the rest of screen does not revert to the background or rasterized rendering.

And then the tiling could also continue to increment throughout the camera changes instead of just rerendering the top left corner.

What are your thoughts?

Also:

> The VelocityShader.js file is from threejs-sandbox: https://github.com/gkjohnson/threejs-sandbox/blob/master/shader-replacement/src/passes/VelocityShader.js

🎉  😁

I'm excited to see how this works with animated geometry, too.

0beqz:
Well so one problem for TR with tiled rendering is that when you keep moving the camera constantly then it will happen that all the tiles (besides the top-left tile) won't be rendered for a lot of frames.
So let's say you have tiling enabled in the `materialBall` scene and then keep moving the camera constantly all around the ball in the scene for a span of multiple seconds (so that it always has to resample each frame). Since none of the tiles besides the top-left one will be rendered during that time we can't use TR anymore otherwise it will just smear a lot since the only info TR has is the frame before you started moving the camera around.

I'm not really sure what would be the best solution here.

gkjohnson:
> Well so one problem for TR with tiled rendering is that when you keep moving the camera constantly then it will happen that all the tiles (besides the top-left tile) won't be rendered for a lot of frames.

Right this is what I was alluding to. We can handle this bit in another PR but I think a good solution is to just not render only the first time when the path tracer has been reset. When TRAA is enabled we'd want to step through all the tiles to render so the reset of the scene doesn't just degrade into a blurry mess. Then you can render 1/4 of the screen per frame while sill having a coherent path traced view.

For this PR, though, I think it would be okay if the rest of the scene other than the top left corner became blurry.

0beqz:
Ah I see, yeah that sounds like the best solution. I will add another commit where it implements that so that it works better with tiling.

0beqz:
Alright, I've updated TR:
- made it re-render the pathtracer till `samples` >= 1 (so that all tiles will be rendered first before we do TR), now it always displays the pathtraced output when using tiling
- Added dilation when fetching velocity to get rid of aliasing and flickering at edges (idea from [here](https://www.elopezr.com/temporal-aa-and-the-quest-for-the-holy-trail/))
- re-worked the TR blending algorithm so that it's less smeary now (especially when there is a lot of movement)
- added `weightTransform` property (based on "Intensity / Color Weighing" from [here](https://www.elopezr.com/temporal-aa-and-the-quest-for-the-holy-trail/)) to reduce noise on camera movement)

N8python:
I literally shouted with joy when I first read this PR. Insane work.

gkjohnson:
I'm slowly taking a look at this - it's gonna take me a little bit to get my head around it. I'm making a PR to make into your branch with some code style changes to make it more consistent with the rest of the project as I go. One thing I noticed is that performance seems to be _worse_ when using tiled rendering with temporal resolve than when rerendering the full frame. Is this something you've noticed, as well?

0beqz:
I see, well yeah makes sense. This [line](https://github.com/0beqz/three-gpu-pathtracer/blob/temporal-resolve/src/temporal-resolve/TemporalResolve.js#L91) here will be the reason:
```javascript
while ( this.ptRenderer.samples < 1 ) this.ptRenderer.update();
```

Right now TR will keep re-rendering the pathtracer till all tiles are rendered to prevent smearing. That's probably more expensive than just rendering a single full frame. I think the optimal solution would be that, if TR is enabled, we don't use tiling for the very first sample and then just re-enable tiling afterwards. To do that I'll have to change the structure of the TR pass so that it calls `ptRenderer.update()` now and changes the ptRenderer's tiling based on if it's the first sample (right now it doesn't change the ptRenderer and only uses its ouput textures). So then you'll no longer need to call both `ptRenderer.update()` and `temporalResolve.update()` when using TR.
So what do you think if I should change that?

gkjohnson:
> I see, well yeah makes sense. This [line](https://github.com/0beqz/three-gpu-pathtracer/blob/temporal-resolve/src/temporal-resolve/TemporalResolve.js#L91) here will be the reason:
>
> ```js
> while ( this.ptRenderer.samples < 1 ) this.ptRenderer.update();
> ```

Got it thanks -- that seems to be it. I'm surprised rendering the tiles separately has such a significant performance impact but that seems to be the issue.

Why don't we remove the while loop understanding that it's not going to look right. I think it should be possible to tile rendering over multiple frames with this temporal resolve and have it look nice but I'm having trouble articulating the plan.

Here's what I'm seeing right now when removing that line - the other tiles render as black in the temporal resolve. Is it possible to have them retain the latest value in the TRAA texture? I know it'll be a blurry mess but it's a first step. Once I figure that out I'll toy around with the other options I'm imagining.

https://user-images.githubusercontent.com/734200/183544503-c9ae5447-4171-4535-91a5-a7812e0eb5ea.mov

---

I'm also seeing that the TRAA texture seems to have the wrong aspect ration once the screen has been resized. You can see some of the hotspots in the texture look stretched when making the window particularly narrow and it started off at a normal aspect ratio.

And lastly I've made https://github.com/0beqz/three-gpu-pathtracer/pull/1 with some style changes to match the project while I was taking a look at things!

0beqz:
Yeah that's possible, you can detect if a pixel of the PT's buffer was not rendered yet in the shader through the alpha value (it will be 0 until the tile of the current texel is rendered).
And I think your [PR](https://github.com/gkjohnson/three-gpu-pathtracer/pull/251) would be a great solution here, I'll check how it looks with TR. But it will definitely reduce smearing with that solution.
One issue would be that if you have a lot of tiles to be rendered e.g. 4x4 tiles then it will result in smearing again since it still takes many frames to have a specific tile rendered.

How's the performance difference between rendering a single tile and rendering the entire frame? I was thinking that we could always render the full frame on the first sample when TR is enabled (even when we use tiling). This should solve the smearing problem. Afterwards we'd resume with tiled rendering.

gkjohnson:
> One issue would be that if you have a lot of tiles to be rendered e.g. 4x4 tiles then it will result in smearing again since it still takes many frames to have a specific tile rendered.

I think that's okay. That's just a consequence of using tiling - you're trading performance for quality.

> And thanks for making the PR. Are you using any formatter for the in-line glsl code btw? Looking for one currently

No... I would really like one too but it doesn't look like anyones made one yet

0beqz:
Alright, I've fixed the issue with resizing and improved tiling so that it's no longer re-rendering the pathtracer. Also improved a few other issues like making box-blur depth-aware to reduce smearing.

I've noticed an issue when you have set `stableTiles = true` and either tiles.x or tiles.y is 1 (while the other isn't). It won't render a tile at all then.

> No... I would really like one too but it doesn't look like anyones made one yet

I haven't checked yet but does the build tool support importing external .glsl / .vert / .frag files into JS files? If so you can use clang-format to format these (as C code), works pretty well for formatting

gkjohnson:
> I've noticed an issue when you have set stableTiles = false and either tiles.x or tiles.y is 1 (while the other isn't). It won't render a certain tile at all then.

Thanks! I just pushed a fix. But I think this all looks pretty awesome! It feels pretty good with the incrementing tiles, too.

> I haven't checked yet but does the build tool support importing external .glsl / .vert / .frag files into JS files? If so you can use clang-format to format these (as C code), works pretty well for formatting

That's something to think about - I hadn't considered using clang.

I'll do one more pass at this in a bit and then I think we can merge it if it all looks good!

example/index.js
```diff
@@ -143,6 +151,13 @@ async function init() {
 	ptRenderer.material.bgGradientTop.set( params.bgGradientTop );
 	ptRenderer.material.bgGradientBottom.set( params.bgGradientBottom );
 
+	temporalResolve = new TemporalResolve( ptRenderer, scene, activeCamera );
+	temporalResolve.temporalResolveMix = 0.9;
+	temporalResolve.clampRadius = 1;
+	temporalResolve.newSamplesSmoothing = 0.5;
+	temporalResolve.newSamplesCorrection = 0.75;
+	temporalResolve.weightTransform = 0;
```
gkjohnson, line 159: It may be best to initialize these parameters from the `params` object - otherwise if we change the demo defaults we have to remember to do it in two places.

example/materialBall.js
```diff
@@ -167,6 +173,13 @@ async function init() {
 
 	scene = new THREE.Scene();
 
+	temporalResolve = new TemporalResolve( ptRenderer, scene, activeCamera );
+	temporalResolve.temporalResolveMix = 0.9;
+	temporalResolve.clampRadius = 2;
+	temporalResolve.newSamplesSmoothing = 0.675;
+	temporalResolve.newSamplesCorrection = 1;
+	temporalResolve.weightTransform = 0;
```
gkjohnson, line 181: Same as above - lets initialize from the params object fields.

src/temporal-resolve/TemporalResolve.js
```diff
@@ -0,0 +1,151 @@
...
+		this.newSamplesSmoothing = 0.675;
+		this.newSamplesCorrection = 1;
+
+		this.fullscreenMaterial = new ComposeTemporalResolveMaterial();
+
+		this.fsQuad = new FullScreenQuad( this.fullscreenMaterial );
+
+		this.renderTarget = new WebGLRenderTarget(
+			typeof window !== 'undefined' ? window.innerWidth : 2000,
+			typeof window !== 'undefined' ? window.innerHeight : 1000,
```
gkjohnson, line 31: Can we change this to initialize to a width and height of 1, 1

src/temporal-resolve/materials/shaders/temporalResolveFragment.js
```diff
@@ -0,0 +1,252 @@
...
+    vec4 velocity = getDilatedTexture( velocityTexture, vUv, pxSize );
+
+	vec2 velUv = velocity.xy;
+    vec2 reprojectedUv = vUv - velUv;
+#else
+    vec4 velocity = textureLod( velocityTexture, vUv, 0.0 );
+
+	vec2 velUv = velocity.xy;
+    vec2 reprojectedUv = vUv - velUv;
+#endif
```
gkjohnson, line 108: nit: looks like there's some inconsistent indentation in this shader chunk.

src/temporal-resolve/passes/TemporalResolvePass.js
```diff
@@ -0,0 +1,151 @@
...
+		this.fullscreenMaterial.curInverseProjectionMatrix.copy(
+			this.camera.projectionMatrixInverse
+		);
+		this.fullscreenMaterial.curCameraMatrixWorld.copy(
+			this.camera.matrixWorld
+		);
+		this.fullscreenMaterial.cameraNear = this.camera.near;
+		this.fullscreenMaterial.cameraFar = this.camera.far;
+
+		// now render this fullscreen pass
```
gkjohnson, line 130: I imagine at some point it would be a good plan to switch to use MRT for this, right? Maybe it's worth making an issue for?

src/temporal-resolve/passes/VelocityPass.js
```diff
@@ -0,0 +1,163 @@
...
+	constructor( scene, camera ) {
+
+		this.scene = scene;
+		this.camera = camera;
+
+		this.cachedMaterials = new WeakMap();
+
+		this.renderTarget = new WebGLRenderTarget(
+			typeof window !== 'undefined' ? window.innerWidth : 2000,
+			typeof window !== 'undefined' ? window.innerHeight : 1000,
```
gkjohnson, line 28: Can we initialize these values to 1, 1

README.md
```diff
@@ -391,6 +391,64 @@ _extends THREE.Camera_
...
+
+How much the last frame should be blended into the current one. Higher values will result in a less noisy look at the cost of more smearing.
+
+### .clampRadius
+
+```javascript
+clampRadius = 1 : Number
+```
+
+An integer to set the radius of pixels to be used for neighborhood clamping. Higher values will result in a less noisy look at the cost of more blurring.
```
gkjohnson, line 412: What does "neighborhood clamping" mean here?

gkjohnson:
Also not sure if you have any thoughts on these issues - maybe there's a quick fix? I noticed them in the main `index.html` demo.

There's some flickering in the background which I assume is because we can't get a "velocity" or before / after pixel for the background samples?

https://user-images.githubusercontent.com/734200/184262860-0ddbec15-84a7-4e0e-a405-7ebfc725b961.mov

And there's a lot more background ghosting than I expected on the trees in the Japanese Bridge model. Any thoughts on what's causing that?

https://user-images.githubusercontent.com/734200/184263060-d4554b45-f410-4a17-bffc-3620d44621ce.mov

And of course the transparent floor causes issues but that's expected. Maybe in the future we can use something like dithering in the velocity pass so it's not quite as much an issue.

example/materialBall.js
```diff
@@ -132,7 +138,7 @@ async function init() {
 
 	const aspect = window.innerWidth / window.innerHeight;
 	perspectiveCamera = new PhysicalCamera( 75, aspect, 0.025, 500 );
-	perspectiveCamera.position.set( - 4, 2, 3 );
+	perspectiveCamera.position.set( - 4, 2, 7 );
```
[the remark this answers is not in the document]
0beqz: Sure, I'll change it back

src/temporal-resolve/passes/TemporalResolvePass.js
0beqz: I think we should drop the depth pass altogether because depth is also saved to the blue channel of the velocity pass. Since the velocity pass uses half float type, the precision shouldn't degrade much when fetching depth (currently it's 32 bit but 16 bit will be okay too).

0beqz: I also noticed that MRT can be slower when rendering 2 half float type textures in one pass vs. rendering one unsigned byte type pass and one half float type pass separately, so I'd have to check the performance between these two options.

README.md
0beqz: You clamp the color of the reprojected pixel by the min and max color of the neighboring pixels from the raytracer's output at the current pixel. This is the most popular method to reduce ghosting for TRAA but since we have a noisy output in the first frames when raytracing, neighborhood clamping will lead to flickering there. So by increasing the number of pixel we use for neighborhood clamping, we reduce flickering but will have more smear.

0beqz:
I'm not sure what's causing the flickering of the background there. Is there anything different in this scene compared to the materialBall.html scene? I'm wondering as the background has no flickering there even with tiling enabled.

Yeah currently TR doesn't work well with objects that have thin geometry like also the Rover for example. You should play around with the settings (temporalResolveMix, clampRadius,...) to reduce smearing. They are somewhat dependant on the scene so in the materialBall you can use 'less aggressive' correcting since the geometry is less complex. But I'll look into it and see if I can improve how correction for disoccluded pixels is handled.

src/temporal-resolve/passes/TemporalResolvePass.js
gkjohnson: Gotcha good to know. I didn't know that MRT might be slower if rendering to higher depth textures. I don't think we have to deal with that now. But I did make https://github.com/gkjohnson/three-gpu-pathtracer/issues/253 to track it in the future.

gkjohnson:
I'll have to dig into the flickering more but my intuition without looking deeper is the the velocity / depth texture don't have values for the background image which means the path traced pixels don't get reprojected and the rasterized render shows through? The environment intensity is higher in the index.html file which is why the pathtraced background might look different from the rasterized one.

I would have imagined that it would be possible to improve the ghosting more but your head has been in this more than me. One thing to keep in mind is that the pathtraced renderer is using antialiasing whereas the TRAA depth and velocity passes do not so there will be at least a small amount of misalignment of color on reprojection. So I did expect _some_ ghosting but maybe not as much as in that video 🤔

README.md
gkjohnson: Thanks for the explanation! Maybe we can include more information on the docs about the role color plays in this process

0beqz:
> I'll have to dig into the flickering more but my intuition without looking deeper is the the velocity / depth texture don't have values for the background image which means the path traced pixels don't get reprojected and the rasterized render shows through? The environment intensity is higher in the index.html file which is why the pathtraced background might look different from the rasterized one.

Yeah so I detect if the current texel is from the background by checking if the alpha channel of the velocity texel is 1 (see here: https://github.com/0beqz/three-gpu-pathtracer/blob/temporal-resolve/src/temporal-resolve/materials/shaders/temporalResolveFragment.js#L113). In the VelocityPass, I set the scene's background alpha to 1 and set the alpha of the gl_FragColor to 0 (https://github.com/0beqz/three-gpu-pathtracer/blob/temporal-resolve/src/temporal-resolve/materials/VelocityShader.js#L111). This way I can detect if a texel wasn't rendered in the VelocityPass which should imply it's from the background. So I'm not sure why it isn't detecting the background in your demo. I'll have to look into it.

> I would have imagined that it would be possible to improve the ghosting more but your head has been in this more than me. One thing to keep in mind is that the pathtraced renderer is using antialiasing whereas the TRAA depth and velocity passes do not so there will be at least a small amount of misalignment of color on reprojection. So I did expect some ghosting but maybe not as much as in that video 🤔

Well I I will re-work the algorithm and switch from detecting disocclusions through depth to detecting them through velocity since that seems to be more stable I noticed in other passes where I use TR.
To reduce smearing, just reduce `temporalResolveMix` for such scenes with a lot of smearing.
But yeah it looks like the blending part needs some re-work to deal more aggressively with disocclusions.
I'm not using AA through e.g. multisampling for the velocity and depth buffer but I use dilation in the shader to deal with flickering at edges and as a way to reduce aliasing for them.

README.md
0beqz: Sure, yeah I'll add a more in-depth explanation for it

gkjohnson:
> Well I I will re-work the algorithm and switch from detecting disocclusions through depth to detecting them through velocity since that seems to be more stable I noticed in other passes where I use TR.

Depth seems like a good way to handle this, as well, but it looks like its taking the foreground object color when an object is occluded instead marking it as black / no data like I'd expect. There will definitely wind up being some kind of discontinuity it's just a matter of how we handle it.

If you have some thoughts on how to immediately improve this that would be great otherwise we can also add a few issues to track these and improve the behavior in the future.

0beqz:
> Depth seems like a good way to handle this, as well, but it looks like its taking the foreground object color when an object is occluded instead marking it as black / no data like I'd expect. There will definitely wind up being some kind of discontinuity it's just a matter of how we handle it.

That is definitely possible. I think the main reason for that issue was that I hadn't found a proper way to detect disocclusions yet. I switched from comparing the current and the reprojected world positions to just comparing the two depth values to find disocclusions. This is more stable now and also removes the need for neighborhood clamping that doesn't work too well for this purpose. I'll add that soon to this PR.

Issue #182, opened 2022-06-13
Add support for vertex colors


gkjohnson:
Requires packing vertex colors into a texture and multiplying it into the albedo (similar to normals, uv data textures).

cc @alvinalexander not sure if this is something that might be interesting to you?

alvinalexander:
Sure, I can take a look at this!

alvinalexander:
@gkjohnson This might be a basic question, but how exactly should vertex colors influence the albedo? Right now albedo is set to material.color * material.map.  Is vertex color just another multiplicative factor? If you have any resources, I would like to get deeper understanding of albedo in general.  

https://github.com/gkjohnson/three-gpu-pathtracer/blob/75024a97bb7b9e0d58bc401e3bbc8a910797c4f2/src/materials/PhysicalPathTracingMaterial.js#L207-L214

gkjohnson:
@alvinalexander it actually looks like the vertex colors get dropped when merging geometry right now. When prepping geometry for upload we merge it all into one mesh which also means ensuring that all the merged geometries have the same attributes.

So when merging meshes in [PathTracingSceneGenerator](https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/src/core/PathTracingSceneGenerator.js) and [DynamicPathTracingSceneGenerator](https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/src/core/DynamicPathTracingSceneGenerator.js) we need to include `color` in the attributes that are going to be retained in the final geometry.

The `mergeMeshes` function also calls [setCommonAttributes](https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/src/utils/GeometryPreparationUtils.js#L65) which ensures that all geometry includes a set of attributes and fills in sensible data if it does not. So if no color attributes are available on the geometry then we'll want to create a new buffer of all zeroes.

That's why the color data is missing anyway - hopefully that all makes sense! Once the color data is available you'll want to generate a texture for use in the shader similar to the ones you've linked to.

alvinalexander:
Thanks @gkjohnson 

[...]

Issue #235, opened 2022-07-27
Add ability to disable env map


gkjohnson:
Currently even when it's set to 0 intensity it takes samples

PR #244, opened 2022-07-29
Consolidate lighting information into a single texture

### LINKED ISSUE #223: Consolidate / remove textures
With #184 we're reaching our texture limit for the renderer. Here are some possible solutions for consolidating / removing the need for some textures.

1. ~Simplify the environment map info struct - specifically the need for a `envMapInfo.totalSum` texture to work around struct precision issues.~
1. Consolidate area light and spotlight textures into a single data texture
1. Merge `iesTextures` into the `textures` sampler array (if possible)
1. Pack floating point vertex attribute data into a common sampler array
1. Pack floating point bvh data into a common sampler array / sampler

### LINKED ISSUE #238: Light Textures: consolidate "count" and texture into a struct so they're automatically in sync

gkjohnson:
Fix #223
Fix #238
Related to #182 so we can have an available texture slot

**TODO**
- [x] Move spotlights into common code path
- [x] Remove "SpotLightsTexture"
- [x] Update shader sampling
- [x] Consolidate "SpotLight" and "AreaLight" structures and sampling
- [x] docs
- [x] Move count into lights structure
- [x] Improve spot light PDF
- [ ] Ensure area light result consistency (MIS vs non MIS)
- [ ] Understand how to PDFs for punctual light sources?

Some references for future note:
- https://www.pbr-book.org/3ed-2018/Light_Sources/Point_Lights#Spotlights
- https://www.pbr-book.org/3ed-2018/Light_Transport_I_Surface_Reflection/Sampling_Light_Sources#LightswithSingularities


gkjohnson:
cc @richardassar I refactored a bit of the SpotLight work here a bit in the following ways:
- Consolidate the area light and spot lights information into a single texture
- Form a lighting struct definition so the light count does not need to be set explicitly
- Change the spot light code path to use the same conditions as area lights (ie remove the constant direct light sampling in non-MIS cases)

Removing the direct light sampling has some side effects -- specifically it means that spotlights don't look as clean as quickly in non-MIS cases and they don't work at all when radius === 0. But it does simplify the code and if I can get the PDF working properly it should be able to fix the case where we don't see bounce light from a mirror wall since these kinds of specular rays require following that forward bounce path. Though it will still require that the spotlight radius is non zero to work correctly:

image

The issue is that while I've gotten the MIS and non MIS cases to look consistent by specifying the PDF based on the disc radius - the spot light result does not look right. I'm not sure if you'd have any other thoughts here:

| MIS Before (PDF = 1.0) | MIS After (Adjusted PDF) |
|---|---|
| image | image |

src/shader/shaderLightSampling.js
```diff
@@ -127,66 +161,67 @@ LightSampleRec randomAreaLightSample( Light light, vec3 rayOrigin ) {
...
+	LightSampleRec lightSampleRec;
+	lightSampleRec.hit = true;
+	lightSampleRec.dist = dist;
+	lightSampleRec.direction = direction;
+	lightSampleRec.emission = light.color * light.intensity * distanceAttenuation * spotAttenuation;
 
+	// TODO: this makes the result consistent between MIS and non MIS paths but at radius 0 the pdf is infinite
+	// and the intensity of the light is not correct
 	lightSampleRec.pdf = 1.0;
+	// lightSampleRec.pdf = lightDistSq / ( light.area * cosTheta );
```
gkjohnson, line 202: @richardassar This is where I've adjusted the PDF for the random light sampling. Of course as the radius becomes zero the PDF value becomes Infinity which causes issues. I'm not sure if you have any thoughts here?

richardassar: At one point I had coupled the circular area light with the spot light, in the submitted PR comments (I think) I mentioned that these were decoupled as I thought circular area lights might be useful on their own and that users of spot lights may or may not want to represent the (approximate) diffuse illumination of the lamp.

I think this PDF was fine as it was for a circular (and rectangular) area light. I think simply clamping the denominator to EPS is sufficient here.

richardassar: > I suppose once answer would be to divide some of the PDF equation into the emission?

This seems to work, shifting `( light.area * cosTheta )` over to the emission whilst leaving only `lightDistSq` as the PDF works. I can set a zero radius without problems.

Btw, there was a duplicate of `float cosTheta = dot( direction, normal );` that I just commented out locally as you'd see it in the logs straight away.

gkjohnson: > At one point I had coupled the circular area light with the spot light, in the submitted PR comments (I think) I mentioned that these were decoupled as I thought circular area lights might be useful on their own and that users of spot lights may or may not want to represent the (approximate) diffuse illumination of the lamp.

Yeah I do recall that - I like the separation we have now since the spotlight coverage is different than that circular area light. And it's nice that soft shadows can be supported by increasing the radius. I'm just trying to grok how this should work with the PDF, MIS, and basic forward path tracing to support the specular reflections.

In the mean time I'd like to keep this simplified code path and merged lighting texture, though. I'm thinking I'll disable spot light support for the non-MIS path for now and make an issue for understanding the details of this in the future. These links suggested on Twitter should be helpful for understanding how to handle this. What do you think of that solution for now?

- https://www.pbr-book.org/3ed-2018/Light_Sources/Point_Lights#Spotlights
- https://www.pbr-book.org/3ed-2018/Light_Transport_I_Surface_Reflection/Sampling_Light_Sources#LightswithSingularities

> Btw, there was a duplicate of float cosTheta = dot( direction, normal ); that I just commented out locally as you'd see it in the logs straight away.

Thanks I noticed that, too. Forgot to commit the fix 😁

richardassar: During development I had encountered https://www.pbr-book.org/3ed-2018/Light_Sources/Point_Lights#fragment-SpotLightMethodDefinitions-1 and some code examples in other projects which echoed this. 

For non-IES lights some examples take the approach of setting the PDF to 1.0 inside and 0.0 outside of the spot light cone, I don't like this as it seems luminosity or intensity should affect importance sampling (as [DassaultSystemes-Technology/dspbr-pt](https://github.com/DassaultSystemes-Technology/dspbr-pt does for the diffuse and specular components, convergence is much faster).

I too had this issue with trying to match everything up and I spent quite a bit of time on it over multiple attempts before taking a different approach. Seems reasonable to drop non-MIS for now in favour of code simplification, then later a fresh set of eyes can close the gap.

gkjohnson: > I don't like this as it seems luminosity or intensity should affect importance sampling

Yeah I agree - this could implicitly help with #235, as well. I made #247 and #248 to track these issues. I'll merge this with just multiple importance sampling support for now and we can revisit later. Thanks!

PR #249, opened 2022-08-07
Add denoiser round 2

gkjohnson:
Related to #85, #194

cc @richardassar here's a version of the denoiser PR with the denoiser separated from the PathTracingRenderer so we can more easily swap different denoising strategies in. And the denoiser shader now includes `/*glsl*/` tags and tonemapping functions are embedded so it can be used as a final pass to the screen.


richardassar:
I took a look through the changes, this looks great. I'll follow this pattern for the ODIN denoiser.

Issue #210, opened 2022-07-15
Remaining fidelity / sampling issues


gkjohnson:
- [ ] Incorect transmission support (#66)
- [ ] Missing volume support (#66)
- [ ] Diffuse surfaces look dark
- [ ] Rough metal surfaces look too bright (#76)
- [x] Missing clearcoat support (#168)
- [ ] Missing sheen support (#167)
- [ ] Missing iridescence support (#96)

bhouston:
You have taken this renderer so far so fast.  It is incredibly impressive.

The highest priority are the glaring differences with sheen and tangents (it is likely these two lines: https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/normalmap_pars_fragment.glsl.js#L35) I would suggest.  The furnace test is hard and beyond what you have done the payoff isn't huge.

BTW because I am lazy and I would prefer to keep the internals of the renderer as simple as possible, I would for both "interleaved geometry support" and "flat normals", just have a pre-processing step that converts them to standard non-interleaved geometry (the opposite of this function: https://github.com/threeify/threeify/blob/master/src/lib/geometry/Geometry.Functions.ts#L28) and with proper vertex-normals for flat surfaces so I do not have to make exceptions in the engine.  People could run that immediately after loading a glTF.  Basically a data normalization step.  Otherwise it would skip them and warn that the geometry needs to be normalized or something like that.

gkjohnson:
Thanks!

> The highest priority are the glaring differences with sheen and tangents

I'm not exactly sure why the tangents aren't working as expected. They're required to be generated ahead of time using MikkT or already present in the model so I'm not sure if that section on shader-generated tangents is relevant there.

This line from the spec seems to describe what I'm already doing - but I'll have to dig in more:

> The bitangent vectors MUST be computed by taking the cross product of the normal and tangent XYZ vectors and multiplying it against the W component of the tangent: bitangent = cross(normal.xyz, tangent.xyz) * tangent.w.

https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html

--

> BTW because I am lazy and I would prefer to keep the internals of the renderer as simple as possible, I would for both "interleaved geometry support" and "flat normals", just have a pre-processing step that converts them to standard non-interleaved geometry (the opposite of this function

[...]

PR #256, opened 2022-08-24
Fix dark diffuse values

### LINKED ISSUE #250: Multiple views
Hi everybody!

I've been trying to render the vanilla [multiple views example](https://threejs.org/examples/?q=multiple#webgl_multiple_views) ([source](https://github.com/mrdoob/three.js/blob/master/examples/webgl_multiple_views.html)) with the PathTracingRenderer and I have made some strides in my `animate` loop but I haven't found a way of setting the scissorTest on the PTR. It doesn not appear to fully extend the three-js raster renderer in that regard.

Has anybody figured out a way of only rendering a portion of the canvas, based on a view, with the PTR? 

I do not think having multiple renderers is a satisfactory, nor efficient solution 😄

gkjohnson:
Related to #210
Fix #250

**TODO**
- [x] Comparisons
- [x] Ensure check the metalness case, too
- [ ] Move weights to consolidate them

### DIFF
--- a/example/areaLight.js
+++ b/example/areaLight.js
@@ -17,19 +17,19 @@ const params = {
 
 	areaLight1Enabled: true,
 	areaLight1IsCircular: false,
-	areaLight1Intensity: 5,
+	areaLight1Intensity: 2,
 	areaLight1Color: '#ffffff',
 	areaLight1Width: 1,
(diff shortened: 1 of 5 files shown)


gkjohnson:
image

PR #257, opened 2022-08-30
Add Vertex Color Support

### LINKED ISSUE #182: Add support for vertex colors

alvinalexander:
# Summary
- Pack color attribute data in to Float Texture
- Update material struct with vertexColors flag
- Interpolate colors across faces

TODO: 
- [ ] Vertex colors are assumed to be RBGA Uint16. The data must be normalized before it's accessed in the fragment shader. Where should I put this normalization code? 

closes #182 

Screen Shot 2022-08-30 at 9 01 59 AM


example/index.js
```diff
@@ -681,9 +681,57 @@ async function updateModel() {
 		material.tangentAttribute.updateFrom( geometry.attributes.tangent );
 		material.uvAttribute.updateFrom( geometry.attributes.uv );
 		material.materialIndexAttribute.updateFrom( geometry.attributes.materialIndex );
+
```
alvinalexander, line 684: @gkjohnson Should this code go into three-mesh-bvh ? https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/gpu/VertexAttributeTexture.js

example/index.html
```diff
@@ -279,6 +279,10 @@
 					url: 'https://raw.githubusercontent.com/gkjohnson/3d-demo-data/main/models/botanists-greenhouse/scene.gltf',
 					credit: 'Model by "riikkakilpelainen" on Sketchfab.',
 				},
+				'Vertex Colors Test': {
+					url: 'https://raw.githubusercontent.com/alvinalexander/3d-demo-data/main/models/vertex-color-cube.glb',
+					credit: 'Model by aghouas on Sketchfab.',
+				},
```
gkjohnson, line 285: I think instead of adding a new test model to the main demo we can use the model-viewer page for testing. See these two model pages here which can be run locally:

- https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/viewerTest.html#khronos-BoxVertexColors
- https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/viewerTest.html#khronos-VertexColorTest

src/materials/PhysicalPathTracingMaterial.js
```diff
@@ -215,6 +218,10 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 
 							}
 
+							if( material.vertexColors ){
+								albedo *= vertexColor;
+							}
```
gkjohnson, line 223: nit: can we add spaces inside the braces and before the first brace to keep in style with the file. Looks like there are some other places in the shaders that need to be updated, as well.

src/uniforms/MaterialsTexture.js
```diff
@@ -349,7 +349,7 @@ export class MaterialsTexture extends DataTexture {
 			// sample 13
 			index ++; // matte
 			index ++; // shadow
-			index ++;
+			floatArray[ index ++ ] = + m.vertexColors; // vertexColors
```
gkjohnson, line 352: nit: For better legibility I'd prefer to use `Number( m.vertexColors )` instead of `+ m.vertexColors`.

example/index.js
gkjohnson: When using the FloatVertexAttributeTexture the values should already be normalized if using an int or uint buffer:

https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/gpu/VertexAttributeTexture.js

gkjohnson (commented):
Nice work! It looks like the `viewerTest` files are hitting an error when rendering so I'd take a look at those examples. But code is looking pretty good.

alvinalexander:
@gkjohnson Thanks for the review. The last few weeks have been very hectic, but I'll get to it this week:)

alvinalexander:
@gkjohnson  I ran the model viewer test locally and I noticed a couple things: 

- In khronos-VertexColor, colors are stored as normalized float 32 values. So should we assume that vertex colors are already normalized (0-1) and stored as float32? I know blender uses Uint16 by default for vertex colors in gltf,  so if we assume float 32, these models will not render correctly.

- As per [ this discussion](https://stackoverflow.com/questions/50896963/reading-gltf-vertex-color-encoding), it seems we need to account for the material `alphaMode` to render colors correctly.  According to [emackey](https://github.com/emackey), vertex color transparency turns off the z-buffer in most standard WebGL renderers, so [he disabled alphaMode for khronos-VertexColor](https://github.com/KhronosGroup/glTF-Sample-Models/commit/5fab250e38c8320598d971d6729f081b94816343). I tried ignoring the alpha channel in my local code and that seems to work.  So should we just ignore the vertex color alpha channel for now, regardless of the material blendmode?

gkjohnson:
> In khronos-VertexColor, colors are stored as normalized float 32 values. So should we assume that vertex colors are already normalized (0-1) and stored as float32? I know blender uses Uint16 by default for vertex colors in gltf, so if we assume float 32, these models will not render correctly.

There's no need to assume. As mentioned in https://github.com/gkjohnson/three-gpu-pathtracer/pull/257#discussion_r961081602 passing an Int array into into the [VertexAttributeTexture](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/gpu/VertexAttributeTexture.js) will automatically normalize the values to [ - 1.0, 1.0 ] (or [ 0.0, 1.0 ] for unsigned). You can also pass in a FloatArray and it will be represented as-is. So basically you can plug a color array in with no changes either way and it should work correctly.

> As per [this discussion](https://stackoverflow.com/questions/50896963/reading-gltf-vertex-color-encoding), it seems we need to account for the material alphaMode to render colors correctly ( or just ignore alpha values entirely). According to [emackey](https://github.com/emackey), vertex color transparency turns off the z-buffer in most standard WebGL renderers, so [he disabled alphaMode for khronos-VertexColor](https://github.com/KhronosGroup/glTF-Sample-Models/commit/5fab250e38c8320598d971d6729f081b94816343). I tried ignoring the alpha channel in my local code and that seems to work. So should we just ignore the vertex color alpha channel for now, regardless of the material blendmode?

For the Vertex Color alpha we can just treat it as opacity and multiply it into the "albedo" value [here](https://github.com/gkjohnson/three-gpu-pathtracer/blob/8b502b0d994ac61f3c550df9c1a2e0df8d8fec8e/src/materials/PhysicalPathTracingMaterial.js#L210-L216). Supporting different blend modes for rendering is outside the scope of this feature, I think, and I don't even feel it's something we should be supporting in the project necessarily.

Thanks!

alvinalexander:
> For the Vertex Color alpha we can just treat it as opacity and multiply it into the "albedo" value [here](https://github.com/gkjohnson/three-gpu-pathtracer/blob/8b502b0d994ac61f3c550df9c1a2e0df8d8fec8e/src/materials/PhysicalPathTracingMaterial.js#L210-L216). Supporting different blend modes for rendering is outside the scope of this feature, I think, and I don't even feel it's something we should be supporting in the project necessarily.

Got it! 

I see your point about not supporting different blend modes. Just note that if we treat the alpha as opacity, the  khronos-VertexColor model will render like this: 
Screen Shot 2022-09-25 at 9 32 08 AM

As opposed to this: 
Screen Shot 2022-09-25 at 9 31 09 AM

gkjohnson:
It looks like three.js will only render transparency when `material.transparent = true` which is why the vertex alphas aren't respected in the three.js forward rendering pipeline. If you set `material.transparent` to true then this is what you get - so the implementation you have seems to match that case pretty well!



But we'll want to replicate that `.transparent` flag behavior in the path tracer, as well. I think we can do that by adding a "transparent" flag to the material definitions and only respecting opacity if it's set to "true". I say we can get this merged as-is right now and then you can make another PR with the changes required for the "transparent" flag? What do you think?

src/materials/PhysicalPathTracingMaterial.js
```diff
@@ -188,6 +190,7 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 							// TODO: should we account for emissive surfaces here?
 
 							vec2 uv = textureSampleBarycoord( uvAttribute, barycoord, faceIndices.xyz ).xy;
+							vec4 vertexColor = textureSampleBarycoord( colorAttribute, barycoord, faceIndices.xyz);
```
gkjohnson, line 193: nit: can we add a space inside the last paren

src/materials/PhysicalPathTracingMaterial.js
```diff
@@ -513,6 +522,7 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 						}
 
 						vec2 uv = textureSampleBarycoord( uvAttribute, barycoord, faceIndices.xyz ).xy;
+						vec4 vertexColor = textureSampleBarycoord( colorAttribute, barycoord, faceIndices.xyz);
```
gkjohnson, line 525: nit: space inside the last paren

gkjohnson (commented):
Two little lint issues and we can get this merged!

alvinalexander:
> It looks like three.js will only render transparency when `material.transparent = true` which is why the vertex alphas aren't respected in the three.js forward rendering pipeline. If you set `material.transparent` to true then this is what you get - so the implementation you have seems to match that case pretty well!
> 
> 
> 
> But we'll want to replicate that `.transparent` flag behavior in the path tracer, as well. I think we can do that by adding a "transparent" flag to the material definitions and only respecting opacity if it's set to "true". I say we can get this merged as-is right now and then you can make another PR with the changes required for the "transparent" flag? What do you think?

Yes sounds good!

gkjohnson:
Excellent! Nice work!

And thanks again

PR #283, opened 2022-10-16
Physics material page

### LINKED ISSUE #252: Add a viewer for PBR database materials
https://twitter.com/80Level/status/1557041514798129152
https://physicallybased.info/
https://github.com/AntonPalmqvist/physically-based-api

Do the properties map directly to glTF properties?

- Provide a drop down list of material options from the site
- Possibly provide a side-by-side image view for comparison
- Similar to the [model viewer test page](https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/viewerTest.html#khronos-DragonAttenuation)
- Use the same environment map as in the examples

gkjohnson:
Fix #252

image

image

image

image

### DIFF
--- a/example/materialDatabase.js
+++ b/example/materialDatabase.js
@@ -334,7 +334,8 @@ function applyMaterialInfo( info, material ) {
 
 	}
 
-	imgEl.src = `https://physicallybased.info/reference/render/${ info.name.replace( /\s+/g, '-' ) }-cycles.webp`;
+	const cleanName = info.name.replace( /\s+/g, '-' ).replace( /[()]+/g, '' );
+	imgEl.src = `https://physicallybased.info/reference/render/${ cleanName }-cycles.webp`;
 	materialEl.innerText = `${ info.description }`;
 	materialEl.style.display = info.description ? 'inline-block' : 'none';
 
--- a/example/materialDatabase.js
+++ b/example/materialDatabase.js
@@ -334,9 +334,6 @@ function applyMaterialInfo( info, material ) {
 
 	}
 
-	window.INFO = info;
-	window.MAT = material;
-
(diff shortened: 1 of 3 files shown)


example/materialDatabase.js
```diff
@@ -0,0 +1,395 @@
...
+	if ( info.specularColor ) material.specularColor.setRGB( ...info.specularColor );
+	if ( 'metalness' in info ) material.metalness = info.metalness;
+	if ( 'roughness' in info ) material.roughness = info.roughness;
+	if ( 'ior' in info ) material.ior = info.ior;
+	if ( 'transmission' in info ) material.transmission = info.transmission;
+
+	if ( material.transmission ) {
+
+		if ( info.color ) material.attenuationColor.setRGB( ...info.color );
+		material.attenuationDistance = 200 / info.density;
```
gkjohnson, line 329: cc @AntonPalmqvist

Do you have any recommendations on references for converting material density to three.js' / glTF "attenuationDistance"?

AntonPalmqvist: To my knowledge, there's no reliable way to convert physical density to attenuation distance.

The reason I've included density in my database is only for physics simulation purposes, such as rigid body dynamics.

gkjohnson: Oh gotcha - thanks! So for your renders how are you determining the light attenuation to use for the materials? Is there just a constant attenuation used? Or is it just a surface color that tints all transmitted light?

AntonPalmqvist: I used Blender Cycles to render the shader balls, with the standard Principled BSDF (Disney). I just set the Base Color to the color of the material, and then I set Transmission to 1. There's no attenuation controls in the shader.

I should use another render engine, with more control, but Blender was the quickest and easiest way to do those initial renders.

gkjohnson: Oh okay thats good information. I'll leave it at the attenuation method, for now, and maybe revisit later.

PR #291, opened 2022-11-10
Update package.json

SergiiSharpov:
Correctly resolve npm dependencies.
Fixes errors during `npm install` in other projects.

UPD:

Reason:
1. Start a new project
2. Npm i three@latest
3. Npm i three-gpu-pathtracer

3rd item will throw error because this project explicitly defines an exact version of threejs in peer dependencies, but at the same time uses almost the latest threejs version.

Solution:
Allow not an exact version of threejs in peer dependencies, but a version greater than N.
That is done in that PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "yargs": "^17.5.1"
   },
   "peerDependencies": {
-    "three": "^0.139.2",
+    "three": ">=0.139.2",
     "xatlas-web": "^0.1.0"
   },
   "scripts": {


gkjohnson:
This doesn't seem correct. A caret (`^`) should allow a compatible version of the library (minor version increase) on installation:

https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json

SergiiSharpov:
@gkjohnson

image

As you can see, it can not resolve the `0.139.2` version from your project, but with `>=` it works.

Why it works in that way you can check here:
1. https://semver.npmjs.com/

And a bit notes here:
2. https://nodejs.org/es/blog/npm/peer-dependencies/

Shortly, from 1st link:
`Caret behavior is different for 0.x versions, for which it will only match patch versions.`
That is what we have for threejs.

So this semver - `^0.139.2`
Will work for `0.139.3`
But not will work for `0.140.0`

SergiiSharpov:
Anyway, replacing a caret with any other form of a semver conditions will work here, except the caret.

gkjohnson:
Oh yes - thank you for explanation. I forgot tilde only worked for patch increase when the major version is 0.

Are you able to work around this or would you like me to make a new release?

SergiiSharpov:
No probs, I'm using my fork right now. (that's good that NPM gives you the ability to use `npm i githubRepoUrl`)
Or even I can use now `npm i yourGithubRepo` :)

I'll switch to the npm version right after you release a new version.

PR #295, opened 2022-11-22
Add Gradient Textures

### LINKED ISSUE #148: Replace support for gradient backgrounds with env map texture
Related to #48

If gradients are going to be supported / used for backgrounds / environments it would simplify things to just support them via a texture rather than interpolated uniforms in the shader. A `GradientEnvMapTexture` could be written that lets you set colors and update the texture contents as needed. Long term different falloff functions or gradients could be used to generate the map.

```js
const envMap = new GradientEnvMapTexture();
envMap.resolution = 10;
envMap.topColor.set( 0xff0000 );
envMap.botColor.set( 0x000000 );
envMap.update();
```

gkjohnson:
Fix #148 

**TODO**

- [ ] Use an envmap for the background
- [ ] Remove GRADIENT defines
- [ ] Auto use background map if defined


gkjohnson:
**Textures**
1. BVH.index;
2. BVH.position;
3. BVH.bvhBounds;
4. BVH.bvhContents;

5. EnvInfo.marginalWeights;
6. EnvInfo.conditionalWeights;
7. EnvInfo.map;

8. backgroundMap
9. normalAttribute;
10. tangentAttribute;
11. uvAttribute;
12. colorAttribute;
13. materialIndexAttribute;
14. materials;
15. iesProfiles;
16. textures;

gkjohnson:
16 -> 13 sampler units

PR #302, opened 2022-11-27
Only traverse visible objects and scenes

ikeough:
I think it makes sense for a renderer to only traverse visible objects. Although perhaps you also want to traverse lights that aren't marked as visible. In that case, having a parameter like `onlyVisible` on the `generate` function could allow us to configure this.

### DIFF
--- a/src/core/PathTracingSceneGenerator.js
+++ b/src/core/PathTracingSceneGenerator.js
@@ -13,7 +13,7 @@ export class PathTracingSceneGenerator {
 
 		for ( let i = 0, l = scene.length; i < l; i ++ ) {
 
-			scene[ i ].traverse( c => {
+			scene[ i ].traverseVisible( c => {
 
 				if ( c.isSkinnedMesh || c.isMesh && c.morphTargetInfluences ) {
 


gkjohnson:
Thanks! I think this makes sense. I'm thinking that `DynamicPathTracingSceneGenerator` should behave the same then, as well. Do you mind taking a look at that one, too? It may be the case that we have to require that the user not adjust the object visibility between "generate" calls. Or we cache initial visibility so we can guarantee consistency in that case. I'm curious as to your thoughts.

ikeough:
Should have provided some context with this PR :) Our application (hypar.io) generates models of building systems and presents them in one 3d view, with each building system corresponding to one scene in three, all children of the main scene. We allow the user to hide and show scenes a lot.

The best case for us would be to generate once, and be able to hide/show various sub-scenes without having to recompute the bvh. During rendering you'd just skip surfaces of elements in hidden scenes as if they weren't there at all. This might be totally impossible - I haven't explored the rest of the rendering code deeply enough to know.

The provided fix was a hack to get our use case to work, but it still requires a re-generation when a user changes sub-scene visibility.

Im happy to implement this in both places for now, if it makes sense, and strategize, with your guidance, about how to best approach minimal regeneration on sub-scene visbility switching.

gkjohnson:
> The best case for us would be to generate once, and be able to hide/show various sub-scenes without having to recompute the bvh

Gotcha - thanks for the context. If you'd like a more dynamic scene (ability to move objects) then the best solution right now is the `DynamicPathTracingSceneGenerator` which regenerates the geometry in place and refits the BVH for faster generation.

And to hide individual objects the best solution is to turn "opacity" to 0 for the material on the object you want to hide (with the dynamic or regular scene generator). But note that the transparent faces are still considered during the traversal so there is a performance impact. In the long term it would be nice to have a "scene" spatial index with objects that can more easily be moved and skipped during traversal but I don't have immediate plans to work on that at the moment.

Also I'm just realizing that the [StaticGeometryGenerator](https://github.com/gkjohnson/three-mesh-bvh/blob/master/src/utils/StaticGeometryGenerator.js) in three-mesh-bvh would need to be changed, as well. Let me think about this a bit more.

src/core/DynamicPathTracingSceneGenerator.js
```diff
@@ -6,19 +6,19 @@ export class DynamicPathTracingSceneGenerator {
 
 	get initialized() {
 
-		return Boolean( this.bvh );
+		return Boolean(this.bvh);
```
gkjohnson, line 9: Small note but it looks like a lot of the code style got changed in a commit.

ikeough:
Ack! Sorry. Pesky linters. Should be fixed now.

gkjohnson:
Okay thanks! I made the necessary change to `StaticGeometryGenerator` in three-mesh-bvh v0.5.19. So I'll merge this, as well, and make a couple changes. The change will be available in the next release.

If you have the opportunity to show what you're working on I'd be interested to see!

Issue #325, opened 2022-12-12
All examples crash on Nvidia RTX 2080 Ti


haroldiedema:
**Describe the bug**

Any example just renders one frame, or mostly none at all, then freezes the browser for a 5-8 seconds, after which the WebGL context is lost.

**Expected behavior**

It shouldn't crash 😄 

**Screenshots and Repro Model**

The following WebGL error appears in the console. It's similar for all examples.

```
three.module.js:18826 THREE.WebGLProgram: Program Info Log: C:\fakepath(3420,10-36): warning X3577: value cannot be NaN, isnan() may not be necessary.  /Gis may force isnan() to be performed
C:\fakepath(3420,10-36): warning X3577: value cannot be NaN, isnan() may not be necessary.  /Gis may force isnan() to be performed
C:\fakepath(1289,1-6): warning X4000: use of potentially uninitialized variable (dyn_index_vec3_int)
error X8000: D3D11 Internal Compiler Error: Invalid Bytecode: Index Dimension 1 out of range (4263 specified, max allowed is 31) for operand #2 of opcode #4322 (counts are 1-based). Aborting.
error X8000: D3D11 Internal Compiler Error: Invalid Bytecode: Can't continue validation - aborting.

Warning: D3D shader compilation failed with default flags. (ps_5_0)
 Retrying with skip validation
C:\fakepath(3420,10-36): warning X3577: value cannot be NaN, isnan() may not be necessary.  /Gis may force isnan() to be performed
C:\fakepath(3420,10-36): warning X3577: value cannot be NaN, isnan() may not be necessary.  /Gis may force isnan() to be performed
C:\fakepath(1289,1-6): warning X4000: use of potentially uninitialized variable (dyn_index_vec3_int)
error X8000: D3D11 Internal Compiler Error: Invalid Bytecode: Index Dimension 1 out of range (4263 specified, max allowed is 31) for operand #2 of opcode #4322 (counts are 1-based). Aborting.
error X8000: D3D11 Internal Compiler Error: Invalid Bytecode: Can't continue validation - aborting.

Warning: D3D shader compilation failed with skip validation flags. (ps_5_0)
 Retrying with skip optimization
C:\fakepath(3420,10-36): warning X3577: value cannot be NaN, isnan() may not be necessary.  /Gis may force isnan() to be performed
C:\fakepath(3420,10-36): warning X3577: value cannot be NaN, isnan() may not be necessary.  /Gis may force isnan() to be performed
[...]

PR #350, opened 2023-01-05
GPU bug workaround. Doesn't pass float arrays as parameters.

mattni:
Hi! Thank you for your brilliant work!

As many has reported, examples chrash on many gpu's since v.0.08. 

Related issues:
https://github.com/gkjohnson/three-gpu-pathtracer/issues/325

I got same shader error using Lenovo laptop (intel gpu)
and using pc (GeForce RTX 3070 Ti).
 
I debugged commits and found that the error comes when passing array of floats to function.
For some reason shader doesn't compile when passing float arrays as function parameter. 

Because this affects so many users, maybe better to pass floats separately?

### DIFF
--- a/src/shader/shaderMaterialSampling.js
+++ b/src/shader/shaderMaterialSampling.js
@@ -303,11 +303,7 @@ vec3 sheenColor( vec3 wo, vec3 wi, vec3 wh, SurfaceRec surf ) {
 }
 
 // bsdf
-#define DIFF_WEIGHT 0
-#define SPEC_WEIGHT 1
-#define TRANS_WEIGHT 2
-#define CC_WEIGHT 3
-void getLobeWeights( vec3 wo, vec3 wi, vec3 wh, vec3 clearcoatWo, SurfaceRec surf, out float[ 4 ] weights ) {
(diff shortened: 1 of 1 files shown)


gkjohnson:
This is great - thanks @mattni! I appreciate the effort in tracking this down. At the end of the day this is something that should be fixed by the browser vendors.

Do you mind making a comment on the [Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1404661) mentioning what you've found? Depending on what their response is maybe we can merge this as a temporary fix since it is pretty egregious 😅

src/shader/shaderMaterialSampling.js
```diff
@@ -423,20 +413,37 @@ float bsdfEval( vec3 wo, vec3 clearcoatWo, vec3 wi, vec3 clearcoatWi, SurfaceRec
...
 
-	float[ 4 ] pdf;
 	vec3 wh = getHalfVector( wo, wi, surf.eta );
-	getLobeWeights( wo, wi, wh, clearcoatWo, surf, pdf );
+	float diffuseWeight;
+	float specularWeight;
+	float transmissionWeight;
+	float clearcoatWeight;
+	getLobeWeights( wo, wi, wh, clearcoatWo, surf, diffuseWeight, specularWeight, transmissionWeight, clearcoatWeight );
+	float[ 4 ] pdf;
```
kenrussell, line 422: This temporary array is unnecessary - it's unused.

PR #376, opened 2023-02-04
Add quilt renderer for LKG

### LINKED ISSUE #284: Investigate Looking Glass Display Rendering
- raytracing directly to the Looking Glass: https://www.shadertoy.com/view/ttXSDN

- orthographic / perspective distortion using ray tracing: https://www.shadertoy.com/view/3tBGDR

**Plan**
- Use "Quilt" method to render out multiple frames per render target for LKG use.
- Add a "subframe" parameter to PathTracingRenderer to specify which quilt frame to render out.
- Use many path tracing renderers that share a material but use different cameras.
- Add an "LKGPathTracingRederer" that extends "PathTracingRenderer" and performs updates with multiple cameras / quilts.

**Questions**
- How does "toe in" or off-axis frustums affect path tracing? (likely just near / far clip plane angle, ray angles) (confirm path tracer works with off-axis frustum)
- How do we pick the frames for each view? What are the parameters? (see [here](https://docs.lookingglassfactory.com/keyconcepts/camera))
  - View cone
  - View count
  - FoV / viewer distance
  - Display size

gkjohnson:
Fix #284 

**TODO**
- Test on LKG
- Fix resolution scale rendering with tiles (0.42 breaks with 4x4 tiles)
- Add utility replicate NPOT texture size generation from image scale and tile count
- Write the texture as a subframe to the final 4k buffer


gkjohnson:
image

PR #399, opened 2023-02-23
Add fog

### LINKED ISSUE #360: Add "volume" support for constant density fog
- Fog color
- Density
- https://raytracing.github.io/books/RayTracingTheNextWeek.html#volumes/constantdensitymediums
- https://www.youtube.com/watch?v=jYym08606mw

**Strategy**
- Make a "Volume" material
- Track when a volume is entered and then exited.
- If in fog check for early particle hit and scatter.
- assume if we exit a fog volume that we've entered a fog-free zone.
- Handle direct light sampling case?
  - Perform raycast and assume shadow if we hit a particle
- Before casting we must determine if the ray origin is within a fog volume which requires a modified bvh "first hit" function.

**Notes**
- Overlapping fog volumes will not work correctly

gkjohnson:
Fix #360 

**TODO**
- [ ] Code clean up
- [x] Get punctual lights working
- [x] Ensure perf consistency
- [x] Add more define gates
- [x] Material does not work unless it's physical/standard material
  - even inheriting these materials breaks the render
- [x] Ensure MIS and non MIS look consistent
  - Does forward path sampling need to be adjusted?
- [x] Add demo
  - spot light slats with fog
- [x] Get emission working 
- [x] Adjust bsdfResult function (or just move fog logic to eval?)
- [x] Account for fog in attenuate hit
- [x] Treat the particle hit as a geometric hit so direct lighting will work
- [x] If in fog check for early particle hit and scatter.
- [x] Make a "Volume" material
- [x] Track when a volume is entered and then exited.
- [x] assume if we exit a fog volume that we've entered a fog-free zone.
- [x] Before casting we must determine if the ray origin is within a fog volume which requires a modified bvh "first hit" function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -958,6 +958,6 @@ Set of randomness and other light transport utilities for use in a shader. See t
 
 [knightcrawler25/GLSL-PathTracer](https://github.com/knightcrawler25/GLSL-PathTracer/)
 
-[DassaultSystemes Path Tracer](https://github.com/DassaultSystemes-Technology/dspbr-pt)
+[DassaultSystemes-Technology/dspbr-pt](https://github.com/DassaultSystemes-Technology/dspbr-pt)
 
 
--- a/README.md
+++ b/README.md
@@ -958,4 +958,6 @@ Set of randomness and other light transport utilities for use in a shader. See t
 
 [knightcrawler25/GLSL-PathTracer](https://github.com/knightcrawler25/GLSL-PathTracer/)
 
+[DassaultSystemes Path Tracer](https://github.com/DassaultSystemes-Technology/dspbr-pt)
+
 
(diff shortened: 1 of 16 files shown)


gkjohnson:
image

gkjohnson:
Unlikely bright particle hits - maybe from a random ray that hits a particle right underneath the light causing it to receive a  lot of energy. Perhaps something like the "roughness" attenuation needs to be used here?

image

View angle dependence with spot lights:
image

And fog is causing a strange edge only with MIS:

image

gkjohnson:
Example:

image

gkjohnson:
image

gkjohnson:
image

PR #404, opened 2023-02-27
Faster env-map MIS calculation, use FloatType when loading env-maps with RGBELoader

0beqz:
By default, the RGBELoader uses `HalfFloatType` to load HDRs:
https://github.com/mrdoob/three.js/blob/dc9570226cf8e541cc71dfebf16f19935dabb74d/examples/jsm/loaders/RGBELoader.js#L19

The HDRs from polyhaven (which are used in the examples) have 32-bits per channel. This fixes the issue of losing precision when loading a 32-bit HDR with the RGBELoader by setting its data type to `FloatType`.

The other fix speeds up the `binarySearchFindClosestIndexOf` function in `EquirectHdrInfoUniform` by changing some offsets and using a different way to calculate the `mid` variable. The most significant optimization (thanks to ChatGPT 😄) is changing
```javascript
const mid = ~ ~ ( 0.5 * upper + 0.5 * lower );
```

to

```javascript
const mid = ( lower + upper ) >> 1;
```

For comparison, updating `EquirectHdrInfoUniform` for a 4096x2048 HDR on my CPU took ~900ms using the old method. With the new method it's ~400ms.

Besides that I could also move the calculation of the weights for MIS in `EquirectHdrInfoUniform` to a worker. For small environment maps (e.g. 1K) it doesn't seem to make sense because without workers I get ~30ms and with workers I get ~120ms. Looks like moving the env map image data as well as the weights arrays between threads is the bottleneck. For higher HDRs it would make sense since calculating the weights on the main thread would freeze the application for hundreds of ms.
Let me know what you'd think of using workers, I'd have a ready implementation.

### DIFF
--- a/src/uniforms/EquirectHdrInfoUniform.js
+++ b/src/uniforms/EquirectHdrInfoUniform.js
@@ -2,15 +2,16 @@ import { DataTexture, FloatType, RedFormat, LinearFilter, DataUtils, HalfFloatTy
 
 function binarySearchFindClosestIndexOf( array, targetValue, offset = 0, count = array.length ) {
 
-	let lower = 0;
-	let upper = count - 1;
+	let lower = offset;
+	let upper = offset + count - 1;
+
 	while ( lower < upper ) {
 
-		const mid = ~ ~ ( 0.5 * upper + 0.5 * lower );
+		const mid = ( lower + upper ) >> 1;
 
 		// check if the middle array value is above or below the target and shift
 		// which half of the array we're looking at
-		if ( array[ offset + mid ] < targetValue ) {
+		if ( array[ mid ] < targetValue ) {
 
 			lower = mid + 1;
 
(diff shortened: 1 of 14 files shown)


src/uniforms/EquirectHdrInfoUniform.js
```diff
@@ -2,15 +2,16 @@ import { DataTexture, FloatType, RedFormat, LinearFilter, DataUtils, HalfFloatTy
...
 
-	let lower = 0;
-	let upper = count - 1;
+	let lower = offset;
+	let upper = offset + count - 1;
+
 	while ( lower < upper ) {
 
-		const mid = ~ ~ ( 0.5 * upper + 0.5 * lower );
+		const mid = ( lower + upper ) >> 1;
```
gkjohnson, line 10: Can we add a comment here stating what this does? Basically truncating the average value of `lower` and `upper`.

gkjohnson:
Great, thank you!

src/uniforms/EquirectHdrInfoUniform.js
0beqz: Done, I've added comments to it

PR #412, opened 2023-03-07
Add detectors

gkjohnson:
**TODO**
- [ ] Test
- [ ] Add worker variant if offscreen canvas is available
- [ ] Docs


gkjohnson:
Shader compilation is still blocking in workers.

PR #416, opened 2023-03-12
add .js to glsl import path

frading:
Hi Garrett,

I'm upgrading three-gpu-pathtracer in polygonjs, and got an error where a .glsl file import path was not found. I believe this is due because my webpack config is set to import `.glsl` files as raw, and it will therefore not try and find a matching `.glsl.js` file.

So this PR adds the `.js` suffix to match with the other import statements.


gkjohnson:
Oops 🙄

Thanks! I'll put this in the next release in the next few weeks. Unfortunately it doesn't look like there's an easy ESLint rule to catch things like this...

frading:
Great, thank you for merging!

PR #422, opened 2023-03-15
Add debug mode flag to show expense of the material

gkjohnson:
**TODO**
- [x] Confirm the accumulated results (read back pixels)

Note it looks like sometimes the scatter ray is going below the surface with very rough materials:

image


gkjohnson:
image

image

PR #438, opened 2023-05-18
add PhysicalCamera.copy

frading:
Hi Garrett,

In Polygonjs I make heavy use of `Object3D.copy` when transferring objects between nodes. So in this case, I've added `.copy` on `PhysicalCamera` to make sure its properties are properly carried down the node tree.

Hope that doesn't had an unnecessary method for you.

### DIFF
--- a/src/objects/EquirectCamera.js
+++ b/src/objects/EquirectCamera.js
@@ -10,11 +10,12 @@ export class EquirectCamera extends Camera {
 
 	}
 
-	copy(source, recursive) {
-		super.copy(source, recursive);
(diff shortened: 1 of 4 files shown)


gkjohnson:
Thanks! Might be worth adding a copy function to `EquirectCamera`, `PhysicalSpotLight`, and `ShapedAreaLight`, as well.

frading:
ah yes, completely missed those. But just added now.

gkjohnson (approved):

gkjohnson:
Thank you!

Issue #450, opened 2023-09-06
Broke on MacOS Sonoma: Chrome + Safari


bhouston:
**Describe the bug**

I am testing with this URL: https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/index.html

There is now a very long pause after the WebgL displays because it tries to show the ray traced image.  This occurs on all machines, but on MacOS M1 and M2 devices both Safari and Chrome crash hard.

When trying to rendering, it just crashes with a ContextLost on Safari, screenshot here:
<img width="1279" alt="Screenshot 2023-09-06 at 6 48 56 AM" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/588541/3b360148-7d46-48ad-a6fd-0dce5eb9ec90">

On Chrome it goes white when it crashes - takes about 30s after it finishes loading before it crashes hard:

<img width="1595" alt="Screenshot 2023-09-06 at 6 51 11 AM" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/588541/e69d5e59-6e48-4c42-b738-08ae983b8af6">

LGL Tracer works on hthese machines fine .

I am running the betas of MacOS Senoma - this is an advanced beta and maybe what is eventually released.  I believe it also crashes on other versions of MacOS as well, but I haven't been able to confirm.

**Expected behavior**

It should work and there should be no pauses during loading.

**Screenshots and Repro Model**

Just the base renderer with default models crashes.

**Platform:**

 - Device: Desktop MasOC
 - OS: MacOS
 - GPU: M1 and M2
 - Browser: Chrome + Safari
 - Three.js version: whatever is currently live on the website.
 - Library version: whatever is currently live on the website.

bhouston:
I was able to work around the bug on Safari and Chrome - I just had to disable MIS.  Then it works.  So something in the MIS code is causing the crash.

gkjohnson:
Also possibly related to #450, #439

JerryI:
I also tried M1 (Air, Sonoma)
- Chrome (Electron) just crash
- Firefox infinite loading
- Safari freezes and crashes

Hope someone will find the solution. M1 is quite powerful, could be great to utilize it

gkjohnson:
Fixed in #483!

Issue #439, opened 2023-06-05
Safari compatibility on Mac & iOS


ChazHollywood:
I haven't been able to get the pathtracer to work at all in Safari on Mac or iOS, both updated to the latest versions as of today. Everything works fine on Chrome. The problem occurs whether I'm using the Github hosted demos or from a local server. 

One clue is the black "Loading" page does appear, however it fails just before the loading percentage begins.

This problem does not occur with the "skinned geometry support" or "morph target support" demo's. All other demos cause the error: 

A problem repeatedly occurred with "https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/basic.html"
[Reload Webpage]

Because of the nature of the error it results in no page loading into the tab and therefore no way that I know of to look for errors.

Mac is a MacBook Pro 16" M1 and the phone is a 14 Pro.

<img width="541" alt="Screenshot 2023-06-05 at 8 42 34 AM" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/135338745/355c3749-3050-48ef-835d-1203bd17d4ec">

gkjohnson:
See https://github.com/gkjohnson/three-gpu-pathtracer/issues/441#issuecomment-1610640494

gkjohnson:
Fixed in #483

PR #457, opened 2023-10-11
Replace 'lights' param with global uniform in 'traceScene'. Closes #456

### LINKED ISSUE #456: Can't see the examples.
Hello, want to check out the examples, the shader errors occur.

I wonder the reason of those errors.

Screenshot 2023-10-06 at 3 40 56 PM

**Platform:**

 - Device: M1 Mac
 - OS: Ventura 13.3.1
 - GPU: [NVidia ??, Radeon ??, ...]
 - Browser: Chrome
 - Three.js version: 0.157.0
 - Library version: 0.0.16

oh-jinsu:
Seems weird, but I figured out the `lights` param in `traceScene` causes the error. (I've checked the shaders are compiled well, but after `gl.linkProgram()`, the `LINK_STATUS` is false. (As I reported to the linked issue, in my device Macbook Pro 13, 2020)

In the shader, only two functions call `traceScene()`. The one in `main` function works but when calling `traceScene()` in `attenuateHit()` with the `lights` param, the error occurs. Do you have any idea why this problem happens? I'm still wondering.

It's not a fundamental solution, but I tried removing the parameter and replace it with the global uniform `lights`. Then I could see finally it's working on my device, but I wonder if using the global uniform is fine. Please let me know if you expect any problem with this commit. I'll try to find another solution. Thank you.

### DIFF
--- a/src/materials/pathtracing/PhysicalPathTracingMaterial.js
+++ b/src/materials/pathtracing/PhysicalPathTracingMaterial.js
@@ -301,7 +301,7 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 						state.firstRay = i == 0 && state.transmissiveTraversals == transmissiveBounces;
 
 						int hitType = traceScene(
-							ray, bvh, state.fogMaterial,
+							ray, state.fogMaterial,
 							surfaceHit, lightRec
 						);
 
--- a/src/materials/pathtracing/PhysicalPathTracingMaterial.js
+++ b/src/materials/pathtracing/PhysicalPathTracingMaterial.js
@@ -301,7 +301,7 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 						state.firstRay = i == 0 && state.transmissiveTraversals == transmissiveBounces;
 
 						int hitType = traceScene(
-							ray, bvh, lights, state.fogMaterial,
+							ray, bvh, state.fogMaterial,
 							surfaceHit, lightRec
 						);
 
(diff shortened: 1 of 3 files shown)


gkjohnson:
> Do you have any idea why this problem happens? I'm still wondering.

It's a bit of a mystery - but that's consistent with other fixes for these types of opaque issues. These seem like hiccups in shader compiler but the lack of error entirely makes these frustrating to track down. Part of me is wondering if theres some problem with shaders using the same name as a global variable or passing potentially large large data structures into functions in some cases.

Either way this fixes the problem for me on Safari on my 2021 M1 Macbook Pro. For consistency maybe we should just move any global variables out of function parameters. So why don't we remove "bvh", as well, for this PR and make a comment referencing this PR #457 at the top of the "TraceScene" function so the issue is noted somewhere.

gkjohnson:
Also possibly related to https://github.com/gkjohnson/three-gpu-pathtracer/issues/450, https://github.com/gkjohnson/three-gpu-pathtracer/issues/439

oh-jinsu:
I moved `bvh` out of the function parameter and commented about the issue on `traceScene()`.

Thanks for your review!

gkjohnson:
Great thanks!

I've tested a bit more and, strangely, I'm seeing that even with the fix the extremely simple "primitives.html" example still _doesn't_ work while the more complex "index.html" and "basic.html" examples do. Are you seeing this, as well? I wonder if it has anything to do with textures not being used?

oh-jinsu:
Well, it's weird but in this time it does work on my device after moving out the global variables from the function paramters. I pulled your updated example and tested it:

Screenshot 2023-10-16 at 9 46 57 PM

gkjohnson:
Interesting 🤔 We'll merge this and I'll look into this issue on my machine.

Thanks again for your work on this!

PR #470, opened 2023-11-25
first tentative modifications for "invisible" area lights

robertoranon:


gkjohnson:
The "dimming" of the area where the light surface technically is is caused because hitting the light technically counts as one of the limited "bounces" that a path is allowed to traverse. So the paths that go around the light surface bounce around the scene 3 times while the paths that hit the light only bounce twice since the light absorbs one.

To that end - here's what I'm thinking:
1. We should remove the ability to hit "lights" from the `traceScene` function since this is causing light surfaces to cast shadows and is causing them to "use" path bounce. If this is addressed then the dimming issue you reported should no longer be an issue.
2. Then we should iterate over all lights in the path tracing loop and add any light contribution from lights that are passed through up to the next ray termination point.
3. On the `state.firstRay || state.transmissiveRay` we find the closest light marked as visible we intersect with if any and terminate so it's visible.

It depends on how you want to handle it - but it might be easiest to handle points 1 and 2 first and then 3 can be done in separately. Let me know if you have any other questions or need more direction on the codebase!

robertoranon:
Hi, sorry for taking so long, but I was quite busy. Here is a modification of this PR that removes the ability to trace lights from the `traceScene` function, as well as other modifications:
- removing all code that was executed when `traceScene` was returning a light hit
- some code refactoring using a variable to hold the value to be returned, instead of early return statement, as I have noticed that they "contributed" to the crashes on macOS. I'll now move into point 2 above.

src/materials/pathtracing/glsl/directLightContribution.glsl.js
```diff
@@ -35,13 +35,15 @@ export const directLightContributionGLSL = /*glsl*/`
 					// weight the direct light contribution
 					float lightPdf = lightRec.pdf / lightsDenom;
 					float misWeight = lightRec.type == SPOT_LIGHT_TYPE || lightRec.type == DIR_LIGHT_TYPE || lightRec.type == POINT_LIGHT_TYPE ? 1.0 : misHeuristic( lightPdf, lightMaterialPdf );
-					return attenuatedColor * lightRec.emission * state.throughputColor * sampleColor * misWeight / lightPdf;
+					result = attenuatedColor * lightRec.emission * state.throughputColor * sampleColor * misWeight / lightPdf;
 
-				}
+			 	}
```
gkjohnson, line 40: nit: extra space added between tabs

src/materials/pathtracing/glsl/directLightContribution.glsl.js
```diff
@@ -35,13 +35,15 @@ export const directLightContributionGLSL = /*glsl*/`
...
 
 			}
 
-		} else {
+		}
+		
+		else {
```
gkjohnson, line 46: nit: can we align this code style with the other else brackets?

src/materials/pathtracing/glsl/directLightContribution.glsl.js
```diff
@@ -78,15 +80,15 @@ export const directLightContributionGLSL = /*glsl*/`
...
+					result = attenuatedColor * environmentIntensity * envColor * state.throughputColor * sampleColor * misWeight / envPdf;
 
 				}
 
 			}
 
 		}
 
-		return vec3( 0.0 );
+		return result;
```
gkjohnson, line 91: For the sake of documenting some of these decisions can we add a comment saying that the Mac regressions seem to be related to early return statements? Maybe with a reference to this PR id, as well.

gkjohnson (commented):
Amazing! Thank you. I added a couple small comments around code style and a comment. I can make the changes later this week when I have time, as well, if I get to it first!

PR #474, opened 2023-12-21
missing variable declaration in directLightContribution shader

robertoranon:
Hi, 

  I think I missed a variable declaration in my last PR, here it is.

Roberto


gkjohnson:
Thanks! I saw this as well but forgot to push the change.

Issue #463, opened 2023-10-17
All demos crashing on OSX 14.0


gkjohnson:
Related to https://github.com/gkjohnson/three-gpu-pathtracer/pull/457#issuecomment-1763429214

The basic and beauty renders demos work - but the primitives demo does not.

robertoranon:
hi, I pulled the latest changes, and, to be sure, tried on a just restarted machine (2021 Macbook M1 Pro with M1 Max, 32 Gb, 24 cores GPU). I am using the latest OS (Sonoma 14.0), Safari 17.0, Chrome Version 118.0.5993.70.

On Chrome, if I load index.html, it hangs for a while, and then crashes, strangely leaving the standard three.js rendering "on". But the path-tracer shaders fail to compile, the webgl context is lost, the browser hangs and is irresponsive. 

On Safari, I get no errors, but a grey-black grid (see screenshot), except when I rotate the model, where I get standard three.js rendering.

<img width="1129" alt="Screenshot 2023-10-17 at 09 44 31" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/6480367/4048cf3b-6408-4c32-abfa-dea47a55c632">

As I said in another issue, disabling MIS makes it work on both browsers. Examples that do not use MIS (e.g. aoRender.html) work without modification.

gkjohnson:
I'm on OSX 13.5 so let me upgrade and see what happens 😓

robertoranon:
no pressure :)  btw, great work!

robertoranon:
wow, I don't have that. Safari fails quite gracefully, only Chrome is
kinda frozen (the whole browser, not just the tab). Anyway, maybe it's
possible to check for macos and disable MIS as a temporary workaround? I
am also trying to make it work on iOS, and, besides disabling MIS, I had
some success by substituting the env maps lookups to a fixed color (i.e.
a white dome), I think it's related with the fact that iOS devices,
despite nominally supporting Float32 textures in WebGL, actually behave
nicely only with Float16 ...

On October 17, 2023, "dependabot[bot]" ***@***.***> wrote:
> Yeah I just upgraded and the crash is worse than ever, now, and hangs
> the browser and stalls the OS...
>
> > disabling MIS makes it work on both browsers
> >
>
> In terms of disabling MIS - as I mention in #456
> <https://github.com/gkjohnson/three-gpu-pathtracer/issues/456> this
> really just removes a large swath of code from the shader. It
[...]

Issue #442, opened 2023-06-08
Android Shader compilation error


NickMundel:
**Describe the bug**
When trying to view the webpage on my android device, an error occurs and rendering is halted. This results in a blank page.

A clear and concise description of what the bug is. Before submitting, please remove unnecessary sections.

**Expected behavior**

Well I used the basic example as a template and haven't changed it.

**Screenshots and Repro Model**
![Screenshot 2023-06-08 214025](https://github.com/gkjohnson/three-gpu-pathtracer/assets/64857616/19c62aba-b355-49e1-86a8-0b78008a9346)

**Platform:**

 - Device: Mobile
 - OS: Android 13
 - GPU: Qualcomm Snapdragon 778G 5G (SM7325)
 - Browser: Chrome 114.0.5735.57, FireFox 113.2.0
 - Three.js version: 0.153.0
 - Library version: 0.0.15

gkjohnson:
See https://github.com/gkjohnson/three-gpu-pathtracer/issues/441#issuecomment-1610640494

gkjohnson:
Fixed in #483 by upgrading three-mesh-bvh version

PR #483, opened 2024-01-13
MacOS crash exploration

gkjohnson:
Related to #439 
Related to #442 
Related to #450 
Related to #463 

**FUTURE**
- Clean up GLSL snippets - functions and structs
- make uniforms obvious so we can use them within functions more easily
- OR for bvh / envHdrInfo expand the functions using macros


gkjohnson:
The call to `sobol3( 6 )` seems to be causing the issue. It's used when generating a sampled light point direction and causes a failure when the value is assigned to `ray.direction`.

PR #527, opened 2024-02-10
HoverInc Stratified samples

### LINKED ISSUE #449: Investigate HoverInc sample stratification
https://github.com/hoverinc/ray-tracing-renderer

gkjohnson:
Fix #449 

**TODO**
- Comparison images
- Reduce swirling effect
  - This is a artifact of each shape sampling algorithm and the correlation between X and Y.
  - Possibly using a 4 color blue noise seed this can be helped
  - Possibly reshuffling the strata more intelligently would help
  - \* Use a 3d texture with X: full strata for one sample, Y: new strata per id, Z: new strata per bounce
  - \* Strata doesn't need to be regenerated until wrapping around?

**NOTES**
- Doesn't work with "stableNoise" (requires predictable randomness) (punt)


gkjohnson:
The stratification approach seems to turn bright spots into swirls

image

gkjohnson:
| | Before | After |
|---|---|---|
| 1 sample | image | image |
| 10 samples | image | image |
| 50 samples | image | image |

gkjohnson:
NEW 10 samples
| Before | After |
|---|---|
| image | image |

Issue #558, opened 2024-03-12
Support BumpMap like three.js


dongho-shin:
Always thanks you for maintain amazing library

I tried to add bumpMap support like three.js
but I got half success(bumpMap Rendered in only two faces and other faces are black) of it do you guys have any ideas?

<img width="508" alt="Untitled (28)" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/108255990/e0e1bd4e-056f-49e4-b467-589f352d2879">

I use this bumpTexture

![bump-mapping](https://github.com/gkjohnson/three-gpu-pathtracer/assets/108255990/bc886a76-ba36-44a6-aa59-552415ff037b)

PhysicalPathTracingMaterial's vertex shader i add vViewPosition 

<img width="453" alt="image" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/108255990/dfdbb936-56ec-4b8f-8364-ffebcbd33cf7">

<img width="623" alt="image" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/108255990/18ddae0a-41c2-4dc3-bcec-e43b7e659ada">

```glsl
// https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/bumpmap_pars_fragment.glsl.js
	vec3 perturbNormalArb( vec3 surf_pos, vec3 surf_norm, vec2 dHdxy, float faceDirection ) {
	
		// normalize is done to ensure that the bump map looks the same regardless of the texture's scale
		vec3 vSigmaX = normalize( dFdx( surf_pos.xyz ) );
		vec3 vSigmaY = normalize( dFdy( surf_pos.xyz ) );
		vec3 vN = surf_norm; // normalized
		
		vec3 R1 = cross( vSigmaY, vN );
		vec3 R2 = cross( vN, vSigmaX );
		
		float fDet = dot( vSigmaX, R1 ) * faceDirection; //fDet is 0
		vec3 vGrad = sign( fDet ) * ( dHdxy.x * R1 + dHdxy.y * R2 );
		vec3 result = normalize( abs( fDet ) * surf_norm - vGrad );
		
		return result;

	}
```
``` glsl
                // getSurfaceRecord.glsl.js at line 164
		vec3 baseNormal = normal;
		if ( material.normalMap != - 1 ) {

			vec4 tangentSample = textureSampleBarycoord(
				attributesArray,
				ATTR_TANGENT,
[...]

PR #559, opened 2024-03-12
Add support for black and white Bump Maps to compute normals PR

dongho-shin:
related : https://github.com/gkjohnson/three-gpu-pathtracer/issues/558

also this pr has a problem(I write a details in related issue #558 )

bumpMap applied not fully success(especially plane)

image

Untitled (28)


gkjohnson:
You'll also want to make sure that tangents are generated when bump maps are present instead of just when normal maps are:

https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/src/core/DynamicPathTracingSceneGenerator.js#L83-L84

dongho-shin:
@gkjohnson i miss that... I'll check tangents I use only bumpMap at mesh and I can't get tangent attributes normally at shader

dongho-shin:
I'm pretty new in glsl or graphics so i don't have a idea how to do dFdx and dFdy manually can i get a hint?

This is as far as I understand it at this point. surf_pos (- vViewPosition) that input of perturbNormalArb(in my pr and three.js bumpMap function) from vertex shader so its value has no effected pathTracing sampling but dFdx, dFdy (surf_pos) has effected right?
because dFdx dFdy get near gl_Fragcoord  and get change of x or y automatically

then should I get a near vertex position in screen space and add by uniform and use it in glsl?

gkjohnson:
> I'm pretty new in glsl or graphics so i don't have a idea how to do dFdx and dFdy manually can i get a hint?

Of course! Feel free to ask questions. I'm happy to have some help adding additional features to the project.

Regarding the dFdX and dFdY functions - they compute the apparent change in a variable from pixel to pixel _in screen space_. It does this by checking sibling pixels and computing the difference between the variable value. This great for traditional rasterization but of course because all of these path tracing operations occur as a light ray bounces around a scene it's not appropriate to compute the change in texture value in screen space.

Likewise the idea of using a "view position" or "direction" (which is based on the view ray from the original camera position) doesn't make sense since the rays can hit a surface from any direction after bouncing around. This is why the `perturbNormalArb` cannot be used.

So on to how to compute the derivative (dFdX, dFdY) of the texture without using the built in functions - the derivative is the change in a variable over time along some dimension. So, instead, to manually compute the change along the X axis, for example, you can sample the value at the point hit and one pixel to the left and right and compute the derivative from those values.

Since a bump map is basically a height map the derivative of a pixel represents the slope of the surface represented in the bump map which you can use to compute the normal.

Hopefully this was helpful - let me know if you need me to elaborate on anything.

dongho-shin:
then approach would simple with conversion of bump to normal... I'm going to find a solution of it

Issue #498, opened 2024-01-18
Make an easy-to-use renderer modeled after WebGLRenderer


gkjohnson:
Use async compilation, handle ios float texture issues, bvh generator, automatic fallback rendering, flag for disabling render to canvas, ability to save etc

Class name: WebGLPathTracer

gkjohnson:
```js
const pathTracer = new WebGLPathTracer( renderer );
pathTracer.tiles.set( 2, 2 );
pathTracer.renderScale = 0.5;
pathTracer.texturesSize.set( 1024 );
pathTracer.filterGlossyFactor = 0.5;
pathTracer.setScene( scene, camera, {
  updateMaterials: true,
  updateTextures: true,
  updateScene: true,
} );

document.appendChild( pathTracer.domElement );

// or

pathTracer.renderToCanvas = false;
pathTracer.target;


```

gkjohnson:
**Next PRs**
- Update materials, textures, lights, material ids even when the bvh is refit.
- Add ability to generate meshes via worker.
- Adjust dynamic generator so that the geometry and bvh expands with a change in geometry count and size.
- Update docs with usage snippet
- Add fade in effect
- Add low-res fallback effect

PR #563, opened 2024-03-15
Add WebGLPathTracer convenience class

gkjohnson:
Related to #498 

**Next PRs**
- Update materials, textures, lights, material ids even when the bvh is refit.
- Add ability to generate meshes via worker.
- Adjust dynamic generator so that the geometry and bvh expands with a change in geometry count and size.
- Update docs with usage snippet
- Add fade in effect
- Add low-res fallback effect

### DIFF
--- a/example/primitives.js
+++ b/example/primitives.js
@@ -56,7 +56,7 @@ camera.lookAt( 0, 0, 0 );
 const renderer = new WebGLPathTracer();
 document.body.appendChild( renderer.domElement );
 renderer.toneMapping = ACESFilmicToneMapping;
-renderer.tiles.set( 3 );
+renderer.tiles.set( 3, 3 );
 renderer.updateScene( camera, scene );
 
 onResize();
--- a/example/primitives.js
+++ b/example/primitives.js
@@ -1,11 +1,5 @@
 import * as THREE from 'three';
-import { FullScreenQuad } from 'three/examples/jsm/postprocessing/Pass.js';
-import {
-	PathTracingSceneGenerator,
-	PathTracingRenderer,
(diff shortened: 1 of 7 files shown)


gkjohnson:
Looks like bounces are set to a lower value?

**Before**
image

**After**
image

gkjohnson:
The darkening is due to the "filterGlossyFactor" - should look into in a future PR.

gkjohnson:
cc @donmccurdy - this PR adds a more simple way to access and render a path traced scene that mirrors the WebGLRenderer API design in some ways. I think this is something you'd wanted previously? Here's what it looks like to path trace a scene now:

```js
const scene = new Scene();

// initialize scene ...

const texture = new GradientEquirectTexture();
texture.bottomColor.set( 0xffffff );
texture.bottomColor.set( 0x666666 );
texture.update();
scene.environment = texture;
scene.background = texture;

const renderer = new WebGLPathTracer();
document.body.appendChild( renderer.domElement );
renderer.toneMapping = ACESFilmicToneMapping;
renderer.tiles.set( 3 );
renderer.setPixelRatio( window.devicePixelRatio );
renderer.setSize( window.innerWidth, window.innerHeight );
renderer.updateScene( camera, scene );

animate();

function animate() {

	// if the camera position changes call "ptRenderer.reset()"
	requestAnimationFrame( animate );

	// update the camera and render one sample
	camera.updateMatrixWorld();

	// rasterizes the main scene if there are not enough path traced samples and renders the path traced scene
	// tiles on top. 
	renderer.renderSample();

}
```

There are some settings to disable rendering to canvas so you can just work with the Path Traced render target, as well. Let me know what you think.

PR #628, opened 2024-04-18
Allow disabling jitter

Codezigineer:
Jitter is what makes raytracers grainy, by shooting pixels at slightly different directions than normal. This allows disabling it, for tasks that don't need it.

### DIFF
--- a/src/core/PathTracingRenderer.js
+++ b/src/core/PathTracingRenderer.js
@@ -195,6 +195,18 @@ export class PathTracingRenderer {
 
 	}
 
+	set jitterEnabled( v ) {
+
+		this._ptMaterial.jitterEnabled = v;
+
+	};
+
+	get jitterEnabled() {
+
+		return this._ptMaterial.jitterEnabled;
+
+	};
+
 	constructor( renderer ) {
 
 		this.camera = null;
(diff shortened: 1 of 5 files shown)


gkjohnson:
Hello! Thanks for the contribution.

> Jitter is what makes raytracers grainy, by shooting pixels at slightly different directions than normal. This allows disabling it, for tasks that don't need it.

Can you provide some screenshots of cases where this disabling jitter helps with the final image and how? Jitter is effectively performing antialiasing - it can contribute to the graininess but every ray bounce is scattered in a random direction regardless of this perpixel jitter which is primarily where the graininess should be coming from.

Codezigineer:
https://www.shadertoy.com/view/4ljGRd is an example.

gkjohnson:
> https://www.shadertoy.com/view/4ljGRd is an example. Some tasks don't need jitter.

This is a very different kind of ray tracer, though. There is no randomness used anywhere and all diffuse lighting is calculated analytically rather than stochastically with random rays. Removing the pixel jitter from this project shouldn't result in the same kind of result.

What's the motivation for this PR? Are you using this for a project and need this change? Do you have screenshots showing what a render looks like before and after so I can understand how disabling the jitter affects the image?

Codezigineer:
I made this for a project I am working on, that I'm specifically using raytracing because RectArealights don't have shadows. Yes that is the only reason.

gkjohnson:
Got it - without a demonstration showing the improvements that this PR will bring to the project I won't be able to merge it. I don't think this will address any of the graininess issues in the image generation due to random sampling.

gkjohnson:
Closing this until it can be shown that there's an improvement to image quality.

PR #630, opened 2024-04-20
fix: update bufferToHash import with .js

rndmguy-91:


gkjohnson:
Ha, thank you very much - bit by VSCode's auto import system again.

PR #632, opened 2024-05-02
Fix to only render objects with visible true

KihwanChoi12:
It makes sense that only objects with visible true are rendered.

The old code used traverseVisible when generating the mesh list, but it was missing in the newer code.


gkjohnson:
Great, thank you for catching this! I'll make a new release soon.

Issue #639, opened 2024-05-22
Is there any way to render overlay on the pathTraced Scene?


dongho-shin:
Hi Previous Version it can draw overlay over the pathTraced Scene like code below
(ex. area light example that use TransformControls in past)

```
renderer.autoClear = false;
quad.render(renderer);
renderer.render(overlayScene, camera);
renderer.autoClear = true;
```

but somehow it can't render overlay 

rasterized Scene it can render
<img width="787" alt="image" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/108255990/6b0cc74c-cc35-4c3a-bd46-564dac7f6f6d">

but when it turns pathTraced Scene overlay cube disappear
<img width="1019" alt="image" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/108255990/7a53deb6-d0f0-463c-b364-a5d7ba45de4d">

is there something i missed?

thanks to read this issue

gkjohnson:
Currently if you'd like to render after the fact with depth you'll have to render to depth first and then draw the opaque objects on top of that. The pathtraced view has never provided a depth buffer state after rendering.

> (ex. area light example that use TransformControls in past)

These controls never interacted with the depth buffer as far as I remember. Can you provide a small working example of what you were able to do in previous examples of the project in a previous verison?

dongho-shin:
I want do same in the 0.0.20 version of three-gpu-pathtracer  /areaLight example TransformControls
Gizmos drawn on the view 
i think it's not related depth buffer I know it's draw over rendered view without clear
<img width="1251" alt="image" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/108255990/39e0bea7-eb65-4f3e-8675-3090da6c5de4">

i open pr for example sharing 
https://github.com/gkjohnson/three-gpu-pathtracer/pull/640

gkjohnson:
The transform controls don't check the depth buffer as far as I remember so the depth state wouldn't have ever really mattered while your box is test depth. But if you clear depth in your PR the objects will render just fine:

```js
renderer.autoClear = false;
renderer.clearDepth();
renderer.render(testMesh, activeCamera);
renderer.autoClear = true;
```

<img width="350" alt="image" src="https://github.com/gkjohnson/three-gpu-pathtracer/assets/734200/b3f41221-43f2-4257-b165-838e299f8c36">

[...]

PR #640, opened 2024-05-23
Example PR for Example overlay question 

dongho-shin:
ref. https://github.com/gkjohnson/three-gpu-pathtracer/issues/639


dongho-shin:
I'll make small example and then request again

PR #644, opened 2024-05-28
fix StaticGeometryGenerator generate result wrong when specific condition

dongho-shin:
Hi I found a bug so i made a pr for reproducing bug sample
check the sample(index.js) in my pr

# how to reproduce a bug

a screenshot after initialization
image

## 1. check the last mesh name in the console
image
<br/>

## 2. make visible false mesh at the last of array(in this case mesh4)
image
<br/>
then you can see artifact at mesh4's position

### Description
at this time console here
image

At first initialization force update true in all meshes because there's no cached

after testMesh4 visible false,
meshes from this._getMeshes make an array that only visible in the scene
then for loop check difference but there's no difference even testMesh4 is disappear(because for loop just check current meshes) so staticGenerator return NO_CHANGE result and then it causes an artifact

## How to fix it

```
		let forceUpdate = false;
		if ( meshes.length !== previousMergeInfo.length ) {

			forceUpdate = true;

		}

		console.log('meshes', meshes.map(m => m.name))
		for ( let i = 0, l = meshes.length; i < l; i ++ ) {
```

if you guys agree my solution i'll clean up this pr that only remain solution

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -8,6 +8,7 @@ import {
 	Mesh,
 	MeshStandardMaterial,
 	PlaneGeometry,
+	BoxGeometry,
 	MeshPhysicalMaterial,
 	Scene,
 	PerspectiveCamera,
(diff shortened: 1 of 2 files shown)


gkjohnson:
Excellent find! Thank you very much. Once the PR is updated to just modify StaticGeometryGenerator we can merge!

dongho-shin:
it's done I clean up this pr

gkjohnson:
Great, thank you!

PR #645, opened 2024-05-30
feat: add shaped area light helper

JaimeTorrealba:
I manage to add the new visibleSurface value, but I couldn't find a way to render the texture as helper. So I believe I need some help here. I should have to modify the src/shader/sampling/light_sampling_functions.glsl.js in order to access to the src/materials/pathtracing/PhysicalPathTracingMaterial.js

(now is just adding it to the alpha channel hideous just to prove that it reach the value, of course also tried to move in other locations like using the `state.firstRay`, but didn't find the right one to render a texture as helper)

Area Light Path Tracing

btw, I'll be off a couple of days, so if this is not priority, with your help we can solve it :)

### DIFF
--- a/src/materials/pathtracing/PhysicalPathTracingMaterial.js
+++ b/src/materials/pathtracing/PhysicalPathTracingMaterial.js
@@ -361,6 +361,7 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 									// NOTE: Only area lights are supported for forward sampling and can be hit
 									float misWeight = misHeuristic( scatterRec.pdf, lightRec.pdf / lightsDenom );
 									gl_FragColor.rgb += lightRec.emission * state.throughputColor * misWeight;
+									gl_FragColor.a = lightRec.visibleSurface; // This is not working as expected
 
 									#else
 
--- a/src/shader/structs/lights_struct.glsl.js
+++ b/src/shader/structs/lights_struct.glsl.js
@@ -17,6 +17,7 @@ export const lights_struct = /* glsl */`
 
 		vec3 position;
 		int type;
+		float visibleSurface;
 
 		vec3 color;
 		float intensity;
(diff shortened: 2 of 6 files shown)


src/materials/pathtracing/PhysicalPathTracingMaterial.js
```diff
@@ -361,6 +361,7 @@ export class PhysicalPathTracingMaterial extends MaterialBase {
 									// NOTE: Only area lights are supported for forward sampling and can be hit
 									float misWeight = misHeuristic( scatterRec.pdf, lightRec.pdf / lightsDenom );
 									gl_FragColor.rgb += lightRec.emission * state.throughputColor * misWeight;
+									gl_FragColor.a = lightRec.visibleSurface; // This is not working as expected
```
gkjohnson, line 364: This section of code only runs when the current ray is _not_ a camera ray (a ray representing light that goes directly into the camera - not reflected) - see the condition above:

```c
if ( ! state.firstRay && ! state.transmissiveRay ) {
```

Conversely we only want to display the light surface and exit if the current ray _is_ a camera ray. So we'll need to add a new loop to find the light surface in an else statement:

```c
if ( ! state.firstRay && ! state.transmissiveRay ) {

    // accumulate light from direct lights hits for reflected rays

} else {

    // find the closest light with a visible surface, exit path tracing

}
```

Hopefully that makes sense?

src/shader/structs/lights_struct.glsl.js
```diff
@@ -44,12 +45,14 @@ export const lights_struct = /* glsl */`
...
 		vec4 s2 = texelFetch1D( tex, i + 2u );
 		vec4 s3 = texelFetch1D( tex, i + 3u );
+		vec4 s4 = texelFetch1D( tex, i + 4u );
 
 		Light l;
 		l.position = s0.rgb;
 		l.type = int( round( s0.a ) );
 
 		l.color = s1.rgb;
+		l.visibleSurface = s4.r;
```
gkjohnson, line 55: The red component of sample 4 is for other data (see [here](https://github.com/gkjohnson/three-gpu-pathtracer/blob/f427d9dae2265cbf1763be1636caf8a1ad9057ef/src/uniforms/LightsInfoUniformStruct.js#L158-L169), confusingly commented as sample 5 for s4, sorry 😅) so we if we read this sample for every light then we'll have an incorrect value for some lights. Alternatively we could pick a pixel component that isn't used by any light (like sample 6 alpha) which would may be helpful for adding support to other lights later.

gkjohnson (commented):
Great thank you for the PR! Just added a few comments to hopefully clarify things - let me know if it's not clear and I'll try again!

JaimeTorrealba:
I'll contribute this pkg differently.

I'm working on a version for Tres library, soon available.

Thanks for the guide

PR #652, opened 2024-06-07
Fix to always set environment intensity

KihwanChoi12:
Reproduction Paths
1. after the environment of Pathtracer is set once, change the environment of scene to null.
2. when pathtracer is executed again with the environment null, it does not set the environment intensity because it is the same as the previous environment, and the old environment of the scene is rendered.

Solution
Always reset environment intensity


gkjohnson:
Sorry for the delay! Nice catch - thank you very much!

PR #655, opened 2024-06-26
Add alphaMapTransform

dongho-shin:
I noticed that alphaMapTransform was missing, I added it and cleaned up the code around MATERIAL_PIXELS.


gkjohnson:
Excellent, thank you very much!

PR #660, opened 2024-07-16
materialDatabase updates

AntonPalmqvist:
### Switched tonemapper from ACES to AgX
Since the reference images use the AgX tonemapper (with the Base Contrast look), this update makes the render match the reference images better.

Before (ACES)
Screenshot 2024-07-15 at 20 30 11
After (AgX)
Screenshot 2024-07-15 at 20 28 46

### Added iridescence for materials that support it
Materials such as Pearl and Soap Bubble now have values for iridescence on the API side and this change enables it whenever a **thinFilmThickness** value is found on a material.
Screenshot 2024-07-14 at 20 03 00

### Changed reference image source
The API provides a **reference** field with a URL to a reference image (same images as before) and is made for just this kind of application. In v2 of the API there will also be different sizes and file types available.

### DIFF
--- a/example/materialDatabase.js
+++ b/example/materialDatabase.js
@@ -183,7 +183,7 @@ function applyMaterialInfo( info, material ) {
 	material.thickness = 1.0;
 	material.iridescence = 0.0;
 	material.iridescenceIOR = 1.0;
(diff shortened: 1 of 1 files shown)


gkjohnson:
Amazing, thank you!

PR #663, opened 2024-07-24
Add OIDN Denoiser.

DennisSmolek:
This PR Adds the OIDN Denoiser ([Denoiser](https://github.com/DennisSmolek/Denoiser)) into the core three-gpu-pathtracer.

It works, and looks decent but there are clear issues to resolve.

### API Added
- `maxSamples` Pathtracer now stops when reaching this number. Also is the number the system denoises at
- `enableDenoiser` If we are using the denoiser (NOTE, the denosier actually initializes regardless of this for webGL compat)
- `weightsUrl` Lets you bypass the jsDelivr URL for your own location of the TZAs. Also how you could pass your own weights in general
- `denoiserQuality` Fast by default, Balanced is better. "high" is really heavy and not supported in the Denoiser yet
- `useAux` Whether or not you want to use just the pathtraced input (fast, but bad) or with the Aux
- `renderAux` If you want to view the aux inputs set this to `albedo` or `normal` `null` gets you the denoised output
- `cleanAux` If you are 100% the aux inputs are clean, ANY noise will be projected into the output (NOTE, upstream fix may be needed)
-  `externalAux` If the user is supplying their own Aux Textures. disables automatic generation
- `denoiserDebugging` Sets the denoiser to log more things including an output report.

#### State API Properties (read-only)
- `isDenoising` If the denoiser is running
- `isDenoised` if the current output is the denoised output
#### Methods

- `generateAux()` runs the albedo/normals pass on the current scene. returns `{ albedo: THREE.Texture, normal: THREE.Texture }`
- `setAuxTextures(albedoTexture, normalTexture)` if the user wants to set the aux themselves, like in a pipeline or has a deferred setup. Deactiviates internal automatic creation (`this.externalAux = true`)
- `denoiseSample(inputTexture, albedoTexture, normalTexture)` loads automatically or can be overridden letting you send to the denoiser directly to the `DenoiserManager` Not sure about exposing this, called internally by `renderSample`

### Changes to Core
- Not much was done to the core itself. The main thing is the addition of another fullScreenQuad pass that blends the pathtracer output with the denoised (or aux if you're debugging) outputs. 
- The second significant change is the addition of `maxSamples` and it's stopping the pathtracer when reached.
- There were other changes to better support the denoiser, but nothing breaking.

One change was added to the `BlendingMaterial` to allow a conversion to happen to texture2. While we resolve the colorspace issues it may be useful.

### Additions
- AlbedoNormalPass: Generates Albedo & Normal textures based on the current scene
- DenoiserManager: Holds all things related to the denoiser and interfaces with the `Denoiser` class directly

### Flow With Denoising
Assuming Denoiser is enabled and all options default/set
1. Pathtracer runs until `maxSamples` is hit
2. Denoiser process started. Block future `denoiseSample` calls until finished
3. Albedo and Normal Textures Generated (only when needed by the denoiser, not every frame)
4. RawPathtracerTexture (`pathtracer.target.texture`), AlbedoTexture, and NormalTexture sent to the DenoiserManager.
5. If the renderer size has changed, regenerate renderTargets and set Denoiser sizes.
6. Render the RawPathtracer texture to a Quad with same tonemapping as default PT so resulting texture matches and within [0, 1]
7. Extract raw `WebGLTextures` from internals of input textures (`THREE.Texture`s)
8. Set those Inputs on the denoiser
9. Execute the denoiser, return `denoisedWebGLTexture`
10. If not already created, create and load a ThreeJS texture that is properly initialized and setup within threejs internals `outputTexture` ( one of the many steps to get WebGL/Three to play nice)
11. Merge the `denoisedWebGLTexture` with the outputTexture 
12. `denoisedTexture` is held in the class ready to be rendered
13. denoiser marked finished, results say it is now denoised
14. In the next `renderSample` call now that `isDenoised` is true, render using `DenoiserManager.renderOutput()`
15. Quad material now set to `BlendingMaterial`
16. Blend between previous PT Output (results of step 6) and the `denoisedTexture` (Note: if RenderAux is passed, will render the aux texture provided so you can see either the albedo or normal textures visually. Not sure the normals render how you'd expect, as they should be outputting in [-1, 1])

If reset is called it all drops back to normal, hides the outputs etc and starts over.

### Known Issues
**1. ColorSpaces/tonemapping/conversion:** Something is clearly not setup right through the pipeline regarding colors. The denoiser DOES NOT CARE what colorspace you input. Whatever you input, it will output. Color and Albedo inputs should be in the same colorspace, and normals are expected to be in linear. The output looks correct but dark. converting makes it look flat. 
This will take tweaking and someone smarter than me with regards to color to follow the renderTargets (all setup with `THREE.SRGBColorSpace` and what we should convert/adjust

**2. Normal Generation:** I have a script setup to use worldspace or viewspace normals and output to [-1, 1]. I don't know if colorSpace/RT's might be effecting this (I don't think it matters). Also, I have added code to accept normalMaps on meshes so they can be used instead of the raw surface. I tried converting the NormalMaps to worldspace with tragic results. For the moment If an object has a map I use it.

Something very important to point out about normals and the denoiser. The denoiser does not actually use them for normal values or any kind of light calculations whatsoever. They can be in world/local space. The normal maps just help define edges and breaks in materials.

**3. Albedo Generation:** [In The OIDN Docs](https://github.com/RenderKit/oidn?tab=readme-ov-file#albedos) they go into detail about the albedos. In general, most matte surfaces can use just the basic color output or textures. This isn't true for highly reflective surfaces (like the floor). Here albedo wants something different, you can read the docs and it says something to the effect that the reflected surface should have the albedo of whatever they are reflecting, or a full 1 value. It gets even weirder for reflections. 

**4.  Edges with floor and background.** Looking at original OIDN examples they do not expect black backgrounds or hard edges where soft edges blend with the original background. The normals/albedo read those as hard crisp edges or flat planes. So the gradient background gets flattened into a weird flat gradient and the floor has sharp edges. So with threejs backgrounds or envMaps we should generate something for both of these passes and include any floor/horizons.

The only one of these issues I see as a blocker is the colorspaces. Generating albedos and normals etc we can work on for a while and it would be fine to release improvements as updates. But gotta get closer on the color outputs IMO.

### Other Changes
I updated the plain example (index) to include denoiser props and some other slight adjustments. I added a rawOutput canvas to the html which is very easy to setup when debugging to see exactly what the denoiser is outputting and confirm your inputs/outputs.

### Notes
I tried commenting a lot to explain what is happening, and I realize my commits are terribly labeled. I was focusing on getting this added as the main thing and not making a lot of gradual changes.


src/core/utils/DenoiserManager.js
```diff
@@ -0,0 +1,208 @@
...
+	// render the blended output
+	renderOutput( bypassTexture ) {
+
+		if ( ! this.pathtracedTexture || ! this.denoisedTexture ) return;
+
+		//const currentTarget = this.renderer.getRenderTarget();
+		this.quad.material = this.blendMaterial;
+		this.blendMaterial.target1 = this.pathtracedTexture;
+		this.blendMaterial.target2 = this.denoisedTexture;
+		this.blendMaterial.t2conversion = false;
```
KihwanChoi12, line 113: When testing against this PR, I found that the results were coming out too dark. Due to the colorspace issue, it looks like you are generating t2conversion uniform. When outputting denoised results, it looks fine if you set t2 conversion to true. Please confirm the fix.

DennisSmolek: I too saw the brighter output, but note that the colors are much flatter.
To test, set the samples really high, like 100. Use that as your Ground Truth. You'll see the results are somewhere between the current output and the converted output.

the t2conversion was my original solution and may be the correct way to go, however, the conversion earlier in the pipeline may yield better results.

Should we try converting the raw pathtracer output before it goes into the denoiser? If we do that we must also convert the albedo pass.

I'm willing to try whichever ways people think best or even add debug options in them so we can test. 

I figured it best to bring out and let others try and see what everyone thinks the best approach is.

KihwanChoi12: When I was testing OIDN, all inputs had to be set to SRGB colorspace (pathtraced, albedo, normal), and when displaying the denoised result, there was no color distortion when outputting the denoised result without changing the colorspace. Let's run a related test.

DennisSmolek: [The OIDN docs](https://github.com/RenderKit/oidn?tab=readme-ov-file#rt) say:
>  srgb | false | the main input image is encoded with the sRGB (or 2.2 gamma) curve (LDR only) or is linear; the output will be encoded with the same curve

Which says they are linear by default and can be SRGB.
[Later](https://github.com/RenderKit/oidn?tab=readme-ov-file#rt) they say the model is trained in srgb space so I agree that we should try to keep input as SRGB.

In the render targets I set the colorspace to `THREE.SRGBColorSpace`. Shouldn't that make the renderer output to that colorspace?

This is confusing because the `t2Conversion` is `LinearToSRGB()`

src/core/utils/AlbedoNormalPass.js
```diff
@@ -0,0 +1,224 @@
...
+		// biome-ignore lint/complexity/noForEach: <explanation>
+		this.albedoMaterials.forEach( ( material ) => {
+
+			if ( material instanceof THREE.ShaderMaterial ) {
+
+				material.uniforms.useWorldSpaceNormals.value = value;
+
+			}
+
+		} );
```
gkjohnson, line 204: Is there a reason to not just use one kind of normals? I'd like to avoid a toggle if possible. We should just pick the best one of there is one.

yarn.lock
gkjohnson, line 1: Yarn isn't being used in this project - lets remove this file.

src/core/WebGLPathTracer.js
```diff
@@ -98,6 +99,75 @@ export class WebGLPathTracer {
...
+
+		return this._denoiserManager.denoiser.tzaUrl;
+
+	}
+
+	set weightsUrl( url ) {
+
+		this._denoiserManager.denoiser.weightsUrl = url;
+
+	}
```
gkjohnson, line 121: Instead of taking the URLs on the WebGLPathTracer object I'd prefer passing something like an "OIDNDenoiser" object to a function to assign it - similar to three.js [GLTFLoader.setDracoLoader](https://github.com/mrdoob/three.js/blob/f5eaae88a09fce661a814d2bd3526f1690302bb6/examples/webgl_loader_gltf_avif.html#L62) or the [setBVHWorker](https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/example/index.js#L137) functions. Then any options can be set on that denoiser instance and it would let use swap in other denoisers in the future:

```js
const pathTracer = new WebGLPathTracer( renderer );
console.log( pathTracer.denoiser ); // null
console.log( pathTracer.enableDesnoiser ); // false

const denoiser = new OIDNDenoiser( options );
denoiser.setTZAPath( ... ); // see DracoLoader's "setDecoderPath" function
denoiser.debugging = true;
denoiser.setAux( ... );
// any other denoiser options etc

pathTracer.setDenoiser( denoiser );

// adjusting options later
pathTracer.denoiser.debugging = false; // etc
```

src/core/WebGLPathTracer.js
```diff
@@ -154,7 +241,7 @@ export class WebGLPathTracer {
 		};
 
 		// initialize the scene so it doesn't fail
-		this.setScene( new Scene(), new PerspectiveCamera() );
+		this.setScene( new Scene(), new PerspectiveCamera(), { blockDenoise: true } );
```
gkjohnson, line 244: I don't think this needs to be an option - let's just add a new flag on WebGLPathTracer called `pathTracer.isGenerating` when a scene is updating that's always set.

src/core/utils/AlbedoNormalPass.js
```diff
@@ -0,0 +1,224 @@
...
+                        vec4 diffuseColor = texture(diffuseMap, vUv);
+                        albedo = diffuseColor.rgb * color;
+                    } else {
+                        albedo = color;
+                    }
+                    gAlbedo = vec4(albedo, 1.0);
+
+                    // Normal
+                    vec3 normal;
+                    if (useNormalMap) {
```
gkjohnson, line 86: We should be able to reuse a lot of the three.js build in shader #includes to generate all these values without so much code.

src/materials/fullscreen/BlendMaterial.js
```diff
@@ -14,6 +14,7 @@ export class BlendMaterial extends MaterialBase {
 				target1: { value: null },
 				target2: { value: null },
 				opacity: { value: 1.0 },
+				t2conversion: { value: false }
```
gkjohnson, line 17: What does "t2conversion" mean?

src/core/utils/DenoiserManager.js
```diff
@@ -0,0 +1,208 @@
...
+		see https://github.com/DennisSmolek/Denoiser/issues/21
+		const denoisedWebGLTexture = await this.denoiser.execute( colorWebGLTexture, albedoWebGLTexture, normalWebGLTexture );
+		*/
+
+		// Set input Textures with await
+		await this.denoiser.setInputTexture( 'color', colorWebGLTexture );
+		if ( albedoTexture ) await this.denoiser.setInputTexture( 'albedo', albedoWebGLTexture );
+		if ( normalTexture ) await this.denoiser.setInputTexture( 'normal', normalWebGLTexture );
+		// Run the denoiser
+		const denoisedWebGLTexture = await this.denoiser.execute();
```
gkjohnson, line 91: I see some logs about tiling happening - are the tiles being rendered over multiple frames? It would be nice if the page didn't lock up quite so much when denoising is happening. The tiles could also be used to display a completion percentage.

gkjohnson (commented):
I left a smattering of comments for some of the things that stuck out to me. Once some of the architecture is nailed down and it's more clear to me exactly what's happening I can help address some of the color space and tone mapping issues.

src/core/utils/DenoiserManager.js
gkjohnson: I noticed the colors being significantly darker, as well - it does look like a color space issue. I can help with that later on.

| raw | denoised |
|---|---|
| image | image |

src/materials/fullscreen/BlendMaterial.js
DennisSmolek: I needed a way to apply colorspace conversion to one of the inputs. In this case "target2". "t2conversion" is just short for "target2 color conversion"

src/core/utils/DenoiserManager.js
DennisSmolek: The tile generation we could probably wire up to wait for a frame before going to the next tile (allowing less lockup) but I wonder if this would slow things down on higher speed machines.[Tensorflow has a method for this](https://js.tensorflow.org/api/latest/#nextFrame)

Tile reconstruction (which is the second half) should happen as fast as possible, I'm not sure we want to divide that up.

I'll log down both strategies to try.

One thing to remember, With the WebGL backend the first execution is SIGNIFICANTLY longer than the others. Tensorflow waits to compile shaders until you actually execute.It's suggested we run a few images through to "prime" the model but this would cause it to be slow at load time which I was avoiding.

I could see a "ready()" callback that fires after a warm up image is passed.

package.json
```diff
@@ -35,20 +35,32 @@
...
+    "browserify-zlib": "^0.2.0",
     "canvas-capture": "^2.0.5",
+    "crypto-browserify": "^3.12.0",
     "eslint": "^7.32.0",
     "eslint-config-mdcs": "^5.0.0",
+    "events": "^3.3.0",
+    "https-browserify": "^1.0.0",
     "node-fetch": "^3.2.9",
     "parcel": "^2.12.0",
+    "path-browserify": "^1.0.1",
```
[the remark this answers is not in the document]
DennisSmolek: They are probably mistakes from when I accidently included a library. I will confirm and remove them

example/index.js
```diff
@@ -108,6 +108,16 @@ const params = {
...
+	// Denoiser
+	maxSamples: 6,
+	limitlessSamples: false,
+	enableDenoiser: false,
+	useAux: true,
+	cleanAux: true,
+	renderAux: '',
+	denoiserQuality: 'fast',
+	denoiserDebugging: false,
+
```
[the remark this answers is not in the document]
DennisSmolek: For sure, I included everything for debugging purposes. When things are working well I'll remove them. `renderAux` is usefull for example to see the output of the auxTextures

example/index.js
```diff
@@ -269,11 +279,70 @@ function onParamsChange() {
...
+		// force the renderscale to be 1 when using the denoiser
+		pathTracer.renderScale = 1;
+		params.renderScale = 1;
+		// set and disable the renderScale gui
+		if ( renderScaleController ) {
+
+			renderScaleController.disable();
+			renderScaleController.setValue( 1 );
+
+		}
```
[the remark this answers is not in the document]
DennisSmolek: When the pixeled are scaled OIDN rejects them. Theres documentation on it somewhere but it considers them as surface microstructures and treats it as intentional. Each pixel has to be 1 pixel in the WebGLTexture. 

If we can somehow get access to the base (unscaled) texture we can denoise that and then scale the result. Otherwise using renderScale will break the denoiser.

src/core/WebGLPathTracer.js
```diff
@@ -479,11 +579,51 @@ export class WebGLPathTracer {
...
 		}
 
+	}
+	setAuxTextures( albedoTexture, normalTexture ) {
+
+		this.externalAux = true;
+		this._auxTextures.albedo = albedoTexture;
+		this._auxTextures.normal = normalTexture;
+
+	}
```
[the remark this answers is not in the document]
DennisSmolek: Totally up to you. This allows the pathtracer to exist in a pipeline or post-processing stage where albedo or normals may already have been generated (like a deferred renderer) and reuse those textures. I'll remove it.

src/core/utils/AlbedoNormalPass.js
```diff
@@ -0,0 +1,224 @@
...
+            `,
+			glslVersion: THREE.GLSL3
+		} );
+
+	}
+
+	async render( renderer, scene, camera ) {
+
+		const target = this.renderTarget;
+		target.setSize( renderer.domElement.width, renderer.domElement.height );
```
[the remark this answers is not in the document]
DennisSmolek: Nice! I didn't know that was an option. I will adjust and this also would clear up the renderScale blocking.
Where is the scaling re-applied to output? The denoiser would return a lower scale texture than the final render.

src/core/utils/AlbedoNormalPass.js
DennisSmolek: Agreed, OIDN said it supports both and I didn't want to force the decision and wasn't sure if one was better than the other (especially considering normal maps) . Whichever way you want to go is fine.

src/core/WebGLPathTracer.js
DennisSmolek: oof ok I'll work on this refactor... I can use the `DenoiserManager` as the basis for this.

src/core/WebGLPathTracer.js
DennisSmolek: The pathtracer starts and stops multiple times during initialization it regularly hit the maxSamples and started the denoiser before there was anything in the scene. Blocking it this way worked. 

With the refactor to make the class external it might not be a problem

example/index.js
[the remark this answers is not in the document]
gkjohnson: See https://github.com/gkjohnson/three-gpu-pathtracer/pull/663#discussion_r1689741838

src/core/utils/AlbedoNormalPass.js
gkjohnson: I think using camera-space normals is the easiest to understand.

src/core/WebGLPathTracer.js
gkjohnson: Gotcha - I can help take a look at this afterward, as well.

src/core/utils/DenoiserManager.js
```diff
@@ -0,0 +1,208 @@
+// Utils to interface with the denoiser
+
+import { Denoiser } from 'denoiser';
+import { FullScreenQuad } from 'three/examples/jsm/postprocessing/Pass.js';
+import { ClampedInterpolationMaterial } from '../../materials/fullscreen/ClampedInterpolationMaterial';
+import { BlendMaterial } from '../../materials/fullscreen/BlendMaterial';
```
gkjohnson, line 6: nit: file extensions

src/core/utils/DenoiserManager.js
gkjohnson: > With the WebGL backend the first execution is SIGNIFICANTLY longer than the others. Tensorflow waits to compile shaders until you actually execute

Is there any way to use the [KHR_parallel_shader_compile](https://developer.mozilla.org/en-US/docs/Web/API/KHR_parallel_shader_compile) extension so the shaders will compile asynchronously? Three.js implements this with [WebGLRenderer.compileAsync](https://threejs.org/docs/?q=webglrenderer#api/en/renderers/WebGLRenderer.compileAsync), which the PathTracer uses [here](https://github.com/gkjohnson/three-gpu-pathtracer/blob/d5b1f148afa82d6f9a28be4870e8d9fb664e593c/src/core/PathTracingRenderer.js#L274) so the page does not block while the path tracing shader compiles.

src/core/utils/DenoiserManager.js
gkjohnson: Colors look a look a lot nicer with `t2conversion = true` but we have banding issues (see lighter color at the top) because we're storing linear srgb colors in 8 bits, causing precision loss. I'll trace the color space of the textures once some of the other changes are made. The quality is looking nice, though!

image

src/core/utils/AlbedoNormalPass.js
```diff
@@ -0,0 +1,224 @@
...
+
+			if ( ! this.originalMaterials.has( object ) )
+				this.originalMaterials.set( object, object.material );
+
+			if ( this.albedoMaterials.has( object ) ) {
+
+				object.material = this.albedoMaterials.get( object );
+				return;
+
+			}
```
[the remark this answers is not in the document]
gkjohnson: I just mean you keep a pool of materials and assign all the necessary material properties for each object when assigning:

```js
// before
meshes.forEach( mesh => {

  const material = materialPool.getMaterial();
  material.normalMap = mesh.material.normalMap;
  material.map = mesh.material.map;
  
  originalMaterials.set( mesh, mesh.material );
  mesh.material = material;

} );

// after
meshes.forEach( mesh => {

  mesh.material.normalMap = null;
  mesh.material.map = null;
  mesh.material = originalMaterials.get( mesh );

} );

originalMaterials.clear();
materialPool.reset();
```

This way you avoid keeping textures around on hidden internal material objects when the user has intended to get rid of them. And likewise you avoid using the Mesh object as a key which will also keep a reference around preventing it from being GC'd when the user has removed all references.

src/core/WebGLPathTracer.js
DennisSmolek: I still had to do this. It doesn't hurt things persay, but the size changes and that causes the model to have to be re-built which adds a lot of cold-start time. Doing this waits until the scene is ready to allow it to denoise.

src/core/WebGLPathTracer.js
DennisSmolek: Done!

src/core/utils/AlbedoNormalPass.js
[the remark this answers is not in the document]
DennisSmolek: In a different message you mentioned doing it in two passes being ok. And as we aren't doing it every frame I went ahead and went that route. Also the normals dont actually have to be physically correct, it's just to help the denoiser demark edges.

So I rewrote this whole thing.

Now we dont maintain anything. All the materials are stored on the mesh itself. So if it gets destroyed they go with it.
I share the plain normal material and clone when a normal map is present. 

The albedo shader is my first pass at responding to different material settings to set albedo instead of just a pure color. Its being weird for some parts so this is where I'll focus next.

I'll mention this more in my post

src/core/utils/DenoiserManager.js
DennisSmolek: After some battling I've worked this out. Also the denoiser hates the gradient background

DennisSmolek:
Screenshot 2024-07-25 022324
I've done most of the things on your list, 
Most importantly made the denoiser external `OIDNDenoiser` and I seem to have resolved the coloring issue for the denoised output at least.

It turns out we were passing Linear to the denoiser. Even though the render target and texture were set to SRGB it still output Linear. Adjusting the output manually in the shader corrected this.

Albedo & Normals were rewritten, still needs some work and the background always renders linear (weird)

I modified the `BlendMaterial` again as it's the best place to debug outputs/colors. I also used a more correct LinearToSRGB function.

Something helpful to check colors and interesting to compare is the "Split" property in the denoiser panel. This will divide the page between the pathtracer output and either the denoised result or the AuxTexture for comparison purposes.

Here you can see the sky is still wrong
Screenshot 2024-07-25 022450

A little messy still. I need a cleanup pass tomorrow.

src/core/utils/AlbedoNormalPass.js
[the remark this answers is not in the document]
gkjohnson: The code looks a lot cleaner by using the built in materials - nice work!

> So if it gets destroyed they go with it.

There is still a memory leak here since the materials on `userData` don't get disposed when an object GC'd so there will be dangling shader references left in the three.js renderer.

I think I still prefer the pool method since it's more self-contained and the pooled materials can be properly disposed when the path tracer and / or denoiser is disposed.

gkjohnson:
> Here you can see the sky is still wrong

This looks like a tone mapping issue - as in tone mapping is not applied to the final denoised image. If you disable "tone mapping" in the example UI the colors look the same:

image

src/core/utils/AlbedoNormalPass.js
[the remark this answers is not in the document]
DennisSmolek: Gotcha, I thought putting it in userdata would track the reference. I'll work on this now

example/index.js
```diff
@@ -29,6 +29,8 @@
 import { getScaledSettings } from './utils/getScaledSettings.js';
 import { LoaderElement } from './utils/LoaderElement.js';
 import { ParallelMeshBVHWorker } from 'three-mesh-bvh/src/workers/ParallelMeshBVHWorker.js';
+import path from 'path';
+import { on } from 'events';
```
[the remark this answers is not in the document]
DennisSmolek: do I have to reply to resolve this?

src/core/WebGLPathTracer.js
```diff
@@ -479,11 +579,51 @@
...
+		}
+
+		this.isDenoised = true;
+
+	}
+
+	//get albedo and normals from the current scene
+	async generateAux() {
+
+		if ( ! this._albedoNormalPass ) this._albedoNormalPass = new AlbedoNormalPass( this._renderer, true );
```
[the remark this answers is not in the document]
DennisSmolek: .

example/index.js
```diff
@@ -29,6 +29,8 @@
 import { getScaledSettings } from './utils/getScaledSettings.js';
 import { LoaderElement } from './utils/LoaderElement.js';
 import { ParallelMeshBVHWorker } from 'three-mesh-bvh/src/workers/ParallelMeshBVHWorker.js';
+import path from 'path';
```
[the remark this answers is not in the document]
DennisSmolek: ,

example/index.js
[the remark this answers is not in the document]
gkjohnson: It says "resolved" for me

gkjohnson:
> I actually already tonemap the texture prior to it going to the denoiser. 
> ...
> At the moment I'm super happy with the color result. The pathtraced and denoised look great.

The quality does look really nice - but there is some color shift once it goes through the denoiser. It's just not clear why to me, yet. I'll have to take a look this weekend.

src/materials/denoiser/createAlbedoMaterial.js
```diff
@@ -0,0 +1,128 @@
+import { Color, ShaderMaterial } from 'three';
+
+export function createAlbedoShaderMaterial( originalMaterial ) {
```
gkjohnson, line 3: Is there a reason we're not just using `MeshBasicMaterial` here? Once you start including things like metalness and  clearcoat and other specular terms you're no longer just rendering albedo.

DennisSmolek: We can but technically it would be incorrect (at least according to OIDN)
[Here in the docs](https://github.com/RenderKit/oidn?tab=readme-ov-file#albedos) They say:
> The albedo image is the feature image that usually provides the biggest quality improvement. It should contain the approximate color of the surfaces independent of illumination and viewing angle.

> For simple matte surfaces this means using the diffuse color/texture as the albedo. For other, more complex surfaces it is not always obvious what is the best way to compute the albedo, but the denoising filter is flexible to a certain extent and works well with differently computed albedos. Thus it is not necessary to compute the strict, exact albedo values but must be always between 0 and 1.

> For metallic surfaces the albedo should be either the reflectivity at normal incidence (e.g. from the artist friendly metallic Fresnel model) or the average reflectivity; or if these are constant (not textured) or unknown, the albedo can be simply 1 as well.

> The albedo for dielectric surfaces (e.g. glass) should be either 1 or, if the surface is perfect specular (i.e. has a delta BSDF), the Fresnel blend of the reflected and transmitted albedos. The latter usually works better but only if it does not introduce too much noise or the albedo is prefiltered. If noise is an issue, we recommend to split the path into a reflected and a transmitted path at the first hit, and perhaps fall back to an albedo of 1 for subsequent dielectric hits. The reflected albedo in itself can be used for mirror-like surfaces as well.

> The albedo for layered surfaces can be computed as the weighted sum of the albedos of the individual layers. Non-absorbing clear coat layers can be simply ignored (or the albedo of the perfect specular reflection can be used as well) but absorption should be taken into account.

Then regarding transparency and reflective items you can either be safe and just set it to 1 or use a Fresnel blend of the reflected and transmitted albedos.

`MeshBasic` will be correct for matte objects and might be ok for others, but this shader was an attempt to accommodate the other cases (metallic, roughness,  clearcoat, etc)

Besides the sky, I don't see much issues with this actually. It probably can be improved, a pathtraced albedo would be even better but would slow things down.

Totally up to you though.

DennisSmolek: The shader is being problematic and I get good results with meshBasic material. I'm removing the shader for now and will think about a strategy for transmission later

src/core/WebGLPathTracer.js
DennisSmolek: With MRT so common (especially with WebGPU) I think it a good idea to have this. it doesn't hurt anything and we can leave it out of the api docs for now. I've left it on the denoiser.

Technically they could still do `denoiser.auxTextures.albedo = myTexture;` but I like having the api for it

DennisSmolek:
After battling render flow, making a bunch of charts  I think I have things in a stable setup. There are still a few problems to resolve.

### Notable changes:

1. Simplified Index example, added Denoiser example with simpler controls but more denoiser controls
2. Got the color pipeline to match.

What the problem was: Threejs does not apply toneMapping to renderTargets. So even though I was copying the exact same textures/materials I was getting different outputs with not only different tonemapping but different colorSpaces. 
I have added (and removed) a few key conversions that take place so we are now sending a SRGB non-tonemapped Pathtraced Image, a SRGB converted Albedo, and a standard Normal to the Denoiser then doing the tonemapping in the final step.

If I recall correctly though this is the wrong order and the tonemapping should happen BEFORE we send to the denoiser for the best outcome (standard for every pipeline with ANY denoiser, not just OIDN) but I could be wrong.


### Problems:

1. Albedos just use basicMesh. I think this is the cause for some issues, most notably edges around the floor. If you set the floor to a 0 opacity and it has to render the background there is very little noise around the edges. But with the reflective floor (and a black albedo) it has issues. Transparency we'll have to test too. (this is a longer term issue to resolve)
2. Tiling seams. All the color conversions/pipelines/flow before & after denoising and not being tonemapped prior to the denoiser running, are causing seams in the tiling to become slightly visible. I dont know if this is the tiling from the pathtracer but I'm treating is as an issue with the denoiser.
I am working on better blending upstream in the main denoiser class but it's a little difficult to test. I'll setup something with this PR and a special version of the denoiser with the new tiling to test both systems out.
3. Speed. In general the speed is not a problem (especially on fast machines) but it can always be better. Initializing and tiling both lock up the thread. Upstream I am working on this so the denoiser will report progress of it's batches. This is a non blocking problem and one I'll continue to address. 

Please take a look and let me know what you think.

PathtracerDenoiserRenderFlow

example/index.html
```diff
@@ -1,374 +1,383 @@
 <!DOCTYPE html>
 <html>
-	<head>
```
gkjohnson, line 3: This file still has a ton of changes - I'm not sure why any of them would be needed?

example/index.js
```diff
@@ -91,7 +92,7 @@ const params = {
 
 	cameraProjection: 'Perspective',
 
-	backgroundType: 'Gradient',
+	backgroundType: 'Environment',
```
gkjohnson, line 95: I'd prefer not to change any of the settings in these examples - we should just add the denoiser.

example/index.js
```diff
@@ -269,11 +278,20 @@ function onParamsChange() {
 
 	}
 
+	// Denoiser
+	else pathTracer.maxSamples = params.maxSamples;
```
gkjohnson, line 282: Why are we conditionally setting max samples here?

example/denoiser.js
```diff
@@ -0,0 +1,779 @@
+import {
```
gkjohnson, line 1: I think ideally this example would just include a single model to keep it focused and simple. An ideal model would include something with opacity and transmission so we can evaluate these kinds of problem cases. See the HDR example or DoF example. Though this is something that can be adjusted later if needed.

gkjohnson:
> Got the color pipeline to match.

I can take a look at the colors

> If I recall correctly though this is the wrong order and the tonemapping should happen BEFORE we send to the denoiser for the best outcome

Do you have a reference for this? I would think ideally tone mapping is always applied before colors have been quantized into 8 bits to avoid banding but it shouldn't be a huge deal. We can take a look at this later I think.

> 1. ... Transparency we'll have to test too.

We should be able to just set the mesh opacity and alpha maps to help with this. Transmission is more complicated because we can't correctly model the refraction that happens in that case.

> 2. Tiling seams. ...

Interesting - I haven't seen any tiling seems before so my assumption is that this is related to the denoiser. I'll try to keep an eye on this why looking at the color processing.

> 3. Speed. ... Initializing and tiling both lock up the thread

Avoiding the lock up I think would be great. The two things that come to mind are performing tiles over multiple frames and performing async material compile so other tasks can run while compiling is happening.

An event or flag that specifies how complete the denoising process is (just 0 to 1) would let us display it. And likewise a way to "cancel" the denoiser for cases where the camera moves, scene changes, etc would be nice.

src/core/OIDNDenoiser.js
```diff
@@ -0,0 +1,297 @@
...
+	get hdr() {
+
+		return this.denoiser.hdr;
+
+	}
+
+	set hdr( v ) {
+
+		this.denoiser.hdr = v;
+		if ( v && this.quality === 'fast' ) this.quality = 'balanced';
```
gkjohnson, line 72: Why do we have to set the quality to "balanced" when enabling HDR?

src/core/OIDNDenoiser.js
```diff
@@ -0,0 +1,297 @@
...
+
+		this.denoiser.dirtyAux = ! v;
+
+	}
+
+	get hdr() {
+
+		return this.denoiser.hdr;
+
+	}
```
gkjohnson, line 67: Is `hdr` functional right now? It looks like we always pass sRGB into the denoiser

DennisSmolek: It was.
Technically the denosier can already support it.
While working out the color issues I matching the original system as much as possible. this included passing the hdr texture. 

When passing HDR and using "fast" quality I got HORRIBLE results, but using balanced was excellent. That's why I added the auto change to those props.

When trying to address the color issues in LDR I resolved it and found the system to be more stable with fast/balanced quality while using LDR.

If you don't ever plan to pass HDR to the denoiser or don't like leaving future handling code I can remove it.

src/core/OIDNDenoiser.js
DennisSmolek: Went into more details in the other comment, but essentially hdr and fast quality gave the worst results I've seen. Balanced gave good results.

example/denoiser.js
DennisSmolek: I get that, and we can certainly limit the models. But I think it easier to switch to models that highlight the edge cases specifically vs trying to find a single model that fits every one.
LMK what you think and I'll adjust it

example/index.js
DennisSmolek: leftover, will remove.

example/index.html
DennisSmolek: This is super weird. I have removed most of my changes from the original and something is making it mark everything as changes. 

I will revert the file and put my changes

example/index.js
DennisSmolek: No worries. My fear is how bad the denoiser treats the gradient background compared to envMaps. But I'll revert it

src/core/OIDNDenoiser.js
gkjohnson: Gotcha - is it the case that we just have to pass the original LinearSRGB Path Traced texture when `hdr` is true and we'll get a LinearSRGB texture out? I can try it a bit to see how it looks.

DennisSmolek:
I've discovered what is adding all the weird stuff to the package.json

[Parcel polyfills when it finds node methods: ](https://parceljs.org/features/node-emulation/#polyfilling-%26-excluding-builtin-node-modules)

Screenshot 2024-07-26 205319

The code that starts all this is using `node-fetch` which I definitely don't. But searching I found in [RenderVideo](https://github.com/gkjohnson/three-gpu-pathtracer/blob/d5b1f148afa82d6f9a28be4870e8d9fb664e593c/example/renderVideo.js#L20)s build file it mentioned. It was in a package for `ffmpeg` which is loaded by the package `canvas-capture`

So when parcel hit that package it started doing it's own thing...

It's a dev-dependency so doesn't add to the bundle size but it's annoying I get that.

Does it not do this when y'all build the package? It runs every time I use `npm start`

src/core/OIDNDenoiser.js
DennisSmolek: Yes, if we pass that texture as the input the denoiser will return it in the same range (or at least it should, I've only tested a few times)
OIDN says that in HDR 1 should be a luminance level of 100 cd/m^2
so for best results it should be similar.

If you do that you may want to disable the SRGBToLinear conversion that takes place on the denoiser result. I do this so we can correctly apply the ToneMapper.

One thing to consider would be dropping the SRGB > Denoiser requirement. This would mean textures dont need to be converted when returned and can be tonemapped directly

DennisSmolek:
Made cleanup/minor changes

Denoiser has a more soft tiler but I now see the issue is only really visible in places the denoiser gets confused. For example the floor with 0 albedo but a reflective surface. Hiding the floor or setting its color to white (albedo white too) And the issue isn't visible.

Added a onProgress upstream and attached handlers to it as well as UI for the denoiser.

Abort will come tomorrow.

In general I'm happy with the colors. It's fast and matches the original.

Transparency will need to be resolved but we can spend time on that after merge. 

Parcel is still weird...

Last little blocking issue IMO is there's a weird few pixel border being added. It's in the upstream one too so I'll spend time tomorrow chasing that down.

example/denoiser.js
gkjohnson: I think just one model is good for now - for debugging we can swap between models to check issues. I think it's best to have a more simple example for users to reference.

DennisSmolek:
Weekend update:
- Merged in @gkjohnson 's two PR's with much better Albedo & Normals handling. 
Still some work to be done there with reflectance but I think that's something we'll be tweaking down the road.
- These PR's removed the Split/Preview feature of the Denoise Demo. Not sure the best way to add them back. Easiest would be to put the split function in the clampedInterpolate, or add a dedicated split material. But I figured you might want to keep the library simple so I figured I make that material in the demo itself as a way to show how to get both outputs. Added to todo list.
- Upstream: Added Abort to main class and Tiler. 
Abort runs Async, but we can't stop a model in process but once it comes back from the GPU it does the "aborted" check. For now this works, although I think I'll add more abort checks through the process to potentially exit faster. But if the first build starts for example I can't stop it. It does stop the output so hopefully no more garbled outputs from an aborted execution. 
- Abort wired to `reset()` and fires the progress to 0. Not sure this is the right place to put it, but it seems to work. 
- Upstream: Added position specific masks and a mask cache. This resolves white borders on edges. 
- Upstream: Added even more complex smoother
- Upstream: Increased standard overlap of tiles to 32 so more blending can happen

These upstream changes to the tiler are to resolve the borders & Seams seen in gradients of colors (particularly over bad albedos). Borders are gone, I might see a phantom of the grid  over areas with 0 albedos when in fast quality, but in balanced I no longer see any of the grid anywhere.

DennisSmolek:
Removed old materials, added DenoiserSplitMaterial and loaded it in the Denoiser example to show how to split outputs from the denoiser in realtime.

I honestly think the ability to flip through different models to see what works best with the denoiser and what it struggles with is very helpful. Especially now until we can work out those finer details ( Like the ring and gelatinous cube)

If the goal of the denoiser example is to show off how to interface with and how the denoiser works I think it's probably the best it can be right now. 

Let me know what you think left and I can work on it.

It would be great if we can get it merged this week, I'd like to work on the upstream examples/docs and the library will likely be moving to Poimandres but I don't want to do that until it's solid here.

src/materials/fullscreen/ClampedInterpolationMaterial.js
```diff
@@ -56,22 +56,37 @@ export class ClampedInterpolationMaterial extends ShaderMaterial {
 			fragmentShader: /* glsl */`
 				uniform sampler2D map;
 				uniform float opacity;
+				uniform bool convertToSRGB;
```
gkjohnson, line 59: none of these changes look like they're needed anymore.

src/materials/fullscreen/LinearToSRGBMaterial.js
```diff
@@ -0,0 +1,78 @@
+import { ShaderMaterial } from 'three';
+
+// Material that tone maps a texture before performing interpolation to prevent
+// unexpected high values during texture stretching interpolation.
+// Emulates browser image stretching
+export class LinearToSRGBMaterial extends ShaderMaterial {
```
gkjohnson, line 6: This class is no longer used

gkjohnson (commented):
Thanks for the changes -

> I honestly think the ability to flip through different models to see what works best with the denoiser and what it struggles with is very helpful. Especially now until we can work out those finer details ( Like the ring and gelatinous cube)

I suppose we can leave it for now but all of these models can be checked in the main example with denoising and urls copied into the denoiser example with more options for dev. These example files have a significant maintenance burden whenever APIs change or need to be updated so keeping them simple is a priority for me.

In terms for functionality and correctness there are still a few things I'm not certain on. I'm happy to help look at any of these but these seem related to the internal "denoiser" package components which I'm not familiar with. If there's somewhere specific you want a second set of eyes I can take a look:

- Normals are currently [stored as sRGB](https://github.com/gkjohnson/three-gpu-pathtracer/pull/663/files#diff-b30d6e57656e9a5805a6a68fc41b245e9283a3bd37edfb9e43891e9576285b39R180) which is atypical. Are we sure the normal image is supposed to be sRGB? The data stored in a normals texture is not considered color data and is usually loaded and read without color space transformations, even when reading from image files.

- I've mentioned it a few times but is it possible to precompile the tensor flow shaders with the [KHR_parallel_shader_compile](https://developer.mozilla.org/en-US/docs/Web/API/KHR_parallel_shader_compile) extension to prevent locking the page?

- I've noticed the denoiser can stall for quite awhile at 99% but it's not clear to me from profiling what's causing this. Are you seeing it, as well?

- It looks like instantiating the Denoiser has a side effect enabling the scissor test for some reason:

```js
const gl = renderer.getContext();
console.log( gl.getParameter( gl.SCISSOR_TEST ) ); // false
this.denoiser = new Denoiser( 'webgl', renderer.domElement );
console.log( gl.getParameter( gl.SCISSOR_TEST ) ); // true
```

- I'm not sure if you saw [this issue here](https://github.com/DennisSmolek/three-gpu-pathtracer/pull/1#discussion_r1692964844) - but it looks like once the desnoiser has started you can't perform any other draw calls without corrupting the denoised image.

- White colors are still getting significantly blown out by the denoise process and it doesn't seem right? We wind up losing a lot of detail in brighter parts of the image for some reason. I'm not sure how difficult it is to test just passing the canvas directly into the denoiser again to see if it's causing the same problem? It looks like a color conversion may be happening somewhere causing us to clamp or lose precision or something.

These images are without tone mapping applied:

| Before | After |
|---|---|
| image | image |

src/core/OIDNDenoiser.js
```diff
@@ -0,0 +1,298 @@
...
+		// Extract the raw webGLTextures
+		const colorWebGLTexture = this.getWebGLTexture( this.pathtracedTexture );
+		const albedoWebGLTexture = this.getWebGLTexture( albedoTexture );
+		const normalWebGLTexture = this.getWebGLTexture( normalTexture );
+
+		//* run the denoiser ----------------------------
+		this.renderer.resetState();
+
+		// Run the denoiser
+		const denoisedWebGLTexture = await this.denoiser.execute( colorWebGLTexture, albedoWebGLTexture, normalWebGLTexture );
```
gkjohnson, line 160: This looks like a memory leak - does the model always return a brand new texture? Do we ever dispose of the webgl memory?

src/core/OIDNDenoiser.js
```diff
@@ -0,0 +1,298 @@
...
+		const oldRenderTarget = this.renderer.getRenderTarget();
+		this.renderer.setRenderTarget( tempRT );
+		this.quad.render( this.renderer );
+		this.renderer.setRenderTarget( oldRenderTarget );
+
+		// get the texture out of the tempRT
+		const texture = tempRT.texture;
+		// dispose?
+		return texture;
+
```
gkjohnson, line 241: What's the purpose of rendering to the new render target here?

src/core/OIDNDenoiser.js
DennisSmolek: The Albedo and Normal ones are just references to the original threejs textures so get disposed when those textures would. Not sure how that works in three, either when the renderTarget renders etc.

The `WebGLTexture` out of the denoiser is bound to the `outputTensor` of the denoiser flow. When the next model is executed and a new output is generated, the old output is disposed, destroying the bound texture.

src/core/OIDNDenoiser.js
```diff
@@ -0,0 +1,298 @@
...
+		this.handleProgress( 0 );
+
+	}
+
+	//* Listeners
+
+	handleProgress( progress ) {
+
+		let outProgress = progress * 100;
+		// we only return 100 when isDenoised is true but the denoiser progress is for tiling
```
gkjohnson, line 256: What does this mean exactly? The denoiser package is returning 100% completion but it's not actually done?

src/core/OIDNDenoiser.js
gkjohnson: The albedo and normal textures I understand - the returned handle from `denoiser.execute` is what I was wondering about. It sound like the life cycle of this handle is internally managed, then.

DennisSmolek:
> I suppose we can leave it for now but all of these models can be checked in the main example with denoising and urls copied into the denoiser example with more options for dev. These example files have a significant maintenance burden whenever APIs change or need to be updated so keeping them simple is a priority for me.

For sure not trying to add burden, I'm just copying the examples that are already in use with the PBR example so if they change or get removed it's just as easy to remove them here. Yes you can run the denoiser in the PBR example, but you can't run the split comparison or look at how the transparent, reflective, metallic, etc objects generate albedos/normals etc. Instead of finding a super-model that tests everything in one go, it might be easier to use the same models we already load and see how they compare to each other. 

Whatever you want do do though is fine with me. I can always make my own example on the main denoiser site with more models. 

> 
> * Normals are currently [stored as sRGB](https://github.com/gkjohnson/three-gpu-pathtracer/pull/663/files#diff-b30d6e57656e9a5805a6a68fc41b245e9283a3bd37edfb9e43891e9576285b39R180) which is atypical. Are we sure the normal image is supposed to be sRGB? The data stored in a normals texture is not considered color data and is usually loaded and read without color space transformations, even when reading from image files.

This is a mistake. This comes from battling color conversions. Regardless of the colorspace of inputs Normals should be in Linear. I will adjust this, The denosier doesn't actually use normals in the common sense, but this will better match training data. 

> * I've mentioned it a few times but is it possible to precompile the tensor flow shaders with the KHR_parallel_shader_compile](https://developer.mozilla.org/en-US/docs/Web/API/KHR_parallel_shader_compile) extension to prevent locking the page?

It's actually implemented from what I can tell but it's burried in github discussions and not doccumented. [I will experiment with this upstream.](https://github.com/DennisSmolek/Denoiser/issues/30)

> * I've noticed the denoiser can stall for quite awhile at 99% but it's not clear to me from profiling what's causing this. Are you seeing it, as well?

I am going to work on a flowchart of the denoiser so it's easier to explain, but right now the progress is only measuring the inference/tiling stage which executes with frame blocking to allow the callbacks. The next steps execute in a single step and theres no way to call back progress until it's totally done. The biggest delay is probably syncing from gpu to cpu. 
Over time the upstream denoiser will get better at "progress" but for now I block at 99% waiting for the renderable `WebGLTexture` to return before finalizing with 100%. 100% is also a "done" marker at this stage too. 

> * It looks like instantiating the Denoiser has a side effect enabling the scissor test for some reason:
> 
> ```js
> const gl = renderer.getContext();
> console.log( gl.getParameter( gl.SCISSOR_TEST ) ); // false
> this.denoiser = new Denoiser( 'webgl', renderer.domElement );
> console.log( gl.getParameter( gl.SCISSOR_TEST ) ); // true
> ```

Weird, I've added it to my list to put in the WebGLStateManager class (although I thought it was there)


> * I'm not sure if you saw [this issue here](https://github.com/DennisSmolek/three-gpu-pathtracer/pull/1#discussion_r1692964844) - but it looks like once the desnoiser has started you can't perform any other draw calls without corrupting the denoised image.

Calling a draw call means Threejs takes over the state. I'm sure this messes things up. With the new abort flag the denoiser shouldn't output anything after the abort is called so if you need to draw, abort the denoiser and it should stop anything returning. It's a new feature so if it doesn't work right lmk and I'll try and test for it


> * White colors are still getting significantly blown out by the denoise process and it doesn't seem right? We wind up losing a lot of detail in brighter parts of the image for some reason. I'm not sure how difficult it is to test just passing the canvas directly into the denoiser again to see if it's causing the same problem? It looks like a color conversion may be happening somewhere causing us to clamp or lose precision or something.
> 

Weird, I thought this got resolved with the correct tonemapping etc. 

We can actually test pretty easily. using `usepassThrough` Honestly, I think this is probably an issue with all the color conversion s and working in SRGB not linear.
I've read a lot of people saying to send input in as SRGB but I think thats not actually the right way. OIDN itself defaults to Linear input, and I"ve actually dug in and found that when you use SRGB it converts the inputs to linear before processing and then reapplies SRGB to output. 
Along with the normals I'll do some tests with a Linear version and I'll add a trigger for passthrough on the denoiser example (we can remove it later)


Hopefully it's just a colorspace thing. The next thing I could think would be an albedos/normal issue? One thing I did also try is using more samples. Using more samples seemed to stop the denoiser guessing so much on things..

src/core/OIDNDenoiser.js
DennisSmolek: Yeah, I'll try to explain better.
To get the WebGL Texture we run something like:
```ts
const outputWebGLTexture = outputTensor.dataToGPU();
```

This `WebGLTexture` is what is bound to the tensor in the backend and what we inject into threejs. 

when the denoiser is executed again I destroy the old output tensor which destroys the WebGLTexture it is bound to.
So no worries here.

src/core/OIDNDenoiser.js
DennisSmolek: Right now the denoiser package reports on the percentage of the tiler and not as a whole. The tiler is the main process that does not block the thread but I will look into adding other progress callbacks if possible.

src/core/OIDNDenoiser.js
```diff
@@ -0,0 +1,282 @@
...
+		const colorWebGLTexture = this.getWebGLTexture( this.srgbPathTracedTarget.texture );
+		const albedoWebGLTexture = this.getWebGLTexture( albedoTexture );
+		const normalWebGLTexture = this.getWebGLTexture( normalTexture );
+
+		// Run the denoiser
+		this.renderer.resetState();
+		const denoisedWebGLTexture = await this.denoiser.execute( colorWebGLTexture, albedoWebGLTexture, normalWebGLTexture );
+		this.renderer.resetState();
+
+		if ( ! this.outputTexture ) this.outputTexture = this.createOutputTexture();
```
gkjohnson, line 166: I'm not exactly sure what might be going on but when I move this intialization to the constructor I get the following error from `WebGLStateManager` in `denoiser` and the screen is white, instead. The width and height are changed to 1 in the function, as well.

I'm seeing the following warning logged after moving the function:

```
webglStateManager.ts:54 WebGL: INVALID_OPERATION: bindBuffer: attempt to use a deleted object
```

DennisSmolek: It might be because in the constructor we create the Denoiser which normally has a "ready" callback that we currently ignore.
If you try to create that texture using the renderer you have tensorflow initializing the WebGL backend and the renderer trying to draw at the same time.

The two systems fight over the state and something breaks. 

That's just my guess from my earlier experience getting the two WebGL systems to play well.

If you don't want the conditional creation like it is now we can try adding the `onReady` listener to the denoiser creation before firing. I can play with this...

gkjohnson: Hm. Even if we move the creation of the texture to before the `Denoiser` creation it has the same issue.

I'm also not fully following how these systems could be fighting or in what way they're asynchronous. Is denoiser or OIDN using "requestAnimationFrame" or "setTimeout" or a webworker something? Otherwise no other JS in main thread can run until all those WebGL calls are issued.

DennisSmolek: I say we just leave it where it is then

DennisSmolek:
Something in the recent commits has broken things...

When you rotate after denoise the denoised image turns to the blank canvas.

Also, the denoiser is being marked dirty on every run so some property is being set over and over.

I'll try to go through the commits to find it

DennisSmolek:
Ok the marking dirty is from removing the check if the size has changed. I will work on this upstream to protect against it.

However the blend not blending is still a problem. Both introduced in "some cleanup"

gkjohnson:
> Something in the recent commits has broken things...
>
> When you rotate after denoise the denoised image turns to the blank canvas.

Just push a line I forgot to commit - sorry about that. You're talking specifically about the denoiser.html example, right? I'm not seeing issues in the index.html page.

> However the blend not blending is still a problem.

What does "blend not blending" mean?

src/core/OIDNDenoiser.js
gkjohnson: We can do that for now - it's more about understanding what's going on and whether it's indicative of other issues. I'll just mark it with a comment and TODO.

gkjohnson: I think the normal texture color space, color blow out issues, and image corruption issue mentioned in https://github.com/gkjohnson/three-gpu-pathtracer/pull/663#pullrequestreview-2204761468 are things I'd like to address or understand more before merging. The rest we can make a new issue to track for later.

I'm a bit concerned about the image corruption because I think users will immediately run into this. We can control whether the path tracer renders or uses the context while denoising but not users and this is a bit of an odd restriction.

DennisSmolek:
Latest Version changes:

- Upstream: Protect against same values marking dirty (stops denoiser rebuilding every run making it SIGNIFICANTLY faster)
- Upstrream: Warmstart using Parallel shader compile
- Upstream: WebGLStateManager improvements
- Upstream: Cleanup
- Upstream: Slight model change looking for speed improvement

- Linear colorspace: All render targets (pathtracer, albedo, normals) now run in Linear colorspace no conversion to SRGB used.
- Names changed to reflect operation not colorspace
- Minor adjustments and fixes
- Debug controls added for now (probably removed later)

I spent time with the models to check color/brightness. In general with the linear flow I like the output better but I need to check the lego model and see if it's doing the same thing.

My main check was colors, the floor, the sky
Screenshot 2024-07-31 231020

Color areas usually checked out. 
I did notice if I zoomed out the NASA label got slightly bright, and when I checked the albedos at this level, it matched the higher brightness. So this is still one thing to keep messing with.

Regarding the state. I added restoring external state to the state manager. so it not only restores it's own state but the running state. This works fine with my original pathtracer setup, but this version it has conflicts every other execution. Luckily I added a "ignoreRestore" flag to the state manager and this seems to correct the every other WebGL error (instance draw without buffer) My guess is something in three changes and when the state restore is run it's out of date. Mixing webGL state is super fragile so for now I think this the best we can get.

Shader compiling/warm starting added, however, there is a minor speed improvement, but for the most part it seems it still waits to send data to the GPU util execution. There might be some exploration here but I don't suspect much gain.

I still get 20-30 second load times to compile the GPU Pathtracer shaders when first starting btw.

Minor model changes. My Mac version now executes 500-1000ms and my PC I've dipped below 1000 but that was probably a smaller size. most of the time my PC (3060) gets ~3500ms

I'm curious to see how WebGPU compares but I'll save that for another day.

DennisSmolek:
Ok spent some time on the X-Wing..
This has a split slightly left of center. You can see the colors themselves are all matching correct so doubt it's a color space issue
Frame 2

Tested the normals and albedo and got this result:
compared


My theory: The 100% white Albedo is being treated as a transparent surface. Combined with the fairly uniform normals it thinks it's a super reflective/transparent surface and is taking the albedo map as the color source. 

I'm not sure what else it could be and it's something we can mess with. The colors in the scene are all correct looking including the backgrounds. 

One thing to note, my normals are significantly different than the ones in the new pass.
Normals

A key to remember for the denoiser is not so much that normals are 100% light correct, but that I think they act as a segmentation mask helping find the edges of shapes. 

This is all just a theory, but I tested the regular X-Wing which uses grey not white, and it renders fine.

src/core/OIDNDenoiser.js
```diff
@@ -0,0 +1,283 @@
...
+	constructor( renderer ) {
+
+		this.renderer = renderer;
+		this.denoiser = new Denoiser( 'webgl', renderer.domElement );
+		this.denoiser.inputMode = 'webgl';
+		this.denoiser.outputMode = 'webgl';
+		this.denoiser.weightsUrl = 'https://cdn.jsdelivr.net/npm/denoiser/tzas';
+		this.denoiser.onProgress( progress => this.handleProgress( progress ) );
+		//this.denoiser.hdr = true;
+		this.denoiser.webglStateManager.ingoreRestore = true;
```
gkjohnson, line 94: spelling issue - ingoreRestore -> ignoreRestore. Can we add an inline comment describing why we need this

gkjohnson:
> Upstrream: Warmstart using Parallel shader compile
> ...
> Shader compiling/warm starting added, however, there is a minor speed improvement, but for the most part it seems it still waits to send data to the GPU util execution.

Nice! Can you point to where this parallel compile is being done? If you attempt to use the shader after starting the parallel compile before it's done then the driver will still stall the page. The goal here is less of a speed improvement and more preventing the page from stalling.

> I still get 20-30 second load times to compile the GPU Pathtracer shaders when first starting btw.

This is an issue with DirectX that's unavoidable. For some reason it takes forever to compile shaders. The Chrome / ANGLE team have said it's a known issue. The use of async compile should mean the page isn't completely locked while the shader compiles, though. [This isCompiling flag](https://github.com/gkjohnson/three-gpu-pathtracer/blob/46b395e6d521f031eb5f357d49cd36cf831d0129/src/core/WebGLPathTracer.js#L97-L101) is used in `renderSample` to prevent the path tracing material from being used until it's ready.

> Ok spent some time on the X-Wing..

I took a look today and it seems to be due to clamping colors when converting from the Float texture to Uint8 target. This clamps and loses data above 1.0 used in the final tone mapping step which is why we're losing some of the bright highlights in the final image.

The solution is to either denoise the final tone mapped image or deoise the hdr path traced image and get an hdr result. Denoising the hdr image would be the ideal solution since it can be nice to adjust tone mapping, exposure, and post effects like bloom (which requires hdr anyway) without another denoising step. But I'm not sure how much work getting hdr working is.

> One thing to note, in my other renderer, the normals are significantly different than the ones in the new pass.

The normals on the left are packed into the range [ 0, 1 ] - looking at the OIDN docs again it looks like they're supposed to be [in the range [ -1, 1 ]](https://github.com/RenderKit/oidn?tab=readme-ov-file#normals), which I must have misread previously:

> Just like any other input image, the normal image should be anti-aliased (i.e. by accumulating the normalized normals per pixel). The final accumulated normals do not have to be normalized but must be in the [-1, 1] range (i.e. normals mapped to [0, 1] are not acceptable and must be remapped to [−1, 1]).

What format are we supposed to feed into the denoiser, then? The current format used for the normals texture will clamp the values of the output normals and negative values will just be black.

This is looking good, though. Looks like just a few final things.

DennisSmolek:
> 1. The solution is to either denoise the final tone mapped image or deoise the hdr path traced image and get an hdr result. Denoising the hdr image would be the ideal solution since it can be nice to adjust tone mapping, exposure, and post effects like bloom (which requires hdr anyway) without another denoising step. But I'm not sure how much work getting hdr working is.

So, we can try working in HDR, but is that the output you want from the pathtracer? The denoiser will output HDR if we want it to. But I didn't know if that's what the pathtracer should send to the texture. We'll need to test a bit as I recall it had some issues. Otherwise I wonder if we can move the clamping to after the tonemap step.

> 2. What format are we supposed to feed into the denoiser, then? The current format used for the normals texture will clamp the values of the output normals and negative values will just be black.

I'm not sure what you mean exactly, right now I take a standard normal texture as input and re-map it. In my base renderer this seems to work ok.

gkjohnson:
> Regarding HDR. I've had hit or miss results with it and found the standard def had better results. We can start working in HDR but wouldn't that make everything HDR or would we clamp it down at the end?

The goal would be to denoise in HDR then tone map the image to SDR. We could tone map before denoising to fix the color clamping but this way is more flexible and let's the user change the tone mapping without denoising and also let's you save a denoised HDR image. Right now the final result from the path tracer in a rendertarget is an HDR float buffer that is tone mapped when when written to the canvas.


> Regarding Normals: I'll have to double check but I'm pretty sure I'm already remapping to [-1, 1] during the texture setting phase (I assume most users and shaders will use [0, 1])

What I'm wondering is what happens when a component is negative. If using an rgb8 texture (as the normal textures are here) any negative components will be clamped to 0 and there will be no distinction in normals. What's right thing to do here? Should we be using float textures? Does OIDN support that?

fabiopolimeni:
Have been following the denoising thread for quite some time, then the project went silent. Would make sense to merge this PR even if denoising is happening in SDR?

DennisSmolek:
> Have been following the denoising thread for quite some time, then the project went silent. Would make sense to merge this PR even if denoising is happening in SDR?

That's on me, I got busy with some other projects and lost the track on this one. I have a big post it note saying to get this done.

I'll make it a priority in a few weeks at the start of the year

fabiopolimeni:
> > Have been following the denoising thread for quite some time, then the project went silent. Would make sense to merge this PR even if denoising is happening in SDR?
> 
> That's on me, I got busy with some other projects and lost the track on this one. I have a big post it note saying to get this done.
> 
> I'll make it a priority in a few weeks at the start of the year

That would be great, and make the path tracer very useful indeed

Milesssssss:
Have an update plan？

gkjohnson:
Closing in favor of WebGPU support, which should help a lot of the issues we ran into here.

Issue #662, opened 2024-07-22
Material Index not matched


dongho-shin:
**Describe the bug**

<img width="325" alt="Untitled (16)" src="https://github.com/user-attachments/assets/9782e35f-c081-44d9-996c-2ba91e90d981">

this model has all the same Materials (8 mesh but share 1 material, share 1 material(same uuid) is a key point of this bug)

1. PathTrace Rendering first after import this model
2. change only material of mesh
<img width="361" alt="image" src="https://github.com/user-attachments/assets/970d95ca-c44c-4e43-8177-b3dc66630d04">

3. PathTrace render again

<img width="406" alt="image" src="https://github.com/user-attachments/assets/4e337267-5ce2-438a-8cb5-c9f9df2942c9">

**Model**

FYI) zip from mac

[Blender_Smeg (2).glb.zip](https://github.com/user-attachments/files/16330712/Blender_Smeg.2.glb.zip)

**Platform:**

 - Device: Desktop
 - OS: MacOS
 - Browser: Chrome
 - Three.js version: r162
 - Library version: 0.0.23

This bug causes here
https://github.com/gkjohnson/three-gpu-pathtracer/blob/d5b1f148afa82d6f9a28be4870e8d9fb664e593c/src/core/utils/GeometryPreparationUtils.js#L46

### sorry for ugly notes
indexOf, if all meshes share 1 materials(in this notes a) geometry.materialIndex attributes are all 0
![Screenshot_20240722_155500_Samsung Notes](https://github.com/user-attachments/assets/f4dbc509-29ff-4463-a218-e8950638bd76)
and then material of mesh has changed(a -> b)
![Screenshot_20240722_155534_Samsung Notes](https://github.com/user-attachments/assets/9aef14e4-f111-49ca-b640-3f91ac74e0b6)

StaticGeometryGenerator, bvhChanged => no changes(because in this scenario no geometry changes) so materialIndex not updated, geometry reuse materialIndex 0 so nothing has been updated

so my sugestion is
before
```js
		const mat = Array.isArray( materials ) ? materials[ group.materialIndex ] : materials;
		const materialIndex = allMaterials.indexOf( mat );
```
after
```js
	//	const mat = Array.isArray( materials ) ? materials[ group.materialIndex ] : materials; // not used
		const materialIndex = group.materialIndex;
```

always welcome any ideas & questions

gkjohnson:
Thanks for reporting and digging in as always!

[...]

PR #664, opened 2024-07-28
Fix Material Index not matched

dongho-shin:
ref .https://github.com/gkjohnson/three-gpu-pathtracer/issues/662


gkjohnson:
Excellent, thank you!

Issue #665, opened 2024-07-29
PathTracingSceneGenerator: only update material indices if material assignments have changed


gkjohnson:
Related to #662
Related to #664 

Right now the material index attribute is _always_ updated and _always_ reuploaded which is in-optimal. This can be addressed by caching the previous set of materials and checking if it's been modified before updating the material attribute and updating the texture.

gkjohnson:
Fixed in #666

PR #666, opened 2024-07-29
updateMaterialIndexAttribute when it needs

dongho-shin:
related : https://github.com/gkjohnson/three-gpu-pathtracer/issues/665

<h1>This needs to be tested to make sure it's okay</h1> but don't worry, I'll test it, if it's okay, I'll comment  <br/> the reason for the PR is that I wonder this solution is right way.

### DIFF
--- a/src/core/PathTracingSceneGenerator.js
+++ b/src/core/PathTracingSceneGenerator.js
@@ -180,7 +180,7 @@ export class PathTracingSceneGenerator {
 		let needsMaterialIndexUpdate = result.changeType !== NO_CHANGE || this._materialUuids === null || this._materialUuids.length !== length;
 		if ( ! needsMaterialIndexUpdate ) {
 
-			for ( let i = 0, length = materials.length; ( i < length ); i ++ ) {
+			for ( let i = 0, length = materials.length; i < length; i ++ ) {
 
 				const material = materials[ i ];
 				if ( material.uuid !== this._materialUuids[ i ] ) {
--- a/src/core/PathTracingSceneGenerator.js
+++ b/src/core/PathTracingSceneGenerator.js
@@ -73,6 +73,8 @@ function getLights( objects ) {
 
 }
 
+let materialUuids = null;
+
 export class PathTracingSceneGenerator {
 
 	get initialized() {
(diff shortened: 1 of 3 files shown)


src/core/PathTracingSceneGenerator.js
```diff
@@ -176,9 +178,29 @@ export class PathTracingSceneGenerator {
 		// generate the geometry
 		const result = staticGeometryGenerator.generate( geometry );
 		const materials = result.materials;
+		let needsMaterialIndexUpdate = result.changeType !== NO_CHANGE || materialUuids === null || materialUuids.length !== length;
+		for ( let index = 0, length = materials.length; ( index < length ) && ! needsMaterialIndexUpdate; index ++ ) {
+
+			if ( materialUuids === null || materialUuids.length !== length ) {
```
gkjohnson, line 184: These conditions can be pulled out of the for loop and used in the above `needsMaterialIndexUpdate` initialization since it's not using the index value at all.

src/core/PathTracingSceneGenerator.js
```diff
@@ -176,9 +178,29 @@ export class PathTracingSceneGenerator {
 		// generate the geometry
 		const result = staticGeometryGenerator.generate( geometry );
 		const materials = result.materials;
+		let needsMaterialIndexUpdate = result.changeType !== NO_CHANGE || materialUuids === null || materialUuids.length !== length;
+		for ( let index = 0, length = materials.length; ( index < length ) && ! needsMaterialIndexUpdate; index ++ ) {
```
gkjohnson, line 182: nit: Lets rename `index` to `i` for consistency with other loops.

And for readability if we could move the `needsMaterialIndexUpdate` check to a condition block that would best, I think:

```js
if ( ! needsMaterialIndexUpdate ) {

  for ( ... ) {

    // ...

  }

}
```

src/core/PathTracingSceneGenerator.js
```diff
@@ -176,9 +178,29 @@ export class PathTracingSceneGenerator {
...
+
+			if ( materialUuids === null || materialUuids.length !== length ) {
+
+				needsMaterialIndexUpdate = true;
+				break;
+
+			}
+
+			const material = materials[ index ];
+			if ( material.uuid !== materialUuids[ index ] ) needsMaterialIndexUpdate = true;
```
gkjohnson, line 192: We can add a `break` statement once this is set to true.

src/core/PathTracingSceneGenerator.js
```diff
@@ -176,9 +178,29 @@ export class PathTracingSceneGenerator {
...
+
+		}
+
+		materialUuids = materials.map( material => material.uuid );
```
gkjohnson, line 196: This can move inside the below condition since the uuids won't have changed otherwise.

```js
if ( needsMaterialIndexUpdate ) {

  updateMaterialIndexAttribute( geometry, materials, materials );
  materialUuids = materials.map( material => material.uuid );

}
```

src/core/PathTracingSceneGenerator.js
```diff
@@ -73,6 +73,8 @@ function getLights( objects ) {
 
 }
 
+let materialUuids = null;
+
```
gkjohnson, line 77: This should be assigned on the `PathTracingSceneGenerator` instance itself. Otherwise if we have two different instances of the generator created they will overwrite each others results.

gkjohnson (commented):
This is great! I've added some comments on small stylistic preference but overall the logic here looks right!

dongho-shin:
fix styles as you recommended(also duplicated logic removed)
also I test it, lgtm

gkjohnson:
Looks excellent! Thank you

Issue #668, opened 2024-08-07
Model disappeared when same 3 models pathtraced


dongho-shin:
**Describe the bug**

in three.js editor you can reproduce 100%

1. Create 1 cube(other models also reproduce)
2. PathTrace Render(set Realistic upper right)
3. Create 2 cubes
4. again PathTrace Render then you can see 2 cubes disappear in pathTraced Scene

https://github.com/user-attachments/assets/bc288c6e-af14-4428-a36e-53dbb0c90a35

### Only this scenario reproduce this bug
add 1 model -> pathTrace render -> add same model of two -> PathTrace render

**Expected behavior**

all 3 cubes rendered in realistic

**Platform:**

 - Device: [Desktop]
 - OS: [MacOS]
 - GPU: [NVidia ??, Radeon ??, ...]
 - Browser: [Chrome]
 - Three.js version: [r167(editor)]
 - Library version: [current library version]

gkjohnson:
Can you reproduce in a jsfiddle or smaller example? Something like initializing / rendering the initial scene with 1 cube and then 1 second later (ore pressing a button) adding two more cubes?

dongho-shin:
@gkjohnson I found a reason but not perfectly digged
this line is a problem
PathTracingSceneGenerator.js in line.217
```js
if ( result.changeType === GEOMETRY_REBUILT ) {

				const bvhOptions = {
					strategy: SAH,
					maxLeafTris: 1,
					indirect: true, //<-- this options causes a problem
					onProgress,
					...this.bvhOptions,
				};
```
because most(react three fiber, etc...) of indirect option is false as default, but PathTracingSceneGenerator this.generateBVH true as a default

so MeshBVH instantiates indirectBuffer at PathTracingSceneGenerator constructor, and MeshBVH's indirect check like below code (MeshBVH line.111)
```js
	get indirect() {

		return ! ! this._indirectBuffer;

	}
```
so it causes a bug. it affects many codes ex. MeshBVHUniformStruct updateFrom // indirect false but true

dongho-shin:
currently I make a smaller example to reproduce a bug, then pr for check an example

dongho-shin:
Open a pr for check a bug @gkjohnson

dongho-shin:
I found a **reason** and why it's 3 same model causes a bug
if indirect set true in bvhOptions it causes an error because bvhToTextures not supported Multi root Nodes
```js
[...]

PR #671, opened 2024-08-12
Reproduce a bug example, (Model disappeared when same 3 models pathtraced)

dongho-shin:
related : #668 

https://github.com/user-attachments/assets/fa341864-2654-4ec8-8eaa-67a5b30d3685

### how to reproduce a bug

1. check enable
2. uncheck enable
3. check add model
4. check enable

### DIFF
--- a/example/primitives.js
+++ b/example/primitives.js
@@ -1,4 +1,4 @@
-import { Scene, SphereGeometry, MeshStandardMaterial, Mesh, BoxGeometry, PerspectiveCamera, ACESFilmicToneMapping, WebGLRenderer } from 'three';
+import { Scene, SphereGeometry, MeshStandardMaterial, Mesh, PerspectiveCamera, ACESFilmicToneMapping, WebGLRenderer } from 'three';
 import { WebGLPathTracer, GradientEquirectTexture } from '..';
 import { getScaledSettings } from './utils/getScaledSettings.js';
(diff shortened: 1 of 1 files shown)


dongho-shin:
Sorry for long time to fix this @gkjohnson ,
Actually this issue was caused by MeshBVHUniformStruct.js wrong length comparison
image
cachedIndexAttr is still being used as an index attribute, but the
indirectBuffer is a Buffer for reference
It is 1/3 the size of the geometry index count because of that part below
image (1)

So the first time it's cached, the simple box's case calls
count(undefined)
length(1) ← default dummy data before adding any boxes at all, which is subsequently deleted.
Add 1 box
→ count(3) length(12 box 1 index data)
Add 2 additional boxes after the string
→ count(36) length(36 box 2 boxes added to match)
So the index length was not growing properly, so the data addition was not working properly.

After the caching comparison, the following logic
image (2)
image (3)

Because of adding by indirectBuffer, the properly cached IndexAttr was not tripled accordingly, so the data was not added.

gkjohnson:
With the WebGPUPathTracer becoming the priority for the project I'm going through and closing all issues related to the WebGLPathTracer specifically to try to clean things up. Once the WebGPUPathTracer is released we'll plan to remove the WebGLPathTracer from the project (see https://github.com/gkjohnson/three-gpu-pathtracer/issues/779).

Issue #692, opened 2025-09-16
Architecture & Plans for WebGPU Support


gkjohnson:
Following up on #691 & #647 - with WebGPURenderer becoming more widely supported and compute shaders being supported in three-mesh-bvh I wanted to lay out some of the architecture and features that a “WebGPUPathTracer” could support. This will break things down into the architecture for getting path tracing working with compute, extra features beyond what was reasonably possible with WebGL, and separate optimizations for improving the practically of the renderer. This issue should be considered a sort of “epic” for the overall work. We can create dedicated issues to discuss some of the subtasks if needed.

The webgpu-specific work can go into a `src/webgpu` folder initially and exported via `three-gpu-pathtracer/webgpu`.

**Other Articles**
- https://research.nvidia.com/publication/2013-07_megakernels-considered-harmful-wavefront-path-tracing-gpus
- https://developer.blender.org/docs/features/cycles/kernel_scheduling/

#### PathTracerCore

This is the upgraded, WebGPU analog to the [PathTracingRenderer](https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/src/core/PathTracingRenderer.js): the core class that path traces a given BVH-packed scene to a render target and manages blending, tiling, settings for the rendering, manage state being render iterations, and creates intermediate buffers and uniforms for AA, randomized values, etc.

The primary architectural changes from the previous iteration of the path tracer is that this will be written in WGSL compute shaders with three.js’ TSL. The basic render iteration could look like the following, based on the concept of “wavefront path tracing”:

- Update any uniforms to change per frame render for AA, etc.
- Choose a “tile” of the screen to render.
- Run a compute kernel to generate camera rays to a ray queue.
- Run a kernel to intersect the scene and add a series of surface hits with material properties (or material indices) to a shade queue.
- Run a kernel to fragment color from surface shading and generate a next ray and shadow ray to evaluate.
- Run a kernel (or iteration of kernels) to evaluate direct light sampling / “shadow” rays or (later on) indirect light sampling using bi directional path tracing (see #246).

All materials, etc will now be able to be uploaded as structures rather than encoded to textures. The compute kernels can be evaluated in a loop in a generator function so the number of rays and operations can be limited per frame to reduce the performance impact.

#### WebGPUPathTracer

[...]

Issue #198, opened 2022-07-04
Add ability to report noise on each pass, allows halting on a noise threshold


bhouston:
**Is your feature request related to a problem? Please describe.**

It would be nice to be able to run a render until a specific noise threshold is reached.  Such as you want noise to be under 0.1% or similar.  This is to the settings in V-Ray and others where you set a noise threshold and the iterative rendering halts at that point.

https://evermotion.org/tutorials/show/12467/v-ray-settings-explained-by-jamie-cardoso

**Describe the solution you'd like**

It is quite easy to implement various halt strategies as long as we can report on each render iteration what the noise floor is currently.

I wonder if one can just compare the previous pixel value with the new pixel value after each pass and calculate what fraction of its RGB values were changed and then sum this up across the whole image.

Values that change very little are likely (I think) close to their final value, where as those who are varying significant are not.

Something along this lines seem relatively straight forward to implement.  Maybe there is a clear standard somewhere for this type of noise calculation?

gkjohnson:
[This stackexchange post](https://blender.stackexchange.com/questions/188137/optimal-noise-threshold-for-adaptive-sampling-with-ai-denoiser) seems to indicate that Blender does something similar to what you're suggesting but I'm trying to understand exactly how it would work per pixel.

Presumably we compare after the new pixel is averaged into the image - so pixels like the background would have zero difference between pixels after the first sample (assuming no blurring) while most pixels would be changing more dramatically especially at the beginning of a render. Though I guess the most we'd expect a pixel to change per frame is `1.0 / num_samples`.

I'd think we'd want to **keep track of the change over multiple frames** to ensure it isn't oscillating or resolved one sample that happened to be the same as the existing fragment. We'd need also need a setting to **render a minimum number of samples**.

To start I think this can be a separate helper class like `NoiseDetector` that provides a noise delta between two images, accumulates the results to a single pixel, and reads it back. Or we could read the whole buffer of noise detection values back and sum them in a worker. Not sure which would be better.

Eventually something like this per-pixel convergence threshold could be used to adaptively sample certain parts of the scene - ignoring the background or shiny surfaces that resolve quickly by masking them with a stencil buffer or second target and focusing samples on the noisy bits of the scene.

PR #705, opened 2025-11-15
[WIP] Webgpu pathtracer prototype

TheBlek:
Hi @gkjohnson! Here's a simple lambertian pathtracer. Both megakernel and wavefront, as discussed in #692. See webgpu_primitives demo. There is still a lot of work to be done here. I will try to keep an up-to-date todo list here so you could follow the progress.

At what point do you propose measuring performance?

I think it is not so difficult to have both megakernel and wavefront pathtracers as they share the essential code (for now at least). And it could be useful to have a point of reference when implementing additional features. And I'm not sure how relative performance will scale with features.

In webgpu by default we can bind maximum 128Mb buffers and there is no way to request bigger limits in three.js. So that leaves us with max tile resolution of 1920x1080 (which I hardcoded in the example for now). I'm wondering whether requesting bigger webgpu limits or features for that matter is a worthwhile feature for three.js. What do you think?

This buffer limit also limits the scene size at least until the TLAS is deployed and tracing is per-mesh instead of on the whole scene.

Also by default we can bind a maximum of 8 buffers for use which is a bottleneck for traceRay kernel. It needs a lot of buffers so I had to move reading material index into the next kernel (bsdf eval phase). This could be fixed if three.js had support for array types in structs and top-level storage buffer backed structs. At least I could not get it to work, so I had to put queue sizes into a separate buffer and pass separately. I also wonder to write an issue about the limitation in three.js but maybe I just didn't find it. 

~~What's your policy on three.js version for webgpu release here? Ideally we would use r181 as it includes support for indirect compute (which is essential for wavefront pathtracing to not read-back values from GPU between calls). But it released like 5 days ago :) For now I copy-pasted compute methods from WebGpuRenderer and WebGpuBackend with little tweaks in order to compute indirectly. see computeIndirect and computeBackendIndirect functions in PathTracerCore.~~

Just saw that the library recently upgraded to r181 - perfect!

Would love any feedback!

TODO:
- [ ] Try SoA memory layout for path state
- [ ] Try path regeneration for megakernel
- [ ] Add support for conventional light sources
- [ ] Add support for emission
- [ ] Merge changes from #731 and create a PR back

### DIFF
--- a/src/webgpu/PathTracerCore.js
+++ b/src/webgpu/PathTracerCore.js
@@ -2,6 +2,7 @@ import {
 	IndirectStorageBufferAttribute,
 	StorageBufferAttribute,
 	Vector2,
+	Matrix4,
 	TimestampQuery,
 	StorageTexture,
 	RGBAFormat,
@@ -18,6 +19,7 @@ import { LogicKernel } from './compute/wavefront2/LogicKernel.js';
 import { EvaluateMaterialKernel } from './compute/wavefront2/EvaluateMaterialKernel.js';
 import { ResetKernel } from './compute/wavefront2/ResetKernel.js';
 import { ZeroOutKernel } from './compute/ZeroOutKernel.js';
+import { EquirectHdrInfoUniform } from '../uniforms/EquirectHdrInfoUniform.js';
 
 function* renderTask() {
 
@@ -187,6 +189,8 @@ export class PathTracerCore {
 			materials: new StorageBufferAttribute(),
 		};
 
+		this.envInfo = new EquirectHdrInfoUniform();
+
 		this.outputTarget = new StorageTexture( 1, 1, );
 		this.outputTarget.format = RGBAFormat;
 		this.outputTarget.type = FloatType;
@@ -255,7 +259,6 @@ export class PathTracerCore {
 		this.materialDispatchBuffer = new IndirectStorageBufferAttribute( new Uint32Array( 3 ) );
 		this.materialDispatchBuffer.name = 'Dispatch Buffer for Material Eval Kernel';
 
-
 	}
 
 	setGeometryData( geometry ) {
@@ -283,6 +286,23 @@ export class PathTracerCore {
 			this.geometry[ propName ] = geometry[ propName ];
 
 		}
+
+	}
+
+	setEnvironment( environment, intensity, rotation, blur ) {
+
+		if ( environment !== null ) {
+
+			this.envInfo.updateFrom( environment );
+			this.logicKernel.envMap = this.envInfo.map;
+			this.logicKernel.kernel.computeNode.parameters.envMapSampler.node.value = this.envInfo.map;
+
+		}
+
+		const rotationMatrix = new Matrix4().makeRotationFromEuler( rotation ).invert();
+		this.logicKernel.envMapRotation.setFromMatrix4( rotationMatrix );
+		this.logicKernel.envMapIntensity = intensity;
+		this.logicKernel.envMapBlur = blur;
 
 	}
 
--- a/src/webgpu/nodes/sampling.wgsl.js
+++ b/src/webgpu/nodes/sampling.wgsl.js
@@ -1371,7 +1371,7 @@ const sampleHemisphereFn = wgslFn( /* wgsl */ `
 
 	}
 
-` );
+`, [ constants ] );
 
 export const sampleBackgroundFn = wgslFn( /* wgsl */ `
 
--- a/src/webgpu/MegaKernelPathTracer.js
+++ b/src/webgpu/MegaKernelPathTracer.js
@@ -33,10 +33,6 @@ function* renderTask() {
 	kernel.inverseProjectionMatrix.copy( camera.projectionMatrixInverse );
 	kernel.cameraToModelMatrix.copy( camera.matrixWorld );
 
-	kernel.envMap = envInfo.map;
-	kernel.kernel.computeNode.parameters.envMapSampler.node.value = envInfo.map;
-	kernel.envMapBlur = 0;
-
 	while ( true ) {
 
 		const tileSize = this.getTileSize( kernel.tileSize );
@@ -144,12 +140,20 @@ export class MegaKernelPathTracer {
 
 	}
 
-	setEnvironment( environment, intensity, rotation ) {
+	setEnvironment( environment, intensity, rotation, blur ) {
 
-		this.envInfo.updateFrom( environment );
+		if ( environment !== null ) {
+
+			this.envInfo.updateFrom( environment );
+			this.kernel.envMap = this.envInfo.map;
+			this.kernel.kernel.computeNode.parameters.envMapSampler.node.value = this.envInfo.map;
+
+		}
+
 		const rotationMatrix = new Matrix4().makeRotationFromEuler( rotation ).invert();
 		this.kernel.envMapRotation.setFromMatrix4( rotationMatrix );
 		this.kernel.envMapIntensity = intensity;
+		this.kernel.envMapBlur = blur;
 
 	}
 
--- a/src/webgpu/WaveFrontPathTracer.js
+++ b/src/webgpu/WaveFrontPathTracer.js
@@ -1,4 +1,4 @@
-import { IndirectStorageBufferAttribute, StorageTexture, Vector2, FloatType, RGBAFormat, LinearFilter, RedIntegerFormat, UnsignedIntType, ColorManagement } from 'three/webgpu';
+import { Matrix4, IndirectStorageBufferAttribute, StorageTexture, Vector2, FloatType, RGBAFormat, LinearFilter, RedIntegerFormat, UnsignedIntType, ColorManagement } from 'three/webgpu';
 import { ZeroOutKernel } from './compute/ZeroOutKernel.js';
 import { PrimeRayGenerationDispatchKernel } from './compute/wavefront/PrimeRayGenerationDispatchKernel.js';
 import { RayGenerationKernel } from './compute/wavefront/RayGenerationKernel.js';
@@ -6,6 +6,7 @@ import { RayIntersectionKernel } from './compute/wavefront/RayIntersectionKernel
 import { UpdateRayQueueParamsKernel } from './compute/wavefront/UpdateRayQueueParamsKernel.js';
 import { ZeroOutBufferKernel } from './compute/ZeroOutBufferKernel.js';
 import { ProcessHitsKernel } from './compute/wavefront/ProcessHitsKernel.js';
+import { EquirectHdrInfoUniform } from '../uniforms/EquirectHdrInfoUniform.js';
 
 // set the buffers to the max possible size supported by default (128MB)
 // TODO: this can be increased based on platform.
@@ -169,6 +170,8 @@ export class WaveFrontPathTracer {
 			materials: null,
 		};
 
+		this.envInfo = new EquirectHdrInfoUniform();
+
 		// targets
 		this.outputTarget = new StorageTexture( 1, 1, );
 		this.outputTarget.format = RGBAFormat;
@@ -238,7 +241,20 @@ export class WaveFrontPathTracer {
 
 	}
 
-	setEnvironment( environemnt ) {
+	setEnvironment( environment, intensity, rotation, blur ) {
+
+		if ( environment !== null ) {
+
+			this.envInfo.updateFrom( environment );
+			this.rayIntersectionKernel.envMap = this.envInfo.map;
+			this.rayIntersectionKernel.kernel.computeNode.parameters.envMapSampler.node.value = this.envInfo.map;
+
+		}
+
+		const rotationMatrix = new Matrix4().makeRotationFromEuler( rotation ).invert();
+		this.rayIntersectionKernel.envMapRotation.setFromMatrix4( rotationMatrix );
+		this.rayIntersectionKernel.envMapIntensity = intensity;
+		this.rayIntersectionKernel.envMapBlur = blur;
 
 	}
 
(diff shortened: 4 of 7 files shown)


src/webgpu/nodes/random.wgsl.js
```diff
@@ -0,0 +1,45 @@
...
+export const pcgStateStruct = wgsl( /* wgsl */`
+	struct PcgState {
+		s0: vec4u,
+		s1: vec4u,
+		pixel: vec2i,
+	};
+` );
+
+export const pcgInit = wgslFn( /* wgsl */`
+	fn pcg_initialize(state: ptr<function, PcgState>, p: vec2u, frame: u32) -> void {
```
gkjohnson: nit: is there a reason we can't store the "state" variable as a global variable? It seems more simple than passing a state argument in every time.

src/webgpu/nodes/reset.wgsl.js
```diff
@@ -0,0 +1,21 @@
+import { wgslFn } from 'three/tsl';
+
+export const resetFn = wgslFn( /* wgsl */ `
+
+	fn reset(
```
gkjohnson: nit: Can we name this function and comment it so it's more clear what it's resetting? What is this for?

src/webgpu/nodes/sampling.wgsl.js
```diff
@@ -0,0 +1,63 @@
...
+
+		return vec4f( direction, pdf );
+	}
+` );
+
+
+export const lambertBsdfFunc = wgslFn( /* wgsl */`
+	fn bsdfEval(rngState: ptr<function, PcgState>, normal: vec3f, view: vec3f) -> ScatterRecord {
+
+		const PI: f32 = 3.141592653589793;
```
gkjohnson: We should probably move this to a constants node.

src/webgpu/PathTracerCore.js
```diff
@@ -0,0 +1,484 @@
...
+
+		} else {
+
+			const dispatchSize = [
+				Math.ceil( dimensions.x / WORKGROUP_SIZE[ 0 ] ),
+				Math.ceil( dimensions.y / WORKGROUP_SIZE[ 1 ] ),
+				1,
+			];
+			// 0. Clean queues
+			_renderer.compute( this.cleanQueuesKernel, 1 );
```
gkjohnson: I'm wondering if we want to / need to clear these kernels? The buffer size limits may make this more complicated but I was originally thinking of these queues as ongoing "work queues" that would have work constantly pushed onto them over multiple frames. Rays from still-unterminated paths from sample 1 could still be in the queue while rays from sample 2 (or 3 or 4) are being pushed onto the same queue, for example. Then there can always be work available on the queue for the compute nodes to pick up. 

"Finishing" a render could mean a few things:

- The maximum samples have been reached for every pixel and the last ray from the queue is complete.
- Some pixels can be terminated early if they are determined to already be "resolved" enough (see #198)

Small tiles can be run to start off new paths and fill up the queue as needed. This would also let users really tune in how much work they want to do per frame - rather than samples the use can say "only process X rays per frame" and that can be something that's auto-tunable.

And maybe a separate thought but I'm also wondering if sorting rays can somehow help with performance and thread utilization.

src/webgpu/nodes/wavefront.wgsl.js
```diff
@@ -0,0 +1,227 @@
...
+
+		atomicStore(&queueSizes[0], 0);
+
+		let count = atomicLoad(&queueSizes[1]);
+		outputBuffer[0] = u32( ceil( f32(count) / f32( workgroupSize ) ) );
+		outputBuffer[1] = 1;
+		outputBuffer[2] = 1;
+	}
+`, );
+
```
gkjohnson: It looks like these are just copying data to a indirect buffers so we can issue indirect compute commands, is that right? Is it the case that we can't use atomic storage buffers are indirect compute arguments?

src/webgpu/PathTracerCore.js
```diff
@@ -0,0 +1,484 @@
...
+
+				// 2. Handle escaped and scattered rays
+				// 2.1 Calcuate dispatch sizes and write it to gpu buffer
+				_renderer.compute( this.writeEscapedRayDispatchSizeKernel, 1 );
+				_renderer.compute( this.writeBsdfDispatchSizeKernel, 1 );
+				// 2.1 Dispatch shaders
+				// When processing escaped rays, calculate new contribution and add that to the result image
+				_renderer.compute( this.escapedRayKernel, this.escapedRayDispatchBuffer );
+				// When processing hit results, sample a new ray according to material's properties
+				_renderer.compute( this.bsdfEvalKernel, this.bsdfDispatchBuffer );
```
gkjohnson: I'm wonder if it's possible to merge this "escaped ray" kernel into one of the other ones?

src/webgpu/nodes/random.wgsl.js
TheBlek: I did not want to think how to do that in nodes system initially. Will take a look if that's possible.

src/webgpu/nodes/reset.wgsl.js
TheBlek: This shader resets result and sample count buffers, so any accumulated color stored becomes and sample count becomes 0. Will adjust name to better reflect the meaning

src/webgpu/PathTracerCore.js
TheBlek: > rather than samples the use can say "only process X rays per frame" and that can be something that's auto-tunable.

I see. This sounds interesting and promising in terms of adapting to hardware's capabilities. However, I think this should be done later, after the essential stuff is setup. Right now, ray generation kernel fully loads the ray queue with rays without checking for overflows so we have to clean the queues.

> And maybe a separate thought but I'm also wondering if sorting rays can somehow help with performance and thread utilization.

This is an interesting idea to explore. I can see how sorting ray in a certain way can improve thread utilization. Certainly a topic to explore. 

Perhaps we should create issues for those and return later on?

gkjohnson: I agree we can take a look at these later. I added some small notes to #692.

> > rather than samples the use can say "only process X rays per frame" and that can be something that's auto-tunable.
>
> I see. This sounds interesting and promising in terms of adapting to hardware's capabilities. However, I think this should be done later

Once the resolution of the render in the demo is right and we can get this merged I'll take a look at this if that's okay. I'm pretty interested in it

src/webgpu/nodes/wavefront.wgsl.js
TheBlek: > It looks like these are just copying data to a indirect buffers so we can issue indirect compute commands, is that right? 

Yeah, we are writing the dispatch size in the indirect buffer.

> Is it the case that we can't use atomic storage buffers are indirect compute arguments?

Well, we possibly could reuse storage buffer for atomics as one of the indirect buffers, but we cant get rid of them all. In the case of escaped rays and bsdf sampling we need two dispatch buffers and if we use queue sizes buffer for one of them, relevant values for other will be lost.

Theoretically we could create only one dispatch buffer for everything and dispatch with offsets. So values [0; 3) of the array will be dispatch size for escaped ray and [3; 6) - for bsdf sample kernel. But I think this is not supported in three.js right now

src/webgpu/nodes/random.wgsl.js
gkjohnson: Agreed! I was just curious if there was a reason it couldn't be done.

src/webgpu/PathTracerCore.js
TheBlek: It is possible. However separation is deliberate and should help with thread utilization. The main reason why wavefront pathtracing would be faster than regular megakernel is more unified kernel execution duration and hence better thread utilization. 

This was my thinking when writing this.

However I can see now, that the [original article](https://jacco.ompf2.com/2019/07/18/wavefront-path-tracing/) had a unified "Shade stage" for escaped and intersected rays. We could experiment with this in the future, once the performance baseline is established.

gkjohnson:
I'm going to try to take look at this this week / weekend but I just ran it and I'm seeing the following errors in the console with a white screen when running "webgpu_primitives":

image

And when disabling "useMegakernal" I see this:

image

I'm not sure if this is intended but just n FYI

gkjohnson:
> How would like to proceed with the feature? Should this PR be closed in favor of https://github.com/gkjohnson/three-gpu-pathtracer/pull/713 ? Is help on this still welcome?

Of course help is welcome! Your work here has been incredibly helpful and #713 is based on this branch - but I've moved things around in the process of understanding everything and trying to address some of the tiling questions.

Hopefully I'll have a first working pass at a wavefront tracer using a cyclic buffer in the next few days and we can see how the timing works out. But even so there are a lot of things that can happen in parallel. Working on the PBR materials as you've been doing here, for example, is fairly separable. Some of the other things that are becoming clear as I learn more about WebGPU compute:

- Interleaving and packing geometry buffer attributes into a single buffer rather than a bunch of separate storage buffers would help reduce the number of necessary storage buffers, which there's apparently a limit for. It would also make it easier to fetch the full set of vertex data. I'm not sure how a larger struct would impact the BVH traversal, though.
- Exploring using the new three-mesh-bvh [ObjectBVH](https://github.com/gkjohnson/three-mesh-bvh/blob/master/example/src/bvh/ObjectBVH.js) for storing a top-level acceleration structure for things like the scene and instanced / batched mesh geometries would be big improvement and make it easier to initialize and render the scene.
- In terms of path tracing architecture - bidirectional path tracing is something I'd like to incorporate, but it's something I'm less familiar with if you have any knowledge or interest here.
- Background transparency is something I'd like to get figured out.

Regarding the PRs - I'm happy to merge the other PR into this one or we can start making PRs into the webgpu-pathtracer branch. Either one is good to me to treat as a the "webgpu dev" branch once a basic wavefront implementation is one.

TheBlek:
> Interleaving and packing geometry buffer attributes into a single buffer rather than a bunch of separate storage buffers

I've been thinking of packing only color + uv + normals into one attributes buffer. (maybe material index?) And leave position in a separate buffer to not hinder ray tracing performance. That also allows for more vertices to be used in a geometry. There is a default limit of 128Mb for a buffer size. So that's <2.8mil vertices if packing every attribute into one buffer.

> Hopefully I'll have a first working pass at a wavefront tracer using a cyclic buffer in the next few days and we can see how the timing works out.

I am curious to try my hand at different architectures to try and optimize the pathtracer. I would like to try to make an architecture in parallel and then compare performance, if that's okay with you. 

With a consistent ray queue 'execution time' becomes a less relevant metric because a megakernel 'sample' and a wavefront 'sample' cast different amount of rays. Ray count seems to be a good metric only in closed scene because there could be rays that go directly to the environment evaluation like in webgpu_primitives. So perhaps a different scene is needed to properly assess performance characteristics.

webgpu-pathtracer branch seems a lot more structured so I will change this PR's target branch and resolve conflicts, hopefully soon.

gkjohnson:
Thanks! There's a lot here so I'm going to focus on the path tracing logic - I think it would be best to keep things like PBR materials to a separate PR so it's easier to follow and discussed what's changed. But thanks for putting this together -

One of the issues that sticks out to me is that even though a subsequent tile is adding rays to the queue it's possible, if not likely, that pixels will be passed over because paths from the previous tile are taking up the available path state slots. Setting the number of iterations to "1" is an extreme example but demonstrates it in this simple scene. In a more complex scene with more bounces you can imagine this happening much more frequently resulting in less even sample distribution across the image:

<div align="center">
image
</div>

These are separable issues, though, and I think there may be some value to being able to keeping path state in a single buffer while modeling the other buffers as uints but it's hard to say what may be a benefit or not with large sets of changes. Ultimately I think the two approaches we've taken are pretty well-aligned and there's likely a solution in the middle that will be a good place to land an initial version of the path tracer logic. It might be best to make smaller more targeted changes so we can discuss the merit of these things and test the performance in a more targeted way (eg switch to a logic kernel, switch to a path state buffer, etc).

To some of the other comments:

> I think we should enable two different textures and copy them only on devices that do not support read_write f32 texture. How critical do you think this is?

To start I think we should just use the same code path for every device and we can use a "ping pong" approach for blending the sample textures. It might be more optimal to check the capabilities and use a single texture if possible but every option like this is another level of complexity and something that needs to be tested so I think it will be easiest to support just one code path for the moment.

> Material samples with pdf <= 0 tripped me up on a number of occasions. Why would code sample directions that are basically impossible?

There are two separate normals per triangle - the "geometric" normals which represent the actual geometry used for raycasting against and "shading" normals which represent the interpolated normals used to imply a rounded surface. When these normals don't align a ray can effectively be coming from "underneath" the shading normal hemisphere resulting in an invalid PDF. The PBR Book mentions it [in section 9.1](https://pbr-book.org/3ed-2018/Materials/BSDFs). There are probably better solutions for this but they haven't been implemented here, yet.

> WaveFrontPathTracer's image appear to become unchanged after a quick while. That looks very much like pRNG being predictable or same after some point. However I could not find a bug. Maybe its more clear to you?

It's working in the other branch so it must be related to changes made in this one. I'm not sure if it's related but I see that the "QueuedHit" struct was modified to add a "side" field without adjusting the structure size.

> WaveFrontPathTracer suffers from low utilization when appending paths to the queue. Empty indirectDispatches seem to seep time. It seems that queue can only handle 1-2 tiles at a time. Maybe we can cycle through tiles, not trying to add all of them all the time?

One of the reasons I marched through all tiles is was to prepare, at least notionally, for skipping pixels that have already been "finished". Once this is the case the marching through only 1 or 2 tiles may only yield a few new rays added to the queue, which would be a different issue. Is it actually the case that dispatching an empty compute call is a time sink? Or is it an artifact of how the timing is being measured? I would this this would be fairly cheap, though I agree that Ideally we would be able to skip it entirely.

PR #713, opened 2026-02-04
WebGPU Support

gkjohnson:
cc @theblek

I've branched from #705 and changed things around quite a bit to address the storage buffer limitations by using storage textures and organized kernels into dedicated classes. So canvas resize etc all works. I have also separated the "MegaKernal" from the "PathTracerCore" so it's easier it will be easier to follow the differences and dependencies between the implementations.

Next I'm going to look into some of the ideas around a ray queue we'd discussed previously. Then we can try some timing to see how things pan out.

image

Relatedly, this write up will be interesting for a wave front path tracer:

https://developer.blender.org/docs/features/cycles/kernel_scheduling/

<details>
<summary>Plans</summary>

### Features
- *Adjust queue sizes based on needs, allow more control over exact number of rays handled per frame
- Next event estimation
- Fog (subsurface scattering?)
- Sheen support
- Transmission support
- Volume attenuation
- Anisotropy
- Dielectric / Transmission multiscatter compensation
- Thin transmission
- MNEE?
- Dispersion?

### Performance
- smarter background, env caching
- *Remove auto-inclusion of uncommonly needed attributes like "color"
- Include support for expanding storage buffer sizes
- *Pack geometry data more tightly, account for alignment

### Validation
- Evaluate performance
- Validate viewer tests
- WebGLPathTracer parity

### Other
- Issue announcing deprecation and future removal of WebGL version, close WebGL-related issues
- Cleanup
- *architecture  updates, code consistency 
- Move compute data class to three-mesh-bvh
- *replace demos

</details>

<details>
<summary>Future</summary>

### Features
- Add "debug" views (sample count, completion visualization, etc)
- Bidirectional path tracing
- Custom material BSDF, definitions
- *Add variance detection
- Add "completion" detection
- WebGPU Denoiser
  - Render normal, albedo to separate array texture layers

### Performnce
- Pack multiple UVs into a vec2 array
- Add async BVH generation
  - Requires ability to async generate BVHs for skinned meshes, etc.
- concurrent handling of rays for the same pixel
- provide "minimal" structs to reduce fetching overhead from storage buffers (eg a struct with only position for bvh traversal). The same buffer can be bound with different struct layouts.
- *consider adding "normalMatrix" to transform struct to avoid recalculation
- Inspect size limits for geometry (non indexed crab)
  - Convert geometry definitions to textures
  - *Allow for expanding storage sizes (needs to be specified at construction time), (sizes should just work?)
- Reuse uniforms, data across backend swaps 
- Add support for partially-updating / fast-updating bvh data
  - just transforms
  - just visibility
  - rewrite / refit bvhs
  - just materials
  - per-mesh, material, etc?
  - adjust ObjectBVH to sort objects based on uuid before construction for reliable ordering

### Other
- wgslTagFn, wgslFn, etc for general usage: add support for an on-the-fly transpiled code node so nodes can be used in both backends.

</details>


gkjohnson:
@TheBlek - I'm seeing some errors when raycasting against particularly small triangles:

image

This happens when setting the sphere geometry radius to a small value and compensating by increasing the ball scale:

```js
const sphereGeom = new SphereGeometry( 0.1, 64, 32 );
// ...
ball1.scale.setScalar( 5 );
ball2.scale.setScalar( 5 );
ball3.scale.setScalar( 5 );
```

The issue is probably a combination of small edge lengths being used for cross products in the triangle intersection test as well as the ray direction length becoming small when transforming into the local coordinate frame of 5x scale transform matrix (ray direction will be transformed to 1/5th size). I'm wondering if theres a way to make the triangle intersection test more robust to this? Or some other approach to prevent these kinds of issues?

_edit_

If we _don't_ scale the transforms up then things look okay, so this may primarily be an issue of the ray direction length?

image

gkjohnson:
Here's another article from AMD on GPU-based path tracing that might have some good insight, as well:

https://gpuopen.com/download/2025_RT_TechReport.pdf

gkjohnson:
An open BSDF implementation from Adobe - maybe good for reference: https://github.com/adobe/openpbr-bsdf

It only implements unidirectional pathtracing, though - not bidirectional. I'm not sure how complicated it is to derive the sibling PDF needed for bidirectional.

gkjohnson:
@TheBlek - not sure if you've seen this but this repo seems like it may have some good references for some the BRDFs:

https://github.com/AcademySoftwareFoundation/OpenPBR

gkjohnson:
Looks like "MNEE" may be something to look at, as well, for transparency. This tweet looks pretty nice:

https://x.com/invalid_degen/status/2065979027790143943

And [the paper](https://jo.dreggn.org/home/2015_mnee.pdf).

PR #717, opened 2026-02-19
WebGPU: Support for custom materials; simple glTF-compliant material

TheBlek:
- Adds webgpu_viewerTest.html example which is almost 1 to 1 with viewerTest.html example
- Adds pbr material bsdf which is ported from glsl version
- Packs color, normal, tangent and uv into attributes buffer
- Adds support for sampling material attributes from textures
- Wavefront: pcgState is passed between stages for good distribution of pseudo random numbers

chrome_yU5Inp6Krv

Transmissive materials need support from the light transport algorithm and will not work correctly for now. I think they can be addressed later.

@gkjohnson Should I divide this PR into multiple for better readability?


gkjohnson:
I forgot to mention that the [material database](https://gkjohnson.github.io/three-gpu-pathtracer/example/bundle/materialDatabase.html) demo can also be useful for testing against materials and Blender screenshots from [physicallybased.info](https://physicallybased.info/), which can be good for evaluating against something like cycles.

TheBlek:
@gkjohnson Hi, sorry for the long wait, I've been going through the math a number of times here. Now this PR adds:
1. Ability to replace BRDF implementation for WebGPUPathTracer (a la fully custom material); Maybe it shouldn't be called a material to not confuse with three.js materials?
2. `GltfCompliantMaterial` implements blending exactly as specified in base spec. Although some of the functions and arguments are not needed (or will not be needed?)
3. Diffuse: disney brdf
4. Specular: GGX

There is a problem with brights fireflies appearing some of the time, so I had to clamp impact of individual ray to 4.0 but I'm not sure they should be there in the first place. It seems we would need to port filter glossy as well.

While on that I noticed a couple of things that could be improved in webgl version:
1. Optimised GGX evaluation routines. It is possible to get rid of all `acos` and `tan` function calls with some clever trigonometry. See FrostBite's implementation in Listing 2 here: https://seblagarde.wordpress.com/wp-content/uploads/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf
2. Mistake in disney diffuse brdf. webgl code:
```glsl
float fl = schlickFresnel( wi.z, 0.0 );
float fv = schlickFresnel( wo.z, 0.0 );

float rr = 0.5 + 2.0 * surf.roughness * fl * fl;
float retro = rr * ( fl + fv + fl * fv * ( rr - 1.0f ) );
float lambert = ( 1.0f - 0.5f * fl ) * ( 1.0f - 0.5f * fv );
```
[Paper](https://blog.selfshadow.com/publications/s2015-shading-course/burley/s2015_pbs_disney_bsdf_notes.pdf):
firefox_1cb0j63KVM
From [earlier paper](https://media.disneyanimation.com/uploads/production/publication_asset/48/asset/s2012_pbs_disney_brdf_notes_v3.pdf):
```
θd is the “difference” angle between l and the half vector (or, symmetrically, v and h)
```
Note the `Rr` formulation. It should be `rr = 2.0 * surf.roughness * pow( dot( wo, wh ), 2 )`

I think it would be good to fix those issues in glsl as well, in another PR

gkjohnson:
This is great! Thanks for taking the time to deep dive into it -

> Ability to replace BRDF implementation for WebGPUPathTracer (a la fully custom material); Maybe it shouldn't be called a material to not confuse with three.js materials?

Yeah I think this may become more clear as we continue to work on things. Eventually we'll want to support multiple BSDFs so we'll need to find a way to allow the user to associate one with their materials. The default ones should be able to be automatically mapped internally, though (eg Standard / PhysicalStandard -> glTF BSDF, BasicMaterial -> simplified / immediately terminated BSDF)

> Although some of the functions and arguments are not needed (or will not be needed?)

What do you mean by this?

> There is a problem with brights fireflies appearing some of the time, so I had to clamp impact of individual ray to 4.0 but I'm not sure they should be there in the first place. It seems we would need to port filter glossy as well.

Yeah fireflies are a pain but clamping is something we should avoid doing. Fireflies can happen particularly when there's a low probability path that happens to hit an particularly bright light after multiple bounces (usually after a diffuse-specular-diffuse bounce pattern). Next Event Estimation can help a bit and I expect bi directional path tracing will help, as well, but I don't think it will get rid of them entirely. The "filterGlossyFactor" was intended to help with this. Basically it artificially increase the roughness of subsequent hit materials based on the roughness of the materials it has hit. Blender includes an option for this reason, as well. See "filter glossy" option [here](https://docs.blender.org/manual/en/2.91/render/cycles/render_settings/light_paths.html) and some more discussion [here](https://docs.blender.org/manual/en/latest/render/cycles/optimizations/reducing_noise.html)

> I think it would be good to fix those issues in glsl as well, in another PR

Happy to get it fixed in the WebGL version, as well

--

I'll have to take some time later to take a look at the code here in more detail but in the mean time I've added a "webgpu_furnace_test" to evaluate energy conservation since it looks like the model had been removed from the gltf sample data. If the BSDF is implemented and blended correctly then the scene should all be the same as the background color. You can see in three.js' furnace demo [here](https://threejs.org/examples/?q=webgl%20fur#webgl_furnace_test). This is what it looks like right now in this PR - my understanding is this can be hard to get exactly right but it may be worth taking a look to see if it can be improved:

image

TheBlek:
> Eventually we'll want to support multiple BSDFs so we'll need to find a way to allow the user to associate one with their materials.

We could reserve some of the bits from material index to signify which BSDF should be used.

Also, with different, complex BSDFs we could explore dividing ProcessHitsKernel into multiple kernels to improve occupancy. Or sort hits by bsdf?

>  Although some of the functions and arguments are not needed (or will not be needed?)

> What do you mean by this?

What I meant is that not every BSDF implementation can fit into the glTF framework. For example, OpenPBR handles coating much differently from what glTF implies: https://academysoftwarefoundation.github.io/OpenPBR/#model/coat. It seems to alter roughness of the base material and emulate darkening explicitly. And that differs from glTF's approach of blending in another specular lobe basically.

> The "filterGlossyFactor" was intended to help with this. Basically it artificially increase the roughness of subsequent hit materials based on the roughness of the materials it has hit. Blender includes an option for this reason, as well. See "filter glossy" option [here](https://docs.blender.org/manual/en/2.91/render/cycles/render_settings/light_paths.html) and some more discussion [here](https://docs.blender.org/manual/en/latest/render/cycles/optimizations/reducing_noise.html)

Yeah. And the blender page on reducing noise is where the idea for clamping samples came from. I agree that we need a filter glossy option, I will look into it a bit later.

I've modified furnace_test to be able to switch between webgl and webgpu versions.

Added energy compensation from FrostBite's [paper](https://seblagarde.wordpress.com/wp-content/uploads/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf) for disney diffuse to conserve energy, here's what it looks like now:
chrome_KQ2Lt8oXUq

I will look into multiple-scatter compensation for GGX next. It should help with high roughness metals (and hopefully dielectrics)

gkjohnson:
> For example, OpenPBR handles coating much differently from what glTF implies: https://academysoftwarefoundation.github.io/OpenPBR/#model/coat.

I understand now - I think we can tackle some of these things as they're requested or as we come to them. Depending on the features it could be that the glTF features are a subset of some of these more complex models. This may not be the case with all parameters you're referring to but in this case it sounds like "coat_darkening" is a [0, 1] that blends in the implicit darkening effect from the coat ior. So in glTF's model that would mean the "darkening" factor is implied to always be 0.

I think the bigger issue is that these parameters are not present on three.js' material objects, which we're using as the medium for reading these parameters. With nodes, though, it may be more possible to add some of these features in on top of the existing materials if we want to make a more complex definition. Problems for later, though, if they're actually needed 😅

> Yeah. And the blender page on reducing noise is where the idea for clamping samples came from. I agree that we need a filter glossy option, I will look into it a bit later.

Oh I didn't realize Blender had that option. Sounds like a last resort in Blender, as well.

gkjohnson:
@TheBlek I may have misunderstood previously - is this ready to be reviewed and merged? I saw you'd made another commit so I wasn't sure. If so I can help clean up the remaining conflicts and errors and we can get this ready. I know getting the furnace test working completely isn't simple so I'd say we should merge this with the current furnace state, which looks pretty good, now!

image

TheBlek:
@gkjohnson No, this is not really ready for review yet. Sorry for the delay here.

I've been trying to add ggx multi-scatter compensation using Turquin'19 method: https://blog.selfshadow.com/publications/turquin/ms_comp_final.pdf. It works by precomputing an energy loss texture for the brdf and then using that in the brdf calculations to compensate. However it seems to be implemented wrong since high roughness metals still show some loss. Dielectrics are not implemented - they would need more careful handling regarding transmission.

There are a couple of other things I want to address before merging as well. So I would like to proceed here as follows:
- [x] Find root cause of overly bright pixels on the edges. They disappear when using sample clamping but this makes algorithm biased, so should be avoided
- [ ] Implement filtered roughness
- [x] Remove sample clamping
- [ ] Remove current implementation of ggx multi-scatter compensation. This is a complex topic that probably could be implemented in a separate PR after transmission implementation. There are a couple of alternatives to the method here that could be considered before committing.

Let me know if you think this should be done differently.

src/webgpu/nodes/material.wgsl.js
```diff
@@ -302,3 +275,187 @@ export const lambertBsdfFunc = wgslFn( /* wgsl */`
...
+
+`, [ ggxSmithVisibilityFunc, ggxDistributionFunc ] );
+
+export const fresnelMixFunc = wgslFn( /* wgslFn */ `
+
+	fn fresnelMix( VdotH: f32, ior: f32, base: vec3f, layer: vec3f ) -> vec3f {
+
+		let f0 = iorToF0( ior );
+  	let F = schlickFresnel( abs( VdotH ), f0 );
+  	return base + F * layer;
```
gkjohnson, line 346: Should this be a "mix" call rather than an additive operation? See the [gltf spec here](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#fresnel).

```wgsl
mix( base, layer, F )
```

src/webgpu/nodes/material.wgsl.js
gkjohnson: Is this supposed to assign to "clearcoatNormal" rather than "normal"?

gkjohnson, line 178: This is supposed to be a mix based on the "g" value of the iridescence texture. See the spec [here](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_iridescence/README.md#properties).

From the glsl version:

```glsl
float iridescenceThicknessSampled = texture2D(
  textures,
  vec3( uvPrime.xy, material.iridescenceThicknessMap )
).g;

iridescenceThickness = mix(
  material.iridescenceThicknessMinimum,
  material.iridescenceThicknessMaximum,
  iridescenceThicknessSampled
);
```

src/webgpu/nodes/material.wgsl.js
```diff
@@ -302,3 +275,187 @@ export const lambertBsdfFunc = wgslFn( /* wgsl */`
...
+	) -> void {
+
+		const INTEGRATION_DIMENSIONS = vec2( 128, 128 );
+
+		let dimensions = textureDimensions( texture ).xy;
+		let uv = ( vec2f( globalId.xy ) + vec2f( 0.5 ) ) / vec2f( dimensions );
+
+		let cosThetaO = uv.x;
+		let roughness = uv.y;
+		let alpha = roughness * roughness;
```
gkjohnson: A bit of a detail: I'm wondering we should just upload the "roughess" as a "roughness^2" (or overwrite the struct on read) since we wind up having to square the roughness everywhere.

src/webgpu/compute/wavefront/RayIntersectionKernel.js
```diff
@@ -131,6 +131,7 @@ export class RayIntersectionKernel extends ComputeKernel {
 						resultColor = sampleEnvironment( background, backgroundSampler, backgroundInfo, input.direction, pcgRand2() );
 
 					}
+					resultColor = clamp( resultColor, vec4( 0.0 ), vec4( 4.0 ) );
```
gkjohnson: Lets remove the clamp for now so the results we're getting are more clear while testing.

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -0,0 +1,161 @@
...
+
+	init( renderer ) {
+
+		const turquinParams = {
+			texture: textureStore( this.turquinTexture ).toWriteOnly(),
+			globalId,
+		};
+		const turquinKernel = new ComputeKernel( albedoIntegralMonteCarlo( turquinParams ), { workgroupSize: [ 16, 16, 1 ] } );
+
+		renderer.compute( turquinKernel.kernel, [ 4, 4, 1 ] );
```
gkjohnson: The above "turquinTexture" is set to 32x32 dimensions but the dispatch here is issues 64x64 threads. Is this intended?

src/webgpu/nodes/ggx.wgsl.js
```diff
@@ -0,0 +1,176 @@
...
+		}
+
+		let theta = acos( clamp( cosTheta, -1.0, 1.0 ) );
+		let tanTheta = tan( theta );
+		let tanTheta2 = pow( tanTheta, 2.0 );
+
+		let denom = PI * cosTheta4 * pow( a2 + tanTheta2, 2.0 );
+		return ( a2 / denom );
+
+	}
```
gkjohnson: I created a debug view of what the "turquin" texture looks like and I'm seeing this (notice the black spots at the bottom of the texture):

Image

Claude has suggested the following replacement for this as a replacement as a simplification of the "ggxDistribution" function, removing divide by zero (or near divide by zero) cases.

```wgsl
var a2 = alpha * alpha;
a2 = max( EPSILON, a2 );

if ( halfVectorAngleCos <= 0.0 ) {

	return 0.0;

}

let denom = halfVectorAngleCos * halfVectorAngleCos * ( a2 - 1.0 ) + 1.0;
return a2 / ( PI * denom * denom );
```

I'm not one to blindly trust AI for stuff like this so this should be validated more concretely but it does result in fixing those gaps in the turquin texture:

Image

src/webgpu/nodes/material.wgsl.js
```diff
@@ -302,3 +275,187 @@ export const lambertBsdfFunc = wgslFn( /* wgsl */`
...
+		let Vis = ggxSmithVisibility( NdotV, NdotL, alpha );
+		let D = ggxDistribution( NdotH, alpha );
+
+		return vec3f( D * Vis );
+
+	}
+
+`, [ ggxSmithVisibilityFunc, ggxDistributionFunc ] );
+
+export const fresnelMixFunc = wgslFn( /* wgslFn */ `
```
gkjohnson: nit: the comment should be `/* wgsl */` rather than `/* wgslFn */`.

I use [this plugin for VSCode](https://marketplace.visualstudio.com/items?itemName=ggsimm.wgsl-literal) which adds inline syntax highlighting for template strings that are tagged with `/* wgsl */`.

gkjohnson (commented):
Thanks - I've just done a bit of a review on a few things I noticed. I'm thinking we can handle some of these comments and get the rest of the non-transparent / transmissive PBR pipeline working (test emission, iridescence, etc) and get this merged. Then we can tackle transmission and dielectric multiscatter in a dedicated PR which will make the discussion more focused and easier to follow.

Regarding pixel brightness, I'm not seeing any overly bright pixels at edges, anymore, when removing the sample value clamp. Is this still an issue?

Image

gkjohnson:
> However it seems to be implemented wrong since high roughness metals still show some loss

Also, regarding this comment, note that because the spheres will be reflecting off of each other, it's possible that one sphere may _look_ like it has more energy loss than is present because it's reflecting a sphere next to it that has more significant energy loss, if that makes sense.

> Find root cause of overly bright pixels on the edges. They disappear when using sample clamping but this makes algorithm biased, so should be avoided

Rereading this, you might be referring to these artifacts? (I thought you were referring to some artifact from fresnel blending). I have accentuated them here by reducing the geometric resolution of the spheres. I suspect this must be related to the discrepancy between the geometric normal and shading normal (described in [this pbr-book link](https://pbr-book.org/3ed-2018/Materials/BSDFs) I'd sent before). Some of the other bright noise may be related to not having any surface offset for the ray when reflecting off a surface. These are things we'll need to address but I don't think this specifically means there's an issue in this implementation.

image

src/webgpu/nodes/material.wgsl.js
TheBlek: I think not. Disney diffuse model already incorporates fresnel term into the calculations and its a bit different from this one. Its `( 1 - F( NdotV ) ) * ( 1 - F( NdotL ) )`

src/webgpu/nodes/material.wgsl.js
TheBlek: Iridescence itself is "r" channel, but thickness is in "g". Fixed that, added a mix

src/webgpu/nodes/material.wgsl.js
TheBlek: I'm not sure about that. Roughness is squared once before material evaluation and alpha is passed as a parameter. Or do you mean turquin texture parametrisation?

src/webgpu/nodes/ggx.wgsl.js
TheBlek: Claude indeed suggested a good ggxDistribution simplification. However that view of the texture is weird (wrong?). When checking through WebGPU Inspector extension I see this:

chrome_NHycfXC7if

Which is consistent with image for ggx distribution from the paper and does not show any black squares. (darkest value is ~0.32)

TheBlek:
> Also, regarding this comment, note that because the spheres will be reflecting off of each other, it's possible that one sphere may look like it has more energy loss than is present because it's reflecting a sphere next to it that has more significant energy loss, if that makes sense.

I understand that and tried to check fully metallic and rough sphere by itself and it still showed some loss.

> Rereading this, you might be referring to these artifacts? (I thought you were referring to some artifact from fresnel blending). I have accentuated them here by reducing the geometric resolution of the spheres.

Yes, sorry for not attaching an image to showcase them. First noticed them on a model, as very bright white highlights. It seems that this was a problem with precision of complex trigonometric formulas. It took simplifying all of them to get rid of those bright highlights. Went back to webgl implementation to check if there were any problems. Turning off nan checks turns a lot of the same pixels that were white in webgpu into black (presumably NaNs). My guess is that trig is handled a little differently in webgl vs webgpu and causes different behavior.

Initially I though filteredRoughness may help here, but it should not, so it could wait for another PR imo.

Also: saved turquin compensation texture into an image file to load. Since its static for a material, we can package the default texture to not calculate it on the device. Let me know if you would like that to be handled differently.

> I'm thinking we can handle some of these comments and get the rest of the non-transparent / transmissive PBR pipeline working (test emission, iridescence, etc) and get this merged. 

To be clear: Do you want other gltf extensions handled in this PR? Like iridescence or clearcoat?

src/webgpu/nodes/ggx.wgsl.js
gkjohnson: Thanks I realized I had the mapping wrong after the fact. Yours looks right!

src/webgpu/nodes/material.wgsl.js
gkjohnson: I'm just seeing it squared in "diffuseBrdfFunc", "albedoIntegralMetallic", "bsdfEvalFunc", "bsdfSample", which seems like a lot of places. Since squaring is really just here to adjust the property to have a more "perceptual" profile of roughness it seems like it would be simplest to do this in one spot at the source and then all functions will naturally be consistently operating on the right roughness factor.

Just a suggestion, though. If you disagree we can keep it as is.

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -0,0 +1,176 @@
...
+import { StorageTexture, RedFormat, LinearFilter, FloatType, TextureLoader } from 'three/webgpu';
+import { wgslTagFn } from '../lib/nodes/WGSLTagFnNode';
+import { PathtracingMaterial } from './PathtracingMaterial';
+import { specularBrdfFunc, diffuseBrdfFunc, fresnelMixFunc, conductorFresnelFunc, albedoIntegralMetallic } from '../nodes/material.wgsl';
+import { diffuseDirectionFunc, getLobeWeightsFunc } from '../nodes/sampling.wgsl';
+import { ggxDirectionFunc, ggxReflectionAdjustedPDFFunc } from '../nodes/ggx.wgsl';
+import { scatterRecordStruct } from '../nodes/structs.wgsl';
+import { pcgRand } from '../nodes/random.wgsl';
+import { ComputeKernel } from '../compute/ComputeKernel';
+import turquinMetal from '../../textures/turquinMetal.png';
```
gkjohnson: This isn't a method we can use for importing textures - it may work for vite but it won't work across the board with other bundlers. A typical method is to use a data url or to use something like this which is _pretty_ reliable across bundlers I believe:

```js
const TURQUIN_URL = new URL( '../../textures/turquinMetal.png', import.meta.url ).toString();
```

src/webgpu/nodes/ggx.wgsl.js
```diff
@@ -0,0 +1,138 @@
...
+export const ggxDistributionFunc = wgslFn( /* wgsl */ `
+	fn ggxDistribution( NdotH: f32, alpha: f32 ) -> f32 {
+
+		let a2 = max( alpha * alpha, EPSILON );
+		let denom = NdotH * NdotH * ( a2 - 1 ) + 1;
+
+		return ( a2 / ( PI * denom * denom ) );
+
+	}
+`, );
```
gkjohnson, line 124: Do we need to add "constants" to the dependencies of this function?

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -0,0 +1,176 @@
...
+				var wh: vec3f;
+
+				if ( r <= cdf.x ) { // diffuse
+
+					wi = diffuseDirection( wo, surf );
+					wh = normalize( wi + wo );
+
+				} else if ( r <= cdf.y ) { // specular
+
+					wh = ggxDirection( wo, vec2( alpha ), pcgRand2() );
```
gkjohnson, line 124: We should add "pcgRand2" to the dependencies here, as well.

src/webgpu/compute/PathTracerMegaKernel.js
```diff
@@ -147,10 +147,18 @@ export class PathTracerMegaKernel extends ComputeKernel {
...
-						let scatterRec = ${ lambertBsdfFunc }( - ray.direction, surface );
+						let scatterRec = ${ material.getBsdfNode() }( - ray.direction, surface );
 
-						// white diffuse surface
-						throughputColor *= scatterRec.color / scatterRec.pdf;
+						if ( scatterRec.pdf <= 0.0 || any( scatterRec.color != scatterRec.color ) ) {
+
+							return;
+
+						}
```
gkjohnson, line 161: I'm thinking this should be "break" rather than "return"? Otherwise we don't increment that we took a sample or save any accumulated emission.

src/webgpu/compute/PathTracerMegaKernel.js
```diff
@@ -147,10 +147,18 @@ export class PathTracerMegaKernel extends ComputeKernel {
...
+
+						resultColor += vec4f( throughputColor * surface.emission, 0.0 );
```
gkjohnson: I'm also thinking this should be moved before the above "invalid scatter" check. Even if the path is ultimately terminated we should still accumulate the emission contribution.

src/webgpu/nodes/material.wgsl.js
gkjohnson: Small catch that may be hard to find later - this should be "clearcoatNormal" rather than "normal"

gkjohnson (commented):
> To be clear: Do you want other gltf extensions handled in this PR? Like iridescence or clearcoat?

Either way! Lets handle these in subsequent PRs since this is just about ready. Very excited to see this 😁

One thing I will note is that I am seeing these artifacts when toggling between wavefront and megakernel path tracers. Do you see it on your machine?

image

--

And just to make a list of remaining PBR features to tackle - is there anything missing here?

- Clearcoat support
- Sheen support
- Transmission support
- Iridescence
- Volume attenuation
- Opacity (handled in #751)
- Alpha cutout
- Anisotropy
- Dielectric / Transmission multiscatter compensation
- Material sided-ness

src/webgpu/nodes/material.wgsl.js
TheBlek: Good catch!

TheBlek:
> One thing I will note is that I am seeing these artifacts when toggling between wavefront and megakernel path tracers. Do you see it on your machine?

Yes. That is weird, looks like there are some problems with low res samples and/or problem with blending them. Looking into it.

gkjohnson:
> Yes. That is weird, looks like there are some problems with low res samples and/or problem with blending them. Looking into it.

It looks like this issue starts to happen in 8fed465. I suspect it's because the turquin texture loads asynchronously but is not awaited, so it's being used before it's fully initialized, resulting in some invalid calculations.

Why don't we just load the texture once in an awaited global variable so we can ensure it's available and then we can revisit the swappable material design later on to be more resilient to these cases.

TheBlek:
> Why don't we just load the texture once in an awaited global variable so we can ensure it's available and then we can revisit the swappable material design later on to be more resilient to these cases.

Did that.

While trying to test with a debug material I noticed that the material change for a pathtracer does not work properly. Fixed that. I think it could be helpful to expose an example with editable debug texture. So you could inspect different values. Maybe even inside the viewerTest example?

Another thing I noticed is normal map seems to have less effect in webgpu.

WebGL:
chrome_UjVw2MKlVS

WebGPU:
chrome_mWn8YwuhOs

TheBlek:
Normal map issue turned out to be missing tangent attribute on the mesh. Added an option to generate common attributes as in StaticGeometryGenerator.

TheBlek:
> And just to make a list of remaining PBR features to tackle - is there anything missing here?
> - Clearcoat support
> - Sheen support
> - Transmission support
> - Iridescence
> - Volume attenuation
> - Opacity (handled in https://github.com/gkjohnson/three-gpu-pathtracer/pull/751)
> - Alpha cutout
> - Anisotropy
> - Dielectric / Transmission multiscatter compensation
> - Material sided-ness

Thin materials would also need special handling afaik

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -0,0 +1,179 @@
...
+import { PathtracingMaterial } from './PathtracingMaterial';
+import { specularBrdfFunc, diffuseBrdfFunc, fresnelMixFunc, conductorFresnelFunc, albedoIntegralMetallic } from '../nodes/material.wgsl';
+import { diffuseDirectionFunc, getLobeWeightsFunc } from '../nodes/sampling.wgsl';
+import { ggxDirectionFunc, ggxReflectionAdjustedPDFFunc } from '../nodes/ggx.wgsl';
+import { scatterRecordStruct } from '../nodes/structs.wgsl';
+import { pcgRand, pcgRand2 } from '../nodes/random.wgsl';
+import { ComputeKernel } from '../compute/ComputeKernel';
+
+const TURQUIN_METAL_URL = new URL( '../../textures/turquinMetal.png', import.meta.url ).toString();
+const TURQUIN_METAL_TEXTURE = await new TextureLoader().load( TURQUIN_METAL_URL );
```
gkjohnson, line 13: "load" returns a Texture instance so await won't do anything here. We should use "loadAsync" if we want to await a promise.

gkjohnson:
> I think it could be helpful to expose an example with editable debug texture. So you could inspect different values. Maybe even inside the viewerTest example?

What do you mean by "inspect different values"? Do you mean like hovering over a pixel to see intermediate calculations? Either way we will need a better few test scenes. We can probably start converting some of the other examples to use WebGPUPathtracer if it's helpful, as well.

TheBlek:
> What do you mean by "inspect different values"?

I mean having some debug overlays that would display intermediate values as color. This is a limited view that would allow to see only one value at a time.

> Do you mean like hovering over a pixel to see intermediate calculations?

That would be very useful. Do you know how that could be achieved? Only thing I can think of is writing intermediate values into storage textures but there is only so many texture binding slots.

Another way to trace calculations is to debug shaders in native tools like [RenderDoc](https://toji.dev/webgpu-profiling/renderdoc) or just [Metal Debugger](https://toji.dev/webgpu-profiling/xcode.html). So maybe we should not resort to very complex means of tracing if such debuggers useful enough. 

I tried debugging with RenderDoc myself but faced problems emitting hlsl shaders without symbol renaming. Although even without that translated shaders are recognizable and debuggable. And buffer contents are inspectable which is especially useful in wavefront case.

Note: useful flags for debugging in chrome: https://dawn.googlesource.com/dawn/+/HEAD/docs/dawn/debugging.md

gkjohnson:
> That would be very useful. Do you know how that could be achieved? Only thing I can think of is writing intermediate values into storage textures but there is only so many texture binding slots.

Yeah this or storage buffers more or less. The good thing is that we can use any extra slots allowed by our hardware since this is for debugging only rather than being limited by the default WebGPU limitations.

> I tried debugging with RenderDoc myself but faced problems emitting hlsl shaders without symbol renaming.

The new three.js inspector may have utilities for stuff like this but I'm less familiar with it and it's unclear how well it works outside the traditional rendering paths.

--

And again - very nice work on the PR. I'll go ahead and merge this along with some of the other opacity-related ones.

gkjohnson:
One thing I've noticed is that are a lot of black pixels showing up when using the wave front path - I'm think it might be from NaNs? Not sure if this is something you might have some ideas about?

### **Megakernel**

image

### **Wavefront**

image

TheBlek:
> lot of black pixels showing up when using the wave front path

Yes, I've also noticed that when working on clearcoat. The problem was that wavefront path tracer did not have a proper guard against scatter records with pdf <= 0. Fixed in #758

PR #745, opened 2026-03-04
WebGPUPathTracer: Get megakernel working in Firefox, Safari

gkjohnson:
"texture_storage_2d" nodes are not supported as function arguments so they've been replaced here with the wgslTagFn injection approach. If this looks okay I can adjust the remainder of the includes to use the template tag variable injection.

- Adjust WgslTagFn to support injecting storage texture nodes
- Adjust nodeProxy to be resilient to the type of checking that wgslTagFn does
- Remove whitespace stripping to make shader warnings, output readable

cc @theblek


TheBlek (approved):

gkjohnson:
When I get a chance I'll make another PR with the remainder of the shader converted along with the wave front version.

PR #746, opened 2026-03-06
WebGPUPathTracer: Improve build performance, fix failures when rendering an empty scene

gkjohnson:
1. Improve the performance to BVHComputeData.update by caching "attributeStruct.getLength()".
2. ensure buffers are declared to have at least 2 values in order to prevent TSL from converting them to a scalar value and breaking our shader code.
3. Ensure bvh bounds storage buffer is not contain "Infinity"

cc @theblek - is there already an issue submitted for point 2?

### DIFF
--- a/src/webgpu/lib/BVHComputeData.js
+++ b/src/webgpu/lib/BVHComputeData.js
@@ -567,7 +567,7 @@ export class BVHComputeData {
 					const n16 = n32 * 2;
 
 					// write bounds
-					targetF32.set( new Float32Array( 6 ), n32 );
+					targetF32.set( new Float32Array( root, i * BYTES_PER_NODE, 6 ), n32 );
 
 					const isLeaf = IS_LEAFNODE_FLAG === rootBuffer16[ r16 + 15 ];
 					if ( isLeaf ) {
@@ -658,7 +658,7 @@ export class BVHComputeData {
 			const { geometry, mesh = null } = bvh;
 			const { vertexStart, vertexCount } = range;
 			const attributesBufferF32 = new Float32Array( target );
-			const stepSize = attributeStruct.getLength();
+			const attrStructLength = attributeStruct.getLength();
 			attributeStruct.membersLayout.forEach( ( { name }, interleavedOffset ) => {
 
 				// TODO: we should be able to have access to memory layout offsets here via the struct
@@ -701,7 +701,7 @@ export class BVHComputeData {
 
 					}
 
-					_vec.toArray( attributesBufferF32, ( writeOffset + i ) * stepSize + interleavedOffset * 4 );
+					_vec.toArray( attributesBufferF32, ( writeOffset + i ) * attrStructLength + interleavedOffset * 4 );
 
 				}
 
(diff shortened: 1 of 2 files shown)


gkjohnson:
With this changes this crab model now takes ~170ms to "update" the compute data down from ~3500ms (!!).

image

One other issue I ran into with this model is that it seems it's nearing the max size of what we can store in our storage buffers with positions, colors, normals, uvs, and tangents. This model has 1,070,805 vertices and 6,424,818 indices. That's ~85 MB and ~25MB respectively for buffers that have ~128MB limit each. And the BVH Buffer is at ~60MB. If we call "toNonIndexed" before render we get to 6,424,818 vertices or around ~513MB of data. This is a particularly geometrically-dense model but even so it would be nice to be able to render more than this.

It seems we _can_ [request larger storage buffers](https://webgpufundamentals.org/webgpu/lessons/webgpu-limits-and-features.html) from the API but it's very device-specific. My mac laptop can request up to a 4GB storage buffer view but my Pixel 10 phone is stuck at a maximum limit of 128MB for storage buffer attachments while still having a max buffer size of 2GB.

Some possible solutions:
- Restrict the end user by their device limits and detect compatibility of a model ahead of time.
- Switch to using textures or texture arrays for large-scale buffers, which have larger base limits (8K, or ~1GB, guaranteed resolution for 2d textures)
- Reduce bit precision (difficult with no guaranteed i8, i16, etc support)

TheBlek (approved):

TheBlek:
> is there already an issue submitted for point 2?

I did not create one. Will you make it?

TheBlek:
> One other issue I ran into with this model is that it seems it's nearing the max size of what we can store in our storage buffers with positions, colors, normals, uvs, and tangents.

Since we have a TLAS, could we possibly divide the data between two/three buffers? It could help in megakernel case, but wavefront is already at the limit of 8 buffers in triangle intersection kernel.

Textures seem to be a good option to expose a lot more data, but they should be slower to read from because of all the sampling and possible compression hardware. I think we anyway should detect scenes that are too large and make it possible for user to know about this. Otherwise fallback to textures.

I don't see any way to specify device limits for the renderer in three.js, so we can't handle that automatically for now. Should this be another three.js issue?

gkjohnson:
> I did not create one. Will you make it?

Just made it here: https://github.com/mrdoob/three.js/issues/33145

> Since we have a TLAS, could we possibly divide the data between two/three buffers?
> ...
> Textures seem to be a good option to expose a lot more data, but they should be slower to read from because

Separating the TLAS is one option - but it's also likely to be very shallow / small in a lot of cases. I'm also curious how much slower, if at all, textures would be if we use data textures to store data and access the pixel values with textureLoad, skipping any compression and sampling interpolation. Something to investigate for another day but I expect we're going to see this more an more especially as we start using other uv coordinates, as well.

> I don't see any way to specify device limits for the renderer in three.js, so we can't handle that automatically for now. Should this be another three.js issue?

Yeah this probably needs an issue, as well 😅

gkjohnson:
> wavefront is already at the limit of 8 buffers in triangle intersection kernel.

One benefit of using textures for geometry data is that it will free up more storage slots for us. I also wonder if it's possible to provide multiple "views" of the same bound storage buffer in a single shader and gain a storage slot? Eg can we merge materials and transforms into a single storage buffer attribute with materials packed at the end with two different accessors and have this count as a single bound storage buffer?

PR #748, opened 2026-03-10
WebGPUPathTracer: Convert kernels to use wgsl tag fn

gkjohnson:
Related to #745

- Convert all wavefront kernels to use wgslTagFn, get it working in Firefox, Safari
- Normalize some parameter names between kernel definitions

Somewhat related: I've noticed that there are issues with mobile rendering, as well, since some devices don't support linear interpolation of float textures. We may need to fall back to non-interpolated sampling or investigate use of Half Float textures (though I had noticed some precision artifacts when using these in webgl).

cc @theblek


TheBlek (approved):

gkjohnson:
@TheBlek Let me know if you'd like me to help get the conflicts resolved in #717 after this merge 🙏

Issue #133, opened 2022-05-04
Investigate floating point averaging bias


gkjohnson:
https://twitter.com/BenHouston3D/status/1521962488807444481?t=gOGmOLpXPF55D4ihTdc9Ig&s=19

See banding in BG gradients

munrocket:
If this is related to RGBA floating point packing, I think I have an idea how to solve this floating point precision issue.

gkjohnson:
The issue is floating point precision and error resulting in artifacts like banding in gradients when performing math operations.

[This three.js SSAA demo](https://threejs.org/examples/?q=ssaa#webgl_postprocessing_ssaa) demonstrates the same kind of issue but with a fixed number of samples. The difference is that the pathtracer is performing a continuous average instead of fixed sample count.

richardassar:
I think Kahan summation can help with this. The thought came to me the other day but is also noted [here](https://twitter.com/createthematrix/status/1290209687174959105).

gkjohnson:
@richardassar - I'm coming back around to this now that a WebGPU version is coming up. Looking into Kahan summation a bit more it sounds like it requires keeping track of lost precision in a separate buffer so it can be added back later, is that right? So in this case we'd need a second 16 bit buffer to track the losses meaning there wouldn't be any memory savings but it would let us work around the lack of floating-point buffer interpolation (or lack of support for float textures entirely) on mobile devices.

PR #749, opened 2026-03-11
WebGPUPathTracer: Use half float storage target w/ Kahan summation

gkjohnson:
Relies on #748 
Related to #133 

- Changes the output buffers to use HalfFloatType to ensure support on mobile and WebGPU in general (float type targets are not guaranteed).
- Half float targets are also always filterable.
- Use Kahan summation to avoid accumulation of error:
  - Quantize the blended color to F16
  - Calculate the delta between the F16 and F32 value
  - Pack the delta into an 8 bit value for each channel
  - Write those 8 bit values into a read/write r32 texture via bitwise ops
  - Read and unpack the delta on subsequent blend to prevent error accumulation
- Add an ssl option to the start command to allow for testing on mobile (WebGPU requires https) - run `npm start -- --mode=ssl --host`
- Adjust the ProceduralEquirectTexture to use HalfFloatType to guarantee its filterable.

With these changes both the wavefront and megakernel path tracers work on my phone 🎉

image

cc @theblek

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "@types/node": "^24.3.1",
         "@types/three": "^0.183.0",
         "@typescript-eslint/eslint-plugin": "^8.40.0",
+        "@vitejs/plugin-basic-ssl": "^2.1.4",
         "canvas-capture": "^2.0.5",
         "eslint": "^8.56.0",
         "eslint-config-mdcs": "^5.0.0",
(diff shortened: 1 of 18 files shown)


gkjohnson:
I'm a bit worried about precision loss on this - it looks okay in this simple scene and simple BRDF but I expect with enough samples and with high enough values we'd start to see some degradation. The primary issue is that float textures are not guaranteed to be filterable but they do seem to always be available in WebGPU (outside of compatibility mode). It may be best to implement custom filtering when writing to the canvas if filtering isn't available.

gkjohnson:
Closing in favor of #750

PR #751, opened 2026-03-22
WebGPUPathTracer: Add support for opacity during BVH traversal

gkjohnson:
- Add a 1D PCG random function
- Merge the "hit" and "ray" queues to free up a storage buffer slot
- Check the opacity of the object material on triangle intersection, continue traversal if transparent
- Skip traversal of the object entirely if opacity is set to 0

This should reduce the number of raycasts required for transparent surfaces.

**Future**
- Sample and include albedo and alpha map opacity
- We could handle triangle sided-ness this way, as well?

cc @theBlek

image


src/webgpu/nodes/PathtracerBVHComputeData.js
```diff
@@ -35,11 +38,177 @@ export class PathtracerBVHComputeData extends BVHComputeData {
...
+
+					let invDir = 1.0 / ray.direction;
+					let tMinPlane = ( boundsMin - ray.origin ) * invDir;
+					let tMaxPlane = ( boundsMax - ray.origin ) * invDir;
+
+					let tMinHit = vec3f(
+						min( tMinPlane.x, tMaxPlane.x ),
+						min( tMinPlane.y, tMaxPlane.y ),
+						min( tMinPlane.z, tMaxPlane.z )
+					);
```
TheBlek: fyi: min supports vector operations so this can be written as `let tMinHit = min( tMinPlane, tMaxPlane );`

TheBlek (approved):

src/webgpu/nodes/PathtracerBVHComputeData.js
gkjohnson: Thanks! Fixed. I'll wait to merge this PR until #717 is done to avoid and potential conflicts

PR #755, opened 2026-04-22
WebGPUPathTracer: Respect material sidedness, alpha test

gkjohnson:
Depends on #752

- Add respect for material sidedness during BVH traversal
- Add support for alpha test values during BVH traversal


src/webgpu/nodes/PathtracerBVHComputeData.js
```diff
@@ -131,11 +131,25 @@ export class PathtracerBVHComputeData extends BVHComputeData {
...
 
-								let opacity = bvh_baseOpacity;
+							// TODO: if material is a transmissive volume we may need to assume double-sidedness
+							if ( material.side != 0 && triResult.side != material.side ) {
+
+								continue;
+
+							}
+
+							if ( material.transparent != 0 || material.alphaTest > 0.0 ) {
```
TheBlek, line 136: `material.transparent` and `alphaTest` seem to be separate checks. Is the reasoning here to try to avoid opacity lookup that could be expensive in the future?

src/webgpu/nodes/PathtracerBVHComputeData.js
```diff
@@ -131,11 +131,25 @@ export class PathtracerBVHComputeData extends BVHComputeData {
 						if ( triResult.didHit && ( ! result.didHit || triResult.dist < result.dist ) ) {
 
 							let material = bvh_material;
-							if ( material.transparent != 0 ) {
 
-								let opacity = bvh_baseOpacity;
+							// TODO: if material is a transmissive volume we may need to assume double-sidedness
+							if ( material.side != 0 && triResult.side != material.side ) {
```
TheBlek, line 130: Just a thought for the future: What if we encoded sidedness like a bit-mask with 0th bit representing front side and 1st - back side? Then this hot path check could be reduced to only one logical check: `triResult.side & material.side != 0`

Regarding transmission: it may be beneficial to write side as double-sided for transmissive materials at the stage of serialization into the gpu buffer to not introduce additional checks.

TheBlek (approved):

src/webgpu/nodes/PathtracerBVHComputeData.js
gkjohnson: Yeah - we'll need to sample the alpha map and diffuse map for alpha values here at some point so I figured it would be best to consolidate them.

src/webgpu/nodes/PathtracerBVHComputeData.js
gkjohnson: It's something to consider. Right now there's some convenience in multiplying side * normal to get the normal for the side we've actually hit. That can obviously be changed but I'm not sure if I'm feeling one way is necessarily better than the other immediately?

TheBlek: Oh, right, side is { -1, 0, 1 }. That could also lend itself to some bit tricks in the future. For some reason I thought is was 0, 1, 2 😅

PR #757, opened 2026-04-23
WebGPUPathTracer: Add support for texture wrap settings, nearest filtering

gkjohnson:
Encodes the wrapS, wrapT, and magFilter settings in the material struct "map" fields then apply them when sampling the texture data.

"NearestFilter" sampling will probably not look exactly right until we pack the textures into an atlas-style structure since they're currently stretched to meat the 3d texture map resolution. Tested with the "khronos-TextureSettingsTest" test model:

**Before**
image

**After**
image

cc @theBlek


src/webgpu/nodes/PathtracerBVHComputeData.js
```diff
@@ -214,24 +214,46 @@ export class PathtracerBVHComputeData extends BVHComputeData {
...
+					textures.push( texture );
 
 				}
 
-				return textureLookUp.get( hash );
+				const idx = textureLookUp.get( hash );
+				const wrapS = encodeTextureWrap( texture.wrapS );
+				const wrapT = encodeTextureWrap( texture.wrapT );
+				const nearest = texture.magFilter === NearestFilter ? 1 : 0;
+				return ( nearest << 28 ) | ( wrapT << 26 ) | ( wrapS << 24 ) | idx;
```
TheBlek, line 252: Perhaps it would be good to document this bit-packing somewhere?

TheBlek (approved):

src/webgpu/nodes/PathtracerBVHComputeData.js
gkjohnson: Any thoughts on where? Maybe inline in the Javascript-side packing function?

TheBlek: I think its better placed beside the struct definition. By the time one is reading javascript function for packing, layout becomes pretty clear.

PR #758, opened 2026-04-24
WebGPUPathtracer: Clearcoat

TheBlek:
@gkjohnson
This PR adds support for KHR_materials_clearcoat extension:
image
- GltfCompliantMaterial extension with an additional specular lobe. It might be beneficial to use a separate function for this as, for example, disney principled brdf uses different specular lobes for clearcoat and base material.
- NdotL (and LdotNc) are now clamped to EPSILON as that was causing issues at zero roughness. Or should the roughness be clamped instead?
- Ray traversal seems to be weird sometimes and causes fireflies here and there. In simple sphere example, clamping bounces to 2 fixes them. But no ray should really be bouncing more than 2 times in that scene. Also they are fixed by offsetting ray in the scatter direction by 0.1, but it seems like a large offset, doesn't it? Fixed most of them for now with wo.z < 0.0 check.
- baseNormal instead of faceNormal for normal mapping. Using face normal caused normal maps to be flat-shaded even on non-flat-shaded objects. However, that causes AntiqueCamera example to path trace very slowly. It seems to be because bitangent starts containing NaN after normalization and poisons everything. I was hesitant to just add a NaN check for normal as that would hide the problem from the user. What do you think? Maybe you have any guesses as to why that may be happening?


AntonPalmqvist:
Re: roughness clamping, here's how a few other engines handles it, for reference:
- Filament clamps at 0.089 https://google.github.io/filament/Filament.md.html#materialsystem/parameterization/remapping
- Frostbite clamps at 0.045 (see footnote in the link above)
- Unreal seems to clamp at 0.001 (not sure if used everywhere though) https://github.com/EpicGames/UnrealEngine/blob/ue5-main/Engine/Shaders/Private/ShadingCommon.ush#L184

gkjohnson:
> Or should the roughness be clamped instead?

If other engines are clamping the roughness then why don't we do the same (thanks @AntonPalmqvist)

> Ray traversal seems to be weird sometimes and causes fireflies here and there. In simple sphere example, clamping bounces to 2 fixes them. But no ray should really be bouncing more than 2 times in that scene. Also they are fixed by offsetting ray in the scatter direction by 0.1, but it seems like a large offset, doesn't it?

Are you referring to the "webgpu_primitives" scene? 0.1 does seem very high - but I wonder this is less fixing things and more hiding the problem. I think we should make a dedicate PR to discuss this offset since I don't think it's as simple as just adding a fixed value.

> baseNormal instead of faceNormal for normal mapping

The vertex normals ("baseNormal" here) should be the correct ones to use. 

> However, that causes AntiqueCamera example to path trace very slowly. It seems to be because bitangent starts containing NaN after normalization and poisons everything. 

We'll probably need to add some checks somewhere but this should still be understood, I think. The only way I can see this happening is if the bitangent is `0,0,0` or NaN to begin, which mean the normal or tangent source values are-ill formed. I'm wondering if the MikkT generation is resulting in some bad values somehow?

After taking a quick look at the geometry after the tangents are generated I am seeing that there are a few vertices with tangents of length "0" as well as quite a few triangles with an area of 0.

I'll take a deeper look at this in a bit but here are some before and after screenshots:

| Before | After |
|---|---|
| image | image |
| image | image |

src/webgpu/nodes/material.wgsl.js
```diff
@@ -238,9 +238,9 @@ export const getSurfaceRecordFunc = wgslFn( /* wgsl */ `
 		surf.specularColor = specularColor;
 		surf.specularIntensity = specularIntensity;
 
-		surf.roughness = roughness;
-		surf.clearcoatRoughness = clearcoatRoughness;
-		surf.sheenRoughness = sheenRoughness;
+		surf.roughness = clamp( roughness, 0.001, 1.0 );
+		surf.clearcoatRoughness = clamp( clearcoatRoughness, 0.001, 1.0 );
+		surf.sheenRoughness = clamp( sheenRoughness, 0.001, 1.0 );
```
gkjohnson: Should we use an constant epsilon value here?

src/webgpu/nodes/material.wgsl.js
```diff
@@ -366,6 +366,19 @@ export const conductorFresnelFunc = ( turquinTexture ) => wgslFn( /* wgsl */ `
...
+export const fresnelCoatFunc = wgslFn( /* wgsl */ `
+
+	fn fresnelCoat( VdotNc: f32, ior: f32, base: vec3f, layer: vec3f, weight: f32 ) -> vec3f {
+
+		let f0 = iorToF0( ior );
+		let F = schlickFresnel( abs( VdotNc ), f0 );
+
+		return mix( base, layer, weight * F );
+
+	}
```
gkjohnson, line 380: What's the functional difference between this "fresnelCoat" function and "fresnelMixFunc"? It looks like one uses "mix" and one doesn't? Is it possible to unify these easily?

src/webgpu/nodes/utils.wgsl.js
```diff
@@ -1,4 +1,5 @@
 import { wgslFn } from 'three/tsl';
+import { scatterRecordStruct } from './structs.wgsl';
```
gkjohnson: nit: we should include the file extension

src/webgpu/compute/PathTracerMegaKernel.js
```diff
@@ -154,7 +155,7 @@ export class PathTracerMegaKernel extends ComputeKernel {
 
 						let scatterRec = ${ bsdfSampleFn }( - ray.direction, surface );
 
-						if ( scatterRec.pdf <= 0.0 || any( scatterRec.color != scatterRec.color ) ) {
```
gkjohnson, line 157: Is this "any( scatterRec.color != scatterRec.color )" line something we should deliberately drop? I assume this was a NaN check?

src/webgpu/nodes/sampling.wgsl.js
```diff
@@ -75,20 +75,25 @@ export const getLobeWeightsFunc = wgslFn( /* wgsl */ `
 		var weights: LobeWeights;
 		weights.diffuse = ( 1.0 - transmission ) * ( 1.0 - diffSpecularProb );
 		weights.specular = transmission * transSpecularProb + ( 1.0 - transmission ) * diffSpecularProb;
+
+		let clearcoatF0 = iorToF0( clearcoatIor );
+		let clearcoatFresnel = schlickFresnel( saturate( woClearcoat.z ), clearcoatF0 );
```
gkjohnson, line 80: It looks like we need to pass these iorToF0 and schlickFresnel functions into the `wgslFn` dependencies.

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -1,17 +1,20 @@
...
-import { scatterRecordStruct } from '../nodes/structs.wgsl';
+import { scatterRecordStruct, surfaceRecordStruct } from '../nodes/structs.wgsl';
 import { pcgRand, pcgRand2 } from '../nodes/random.wgsl';
 import { ComputeKernel } from '../compute/ComputeKernel';
 
 const TURQUIN_METAL_URL = new URL( '../../textures/turquinMetal.png', import.meta.url ).toString();
 const TURQUIN_METAL_TEXTURE = await new TextureLoader().loadAsync( TURQUIN_METAL_URL );
 
+const CLEARCOAT_IOR = float( 1.5 );
+const EPSILON = float( 1e-3 );
```
gkjohnson: Small nit: it might be worth naming this something a bit more explicit? Something like COS_EPSILON or NEAR_ZERO_EPSILON or something?

src/webgpu/nodes/material.wgsl.js
```diff
@@ -366,6 +366,19 @@ export const conductorFresnelFunc = ( turquinTexture ) => wgslFn( /* wgsl */ `
 
 `, [ schlickFresnelVecFunc, turquinTexture ] );
 
+export const fresnelCoatFunc = wgslFn( /* wgsl */ `
+
+	fn fresnelCoat( VdotNc: f32, ior: f32, base: vec3f, layer: vec3f, weight: f32 ) -> vec3f {
```
gkjohnson, line 373: It looks like the "fresnelMixFunc" function takes cosign between the half vector and the view vector rather than the view and normal vectors taken here:

```
fn fresnelMix( VdotH: f32, ior: f32, base: vec3f, layer: vec3f )
```

Should this function be taking something like `VdotHc`, instead?

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -74,105 +79,147 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 
 		const bsdfEvalFunc = wgslTagFn/* wgsl */`
 
-			fn bsdfEval( NdotL: f32, NdotV: f32, NdotH: f32, VdotH: f32, surf: SurfaceRecord ) -> vec3f {
+			fn bsdfEval(
+				NdotL: f32, NdotV: f32, NdotH: f32, VdotH: f32,
+				LdotNc: f32, VdotNc: f32, HdotNc: f32,
+				surf: SurfaceRecord
```
gkjohnson: nit: we should use the "SurfaceRecordStruct" as a dependency here:

```js
surf: ${ surfaceRecordStruct }
```

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -74,105 +79,147 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 
 		const bsdfEvalFunc = wgslTagFn/* wgsl */`
 
-			fn bsdfEval( NdotL: f32, NdotV: f32, NdotH: f32, VdotH: f32, surf: SurfaceRecord ) -> vec3f {
+			fn bsdfEval(
+				NdotL: f32, NdotV: f32, NdotH: f32, VdotH: f32,
+				LdotNc: f32, VdotNc: f32, HdotNc: f32,
```
gkjohnson: I'm wondering if we'd want to consolidate these intermediate values into a single kind of "Record" struct at some point to keep the function signatures more clear, though a clear name isn't coming to me immediately.

src/webgpu/nodes/material.wgsl.js
TheBlek: Official sample gltf implementation states that VdotNc is used to make material energy conserving here:

> In order to make the material energy conserving with a simple layering function, we compute the microfacet Fresnel term with NdotV instead of VdotH. That means that we ignore the orientation of the microsurface. As the clearcoat roughness is usually very low the microfacets orientation is very close to the normal direction, and NdotV ≈ NdotL.

Though I too find this weird and not intuitive.

src/webgpu/compute/PathTracerMegaKernel.js
TheBlek: This is a NaN check, yes. It will prevent pixels from becoming NaN because of invalid material color evaluation. Which is good for a user, but for developers that would hide that something wrong is going on under the hood. Same as NaN checks for normals in normal mapping. If we had an option, I would say this should be disabled in some kind of debug(fast?) mode. We could just conditionally add code once at the creating time depending on the constant defined in a file. What do you think? Or should we just stick with a safe option?

src/webgpu/materials/GltfCompliantMaterial.js
TheBlek: Unreal Engine code seems to have BxDFContext struct that incorporates all that [link](https://github.com/EpicGames/UnrealEngine/blob/ue5-main/Engine/Shaders/Private/BRDF.ush):

```
struct BxDFContext
{
	half NoV;
	half NoL;
	half VoL;
	half NoH;
	half VoH;
	half XoV;
	half XoL;
	half XoH;
	half YoV;
	half YoL;
	half YoH;
};
```

Seems good to me.

src/webgpu/nodes/material.wgsl.js
TheBlek: Added MIN_ROUGHNESS constant

src/webgpu/nodes/material.wgsl.js
TheBlek: I can't think of a way. They are supposed to have different API according to gltf spec and perform mixing differently. For example, base is unaffected in fresnel_mix.

```
dielectric_brdf =
  fresnel_mix(
    ior = 1.5,
    base = diffuse_brdf(
      color = baseColor),
    layer = specular_brdf(
      α = roughness ^ 2))
```

```
# layering
coated_material =
  fresnel_coat(
    normal = clearcoatNormal,
    ior = 1.5,
    weight = clearcoat,
    base = material,
    layer = clearcoat_brdf)
```

src/webgpu/materials/GltfCompliantMaterial.js
TheBlek: Renamed to MIN_INCIDENT_COS

src/webgpu/materials/GltfCompliantMaterial.js
gkjohnson: Looks good to me - are Xo* and Yo* fields for anisotropy?

src/webgpu/nodes/material.wgsl.js
gkjohnson: Interesting - it might just be worth adding a comment stating this, then

src/webgpu/compute/PathTracerMegaKernel.js
gkjohnson: > Which is good for a user, but for developers that would hide that something wrong is going on under the hood.

Agreed - lets keep it removed for now and we can decide whether it's actually needed later. Ideally we'd be addressing the underlying issues rather than using a check like this.

src/webgpu/materials/GltfCompliantMaterial.js
TheBlek: Yep
```
// [Heitz 2014, "Understanding the Masking-Shadowing Function in Microfacet-Based BRDFs"]
float Vis_SmithJointAniso(float ax, float ay, float NoV, float NoL, float XoV, float XoL, float YoV, float YoL)
{
	float Vis_SmithV = NoL * length(float3(ax * XoV, ay * YoV, NoV));
	float Vis_SmithL = NoV * length(float3(ax * XoL, ay * YoL, NoL));
	return 0.5 * rcp(Vis_SmithV + Vis_SmithL);
}
```

src/webgpu/nodes/structs.wgsl.js
```diff
@@ -214,3 +204,15 @@ export const lobeWeightsStruct = new StructTypeNode( {
...
+
+export const bxdfContextStruct = new StructTypeNode( {
+	NdotV: 'float',
+	NdotL: 'float',
+	NdotH: 'float',
+	VdotH: 'float',
+
+	VdotNc: 'float',
+	LdotNc: 'float',
+	HdotNc: 'float',
```
gkjohnson, line 216: I think my only feedback here is it's odd that the vector ordering on the variables for the clearcoat vs albedo products are in different orders. Eg `NdotV` vs `VdotNc` - why not `NdotVc`?

I'll merge for now but something to consider for a future changes unless there's a reason I'm missing.

PR #759, opened 2026-04-25
Add support for component from maps

gkjohnson:
Testing with the "khronos-AlphaBlendModeTest" model:

| Before | After |
|---|---|
| image | image |

- Sample the alpha contribution during BVH traversal
- Adjust the render target 2d array generation to use a mesh basic material skip any alpha premultiplication

Note the UVs are generated separately since we'll need to account for the use of different UV channels in the future.

cc @theblek

### DIFF
--- a/src/webgpu/nodes/PathtracerBVHComputeData.js
+++ b/src/webgpu/nodes/PathtracerBVHComputeData.js
@@ -149,8 +149,9 @@ export class PathtracerBVHComputeData extends BVHComputeData {
 									let b = ${ storage.attributes }[ i1 ].uv.xy;
 									let c = ${ storage.attributes }[ i2 ].uv.xy;
 									let uv = barycoord.x * a + barycoord.y * b + barycoord.z * c;
+									let uvPrime = material.mapTransform * vec3f( uv, 1 );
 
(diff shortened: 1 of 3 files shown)


src/webgpu/nodes/PathtracerBVHComputeData.js
```diff
@@ -135,8 +139,34 @@ export class PathtracerBVHComputeData extends BVHComputeData {
...
+								var opacity = ${ baseOpacityScalar };
+
+								// account for alpha component of albedo map
+								if ( material.map != - 1 ) {
+
+									let barycoord = triResult.barycoord;
+									let a = ${ storage.attributes }[ i0 ].uv.xy;
+									let b = ${ storage.attributes }[ i1 ].uv.xy;
+									let c = ${ storage.attributes }[ i2 ].uv.xy;
+									let uv = barycoord.x * a + barycoord.y * b + barycoord.z * c;
```
TheBlek, line 151: I believe uvs should be multiplied by `mapTransform` and `alphaMapTransform` matricies from the object's material.

gkjohnson: Yes, thank you! Just added

PR #760, opened 2026-04-27
WebGPU: viewerTest example improvements

TheBlek:
@gkjohnson
Made some changes/improvement to viewerTest example to help compare performance between implementations:
- Added samplesPerFrame variable to try computing multiple samples on one requestAnimationFrame. Because of tiling, sample calculation is often pretty fast, so I think there should be an option to calculate multiple samples. It seems like an easy enough detail to leave to the user. Ideally, we would have a way to automatically determine this number to maximize gpu usage. What do you think? Should this be part of the pathtracer API?
- Fixed `tiles` not setting correctly for webgl pathtracer and webgpu/megakernel
- Implemented getRenderTime and getDetailedSampleCount for webgl pathtracer. Detailed sample counts from are difficult to get here as they are not saved or calculated anywhere. I've tried an approach with keeping 1x1 texture with sample count value and increment it after each sample - it turned out to differ only by a couple of frames. This I think is acceptable price for not increasing complexity of existing system. Or should sample counts be calculated precisely here too?

### DIFF
--- a/example/viewerTest.js
+++ b/example/viewerTest.js
@@ -47,6 +47,8 @@ const params = {
 	tiles: tiles,
 	scale: scale,
 
+	samplesPerFrame: 1,
+
 	model: '',
 	checkerboardTransparency: true,
 	displayImage: false,
@@ -210,7 +212,11 @@ function animate() {
 		pathTracer.enablePathTracing = params.enable;
 		pathTracer.pausePathTracing = params.pause || pathTracer.samples > maxSamples && maxSamples !== - 1;
 
-		pathTracer.renderSample();
+		for ( let i = 0; i < params.samplesPerFrame; i ++ ) {
+
+			pathTracer.renderSample();
+
+		}
 
 	} else if ( ( delaySamples > 0 || ! params.enable ) && renderer.initialized !== false ) {
 
@@ -250,7 +256,7 @@ function onHashChange() {
 
 function onParamsChange() {
 
-	if ( pathTracer.tiles.x * pathTracer.tiles.y !== 1.0 ) {
+	if ( pathTracer.tiles !== 1.0 ) {
 
 		delaySamples = 1;
 
@@ -353,6 +359,8 @@ function buildGui() {
 	} );
 	pathTracingFolder.add( params, 'bounces', 1, 20, 1 ).onChange( onParamsChange );
 	pathTracingFolder.add( params, 'transmissiveBounces', 1, 20, 1 ).onChange( onParamsChange );
+
+	pathTracingFolder.add( params, 'samplesPerFrame', 1, 30, 1 );
 
 	const comparisonFolder = gui.addFolder( 'Comparison' );
 	comparisonFolder.add( params, 'displayImage' );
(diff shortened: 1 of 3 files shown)


gkjohnson:
> Because of tiling, sample calculation is often pretty fast, so I think there should be an option to calculate multiple samples. It seems like an easy enough detail to leave to the user. Ideally, we would have a way to automatically determine this number to maximize gpu usage. What do you think? Should this be part of the pathtracer API?

I think keeping this as a loop is fine. I think the wave front approach will be more reliable in terms of a more granular "knob" to turn in the long run, especially as we add NEE, etc. Though it might be worth calling this "iterations per frame" or "tiles per frame" in the UI for the page since it doesn't map directly to the number of samples performed.

> Detailed sample counts from are difficult to get here as they are not saved or calculated anywhere

What do you mean by detailed samples are not saved anywhere? The sample count will be the same (+- 1) across the entire image since a single sample is evaluated all at once every frame. I suppose with NEE you may consider multiple samples to be taken at once but I think we can handle that nuance later.

TheBlek:
> What do you mean by detailed samples are not saved anywhere? The sample count will be the same (+- 1) across the entire image since a single sample is evaluated all at once every frame. I suppose with NEE you may consider multiple samples to be taken at once but I think we can handle that nuance later.

Well, in WebGL pathtracer no sample count is kept on the gpu, only the cpu side keeps track. And they might differ because of asynchronous execution. Which is why using cpu-side sample count is not _precise_. However, and I tested that with an above-mentioned 1x1 texture, they differ by only a couple of samples which is negligible at higher sample counts.

gkjohnson:
> And they might differ because of asynchronous execution. Which is why using cpu-side sample count is not precise. However, and I tested that with an above-mentioned 1x1 texture, they differ by only a couple of samples which is negligible at higher sample counts.

I'm trying to understand how this could happen unless there's a bug in the current sample counting or the added GPU side approach that had been used. Every time "renderSample" is called the sample count is incremented by the ratio relative to the screen of the tile rendered (so 1 fill sample will be tallied once all tiles re issued). If you issue a readback of some kind for an equivalent GPU-side sample count texture or storage then all of the prior issued commands will have finished and you should see that the same number of samples will have finished GPU-side as had been issued at the time of initiating the readback.

The only thing I could see happening is if you're using an asynchronous readback mechanism for the GPU-side texture then by the time the value has been returned to the CPU other render operations will have been issued and so the CPU-side variable will have been incremented further.

I think there must be something else happening here.

TheBlek:
> The only thing I could see happening is if you're using an asynchronous readback mechanism for the GPU-side texture then by the time the value has been returned to the CPU other render operations will have been issued and so the CPU-side variable will have been incremented further.

Yes, this was an asynchronous readback so CPU will have already issued some render operations by the time it finishes. But really, at any given time CPU counter might be behind the GPU one because either GPU commands are not started or currently processing.

This is a very minor detail that I perhaps inadvertently drew too much attention. I meant to justify using cpu side counters for performance here.

gkjohnson:
Yeah I think we agree. Ultimately this behavior of the WebGLPathTracer.getDetailedSampleCount are returning the same thing: the number of samples taken at the time the function is called. The WebGLPathTracer version will just resolve more quickly.

PR #761, opened 2026-04-29
WebGPU: Sobol random number generator

TheBlek:
@gkjohnson 

This PR makes a lot of changes in order to support selecting random number generator in API:
 - Implements sobol numbers by porting webgl implementation in two version: with and without texture lookup; From my cursory performance testing it's not conclusive whether texture lookup makes sizable difference, curious to know how it work on your computer
 - Create and manage function nodes for random numbers in WebGPUPathtracer. That means passing them down to everything that needs them: PathtracerBVHComputeData, material and backend in general.
 - Current material is stored and handled inside WebGPUPathtracer now. This allows handling random numbers for a material by default and leaving `setRandomFunctions` on PathtracerBackend unspecified, so derivations don't have to call super before doing their stuff. Not sure about this decision, but it felt easier to think about the system with material handled here. Let me know what you think.
 - In order to make random number nodes hot-swappable I had to resort to a lot of `proxyFn` nodes which feels a little messy. Perhaps there could be one place that creates proxy nodes and then distributes them? Not sure how this should be handled

Here's some image comparisons between pcg and sobol random numbers. It looks like diffuse surfaces are not much by more uniform random numbers, unlike specular which has less noise (see clearcoat image).
| PCG | Sobol |
| ---- | ------- |
| khronos-ClearCoatCarPaint | sobol-khronos-ClearCoatCarPaint |
| pcg-khronos-MetalRoughSpheres |  khronos-MetalRoughSpheres |
| pcg-khronos-Lantern | khronos-Lantern |


TheBlek:
@gkjohnson Let me know if you want this PR divided or just want to take a different direction for implementing different random number generators.

src/webgpu/nodes/material.wgsl.js
```diff
@@ -407,15 +393,15 @@ export const albedoIntegralMetallic = wgslFn( /* wgsl */ `
 		var result = 0.0;
 		for ( var i = 0; i < INTEGRATION_SAMPLES; i++ ) {
 
-			let wh = ggxDirection( wo, vec2( alpha ), pcgRand2() );
+			let wh = ${ ggxDirectionFunc }( wo, vec2( alpha ), ${ pcgFunctions.vec2f }() );
```
gkjohnson: It looks like this is _always_ using "pcg" randomness. Is this intended?

src/webgpu/WaveFrontPathTracer.js
```diff
@@ -257,7 +273,7 @@ export class WaveFrontPathTracer extends PathTracerBackend {
 				primeRayGenerationDispatchKernel.outputDispatch = rayGenerationDispatch;
 
 				// set up the ray generation kernel
-				enqueueRaysKernel.seed ++;
+				enqueueRaysKernel.seed = this.seed;
```
gkjohnson: Can you explain the need for this change? We should still need to store a different seed per ray because they may each be at a different point in the path.

gkjohnson:
Here's a quick comparison of the random functions on the `khronos-AnimatedCube` w/ 100 samples. It's subtle but I just wanted to make sure there was a perceptible difference. I somewhat expected a bit more, though?

| PCG | Sobol |
|---|---|
| image | image |
| image | image |

> From my cursory performance testing it's not conclusive whether texture lookup makes sizable difference, curious to know how it work on your computer

I'm not seeing any difference between the performance of PCG, SOBOL, or SOBOL_TEXTURE. I expect the performance won't be dominated all that much by the random function implementation.

I'm thinking we can just keep the non-texture version of the SOBOL randomness unless there's a good reason to do otherwise?

> Create and manage function nodes for random numbers in WebGPUPathtracer. That means passing them down to everything that needs them: PathtracerBVHComputeData, material and backend in general.
>
> In order to make random number nodes hot-swappable I had to resort to a lot of proxyFn nodes which feels a little messy. Perhaps there could be one place that creates proxy nodes and then distributes them? Not sure how this should be handled

I'll think through this, as well - I agree it's not ideal to have to string the randomness functions down through to every subfunction.

src/webgpu/nodes/random.wgsl.js
```diff
@@ -1,27 +1,36 @@
-import { wgsl, wgslFn } from 'three/tsl';
+import { uint, float, wgsl, wgslFn, texture } from 'three/tsl';
+import { wgslTagFn } from '../lib/nodes/WGSLTagFnNode';
+
+// Alpha test should be the last index as it is summed with triangle index
+export const RNG_INDEX_RAY_JITTER = 0;
+export const RNG_INDEX_ENVIRONMENT_SAMPLE = 1;
+export const RNG_INDEX_SCATTER_TYPE = 2;
+export const RNG_INDEX_SCATTER_DIRECTION = 3;
+export const RNG_INDEX_ALPHA_TEST = 4;
```
gkjohnson, line 9: I'm not fully understanding what this is for?

src/webgpu/compute/wavefront/RayGenerationKernel.js
```diff
@@ -92,7 +101,6 @@ export class RayGenerationKernel extends ComputeKernel {
 				rayQueue[ index ].pixel = indexUV;
 				rayQueue[ index ].throughputColor = vec3f( 1.0 );
 				rayQueue[ index ].currentBounce = 0;
-				rayQueue[ index ].pcgStateS0 = ${ getPcgSeed }();
```
gkjohnson, line 95: Can you explain how this is working, now? Is this replaced by RNG_INDEX_RAY_JITTER?

gkjohnson:
> > Create and manage function nodes for random numbers in WebGPUPathtracer. That means passing them down to everything that needs them: PathtracerBVHComputeData, material and backend in general.
> >
> > In order to make random number nodes hot-swappable I had to resort to a lot of proxyFn nodes which feels a little messy. Perhaps there could be one place that creates proxy nodes and then distributes them? Not sure how this should be handled
>
> I'll think through this, as well - I agree it's not ideal to have to string the randomness functions down through to every subfunction.

Having thought through this a bit more - I'm thinking we don't really need a user-settable randomness function (at least for now). We can keep the PCG functions around but if we're seeing that Sobol is better than we should just export and use that everywhere. Then for dev we can change the randomness functions around in that source file if we have to. I don't think the toggle features is worth the implementation complexity atm.

src/webgpu/nodes/material.wgsl.js
TheBlek: Yes, kind of. I think it does not really matter which random function we use here since there are many samples taken.

src/webgpu/nodes/random.wgsl.js
TheBlek: Those are unique numbers for each sample of random numbers that are used for sobol number generation. They appear in webgl pathtracer as plain numbers:

`state.isShadowRay = scatterRec.specularPdf < rand( 4 );`

For better structure and to ensure there are no duplicates, I decided to define those constants here.

src/webgpu/compute/wavefront/RayGenerationKernel.js
TheBlek: Not really, it just initializes each time without carrying it over. Will add a little fix to use current bounce in initialization. Although, if we are to keep only one random number generation algorithm, it should not matter as sobol numbers' state can be properly initialized on each stage properly.

TheBlek:
> Then for dev we can change the randomness functions around in that source file if we have to.

I understand not exposing the toggle to the users, but even implementing new random numbers and changing them in every place is tedious and make comparisons harder.

I simplified the implementation a bit by extracting proxyFn nodes into a single object that is passed around. Let me know if complexity is still an issue.

src/webgpu/WaveFrontPathTracer.js
TheBlek: You're right, if we're being proper, seed should be different for each ray. Fixed that.

src/webgpu/nodes/material.wgsl.js
gkjohnson: I think we should be using the same randomness function everywhere so we can take advantage of the sample stratification.

src/webgpu/nodes/random.wgsl.js
gkjohnson: Ha I seem to have forgotten about this 😅  I'm going to have to remind myself of this when I have a chance this weekend.

src/webgpu/compute/wavefront/RayGenerationKernel.js
gkjohnson: Okay I'll look into this over the weekend, as well - I don't recall exactly how sobol is working but ensuring this pcg seed is consistent between rays in the path provided some noticeable improvements.

gkjohnson:
> I understand not exposing the toggle to the users, but even implementing new random numbers and changing them in every place is tedious and make comparisons harder.

Yeah I didn't intend for this to be the approach. More that we could define a common "interface" for these functions and just replace the underlying definition in the source file depending on which randomness solution we wanted to use. Something like so:


```js
// random.wgsl.js

// PCG Definitons
// export {
//   pcgRand1 as rand1,
//   pcgRand2 as rand2,
//   pcgSetSeed as setSeed,
//   ...
// };

// Sobol Definitons
export {
  sobolRand1 as rand1,
  sobolRand2 as rand2,
  sobolSetSeed as setSeed,
  // ...
};
```

Then in order to swap between the two randomness types it would just be a matter of changing the export block used. That might mean there are some "dead" parameters in some of these functions but I think that's okay for now.

--

Sorry for delays on some of my responses - I've been balance a few things recently. I'll try to respond to the other PR and this one a bit further in the next few days!

src/webgpu/materials/PathtracingMaterial.js
```diff
@@ -23,7 +25,23 @@ export class PathtracingMaterial {
...
 
-		return lambertBsdfFunc;
+		return wgslTagFn`
+
+			fn bsdfSample( worldWo: vec3f, surf: SurfaceRecord ) -> ScatterRecord {
+
+				var record: ScatterRecord;
+
+				let wo = normalize( surf.normalInvBasis * worldWo );
+				let wi = ${ diffuseDirectionFunc }( wo, ${ this.rng.vec2f }( ${ RNG_INDEX_SCATTER_DIRECTION } ) );
```
gkjohnson: IS this supposed to be `rand2`?

src/webgpu/nodes/material.wgsl.js
```diff
@@ -392,7 +377,8 @@ export const albedoIntegralMetallic = wgslFn( /* wgsl */ `
 	) -> void {
 
 		const INTEGRATION_SAMPLES = ( 1 << 20 );
-		pcgInitialize( globalId.xy, 0 );
+		let pixelIndex = ( indexUV.x << 16 ) | indexUV.y;
+		${ rngInit }( pixelIndex, 0, 0 );
```
gkjohnson, line 516: Is this correct? Is "indexUV" defined anywhere?

src/webgpu/compute/wavefront/RayGenerationKernel.js
```diff
@@ -80,10 +80,11 @@ export class RayGenerationKernel extends ComputeKernel {
 				let queueCapacity = arrayLength( rayQueue );
 				let index = atomicAdd( &queueSizes[ 1 ], 1 ) % queueCapacity;
 
-				${ pcgInit }( indexUV, seed );
+				let pixelIndex = ( indexUV.x << 16 ) | indexUV.y;
+				${ rngInit }( pixelIndex, seed, 0 );
```
gkjohnson, line 84: This should probably use the current ray bounce index when generating the rays - in the megakernel, for example, this is incremented every iteration so it will be set before the subsequent ray is generated.

src/webgpu/compute/PathTracerMegaKernel.js
```diff
@@ -135,6 +135,8 @@ export class PathTracerMegaKernel extends ComputeKernel {
 
 				for ( var bounce = 0u; bounce < bounces; bounce ++ ) {
 
+					${ rngNextBounce }();
```
gkjohnson: I'm thinking this should be at the _end_ of the loop? Or we can call `rngInit( pixelIndex, seed, bounce )` at the top here. Otherwise the first ray will always run with a bounce index of "1" vs "0", as the wavefront version does. I'm just trying to make sure we're matching the state across both kernels as much as we can

gkjohnson (commented):
Some final comments but this looks good! The random handling looks much easier to reason about now, I think.

src/webgpu/compute/wavefront/RayGenerationKernel.js
TheBlek: In the megakernel we do not generate multiple rays per pixel in one invocation, if I correctly understood what you mean. This kernel runs once per path, at the start, and it runs with bounce 0, same as initial ray generation in megakernel. Did I misunderstand you intent here?

gkjohnson: Yup! You're right - this will always be the ray for the first bounce.

gkjohnson:
Amazing, thanks!

PR #762, opened 2026-04-29
WebGPUPathtracer: Iridescence extension

TheBlek:
This PR extends brdf to take into account iridescene. Code ported from official gltf example implementation and effectively equivalent to webgl. Iridescent fresnel calculation atleast seems to be the same. However, in practice images appear to be very different:
| webgl | webgpu |
| ------ | --------- |
|  khronos-IridescenceSuzanne | webgpu-khronos-IridescenceSuzanne |

Notably:
- Metallic suzanne appears to be mostly correct, except for the directional light's highlights
- Dielectric suzanne is overly saturated but over time is becoming more white-ish. I wonder if this will be fixed when light source sampling for last vertex is implemented

I'm not confident in this implementation because that and would like to hold on to this until we can validate dielectric materials correctness with better sampling techniques.

### DIFF
--- a/src/webgpu/materials/GltfCompliantMaterial.js
+++ b/src/webgpu/materials/GltfCompliantMaterial.js
@@ -2,7 +2,7 @@ import { texture, textureStore, globalId, float } from 'three/tsl';
 import { StorageTexture, RedFormat, LinearFilter, TextureLoader, HalfFloatType } from 'three/webgpu';
 import { wgslTagFn } from '../lib/nodes/WGSLTagFnNode';
 import { PathtracingMaterial } from './PathtracingMaterial';
-import { specularBrdfFunc, diffuseBrdfFunc, fresnelMixFunc, conductorFresnelFunc, albedoIntegralMetallic, fresnelCoatFunc } from '../nodes/material.wgsl.js';
+import { specularBrdfFunc, diffuseBrdfFunc, fresnelMixFunc, conductorFresnelFunc, albedoIntegralMetallic, fresnelCoatFunc, iridescentDielectricLayerFunc, iridescentConductorLayerFunc } from '../nodes/material.wgsl.js';
 import { diffuseDirectionFunc, getLobeWeightsFunc } from '../nodes/sampling.wgsl.js';
 import { ggxDirectionFunc, ggxReflectionAdjustedPDFFunc } from '../nodes/ggx.wgsl.js';
 import { bxdfContextStruct, scatterRecordStruct, surfaceRecordStruct } from '../nodes/structs.wgsl.js';
@@ -27,6 +27,8 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 			fresnelMix = fresnelMixFunc,
 			conductorFresnel = conductorFresnelFunc,
 			fresnelCoat = fresnelCoatFunc,
+			iridescentDielectricLayer = iridescentDielectricLayerFunc,
+			iridescentConductorLayer = iridescentConductorLayerFunc,
 			calculateTurquinTexture = false,
 		} = options;
 
@@ -53,6 +55,8 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 		this.fresnelMix = fresnelMix;
 		this.conductorFresnel = conductorFresnel( turquinNode );
 		this.fresnelCoat = fresnelCoat;
+		this.iridescentDielectricLayer = iridescentDielectricLayer;
+		this.iridescentConductorLayer = iridescentConductorLayer;
 		this.calculateTurquinTexture = calculateTurquinTexture;
 
 	}
@@ -85,9 +89,19 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 
 				let specular = ${ this.specularBrdf }( ctx.NdotL, ctx.NdotV, ctx.NdotH, alpha );
 				let diffuse = ${ this.diffuseBrdf }( ctx.NdotV, ctx.NdotL, ctx.VdotH, surf );
-				let dielectric = ${ this.fresnelMix }( ctx.VdotH, surf.ior, diffuse, specular );
+				let dielectricBase = ${ this.fresnelMix }( ctx.VdotH, surf.ior, diffuse, specular );
 
-				let metallic = ${ this.conductorFresnel }( ctx.NdotV, ctx.VdotH, surf.color, specular, alpha );
+				let dielectric = ${ this.iridescentDielectricLayer }(
+					dielectricBase, diffuse, specular, ctx.VdotH, /* outsideIor */ 1.0,
+					surf.ior, surf.iridescenceIor, surf.iridescenceThickness, surf.iridescence
+				);
+
+				let metallicBase = ${ this.conductorFresnel }( ctx.NdotV, ctx.VdotH, surf.color, specular, alpha );
+
+				let metallic = ${ this.iridescentConductorLayer }(
+					metallicBase, specular, surf.color, ctx.VdotH, /* outsideIor */ 1.0,
+					surf.iridescenceIor, surf.iridescenceThickness, surf.iridescence
+				);
 
 				let material = mix( dielectric, metallic, surf.metalness );
 
--- a/src/webgpu/nodes/material.wgsl.js
+++ b/src/webgpu/nodes/material.wgsl.js
@@ -1,4 +1,4 @@
-import { mat3, wgsl, wgslFn } from 'three/tsl';
+import { mat3, wgslFn } from 'three/tsl';
 import {
 	inverseMat3x3Func,
 	getBasisFromNormalFunc,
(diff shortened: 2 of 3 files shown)


gkjohnson:
This looks great!

> Metallic suzanne appears to be mostly correct, except for the directional light's highlights

Directional lights will require next event estimation so this is fine.

> Dielectric suzanne is overly saturated but over time is becoming more white-ish. I wonder if this will be fixed when light source sampling for last vertex is implemented

I would expect this to be the case. But the noise will potentially be primarily a result of the environment map and no NEE rather than issues with the iridescence path. Using a purely white or a gradient env map would make it easier to compare the results on the left-most dielectric model. I'm thinking if we can get that looking right then we can merge this?

src/webgpu/nodes/material.wgsl.js
```diff
@@ -353,6 +357,147 @@ export const fresnelMixFunc = wgslFn( /* wgsl */ `
...
+const iridConst = wgsl( /* wgsl */ `
+
+const XYZ_TO_REC709 = mat3x3(
+	3.2404542, - 0.9692660, 0.0556434,
+	- 1.5371385, 1.8760108, - 0.2040259,
+	- 0.4985314, 0.0415560, 1.0572252
+);
+
+
+` );
```
gkjohnson: Is this something that can be done with a mat3 in TSL, now? In the latest version of three.js global variable declarations are working:

```js
const XYZ_TO_REC709 = mat3(
	3.2404542, - 0.9692660, 0.0556434,
	- 1.5371385, 1.8760108, - 0.2040259,
	- 0.4985314, 0.0415560, 1.0572252,
).setName( 'XYZ_TO_REC709' );
```

The variable will be declared globally by default.

src/webgpu/nodes/material.wgsl.js
```diff
@@ -353,6 +357,147 @@ export const fresnelMixFunc = wgslFn( /* wgsl */ `
...
+    const val = vec3(5.4856e-13, 4.4201e-13, 5.2481e-13);
+    const pos = vec3(1.6810e+06, 1.7953e+06, 2.2084e+06);
+    const _var = vec3(4.3278e+09, 9.3046e+09, 6.6121e+09);
+
+    var xyz = val * sqrt(2.0 * ${ Math.PI } * _var) * cos(pos * phase + shift) * exp(-phase * phase * _var);
+    xyz.x += 9.7470e-14 * sqrt(2.0 * ${ Math.PI } * 4.5282e+09) * cos(2.2399e+06 * phase + shift.x) * exp(-4.5282e+09 * phase * phase);
+    xyz /= 1.0685e-7;
+
+    let rgb = XYZ_TO_REC709 * xyz;
+    return rgb;
```
gkjohnson: nit: some indentation issues. I wonder if there's any way to get these template strings to be linted in any way 🤔

src/webgpu/nodes/material.wgsl.js
```diff
@@ -353,6 +357,147 @@ export const fresnelMixFunc = wgslFn( /* wgsl */ `
...
+const XYZ_TO_REC709 = mat3x3(
+	3.2404542, - 0.9692660, 0.0556434,
+	- 1.5371385, 1.8760108, - 0.2040259,
+	- 0.4985314, 0.0415560, 1.0572252
+);
+
+
+` );
+
+const evalSensitivityFunc = wgslTagFn`
```
gkjohnson: nit: Would be good to include the wgsl tag here for syntax highlighting:

```js
wgslTagFn/* wgsl */`
```

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -85,9 +89,19 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
...
+					dielectricBase, diffuse, specular, ctx.VdotH, /* outsideIor */ 1.0,
+					surf.ior, surf.iridescenceIor, surf.iridescenceThickness, surf.iridescence
+				);
+
+				let metallicBase = ${ this.conductorFresnel }( ctx.NdotV, ctx.VdotH, surf.color, specular, alpha );
+
+				let metallic = ${ this.iridescentConductorLayer }(
+					metallicBase, specular, surf.color, ctx.VdotH, /* outsideIor */ 1.0,
+					surf.iridescenceIor, surf.iridescenceThickness, surf.iridescence
+				);
```
gkjohnson, line 104: I'm guessing for "outsideIor" this is something we'll need to track per ray? Something like the "current medium"? I expect other path tracers will track multiple mediums but I'm not exactly sure what the right way to do this would be considering we can have overlapping geometries 🤔

TheBlek: Yeah, ior tracking is not really a viable option since we can enter a transmissive body, never hit a back facing triangle and still "exit" the body because of alpha transparency/non-watertight meshes.
If we think about it as an ior of medium that *interfaces* with iridescence layer, then its 1.0 ( other value in case there is a clearcoat layer? ) when triangle is hit from the front and ior of the object itself when hit from the back.

src/webgpu/nodes/material.wgsl.js
TheBlek: This is very weird, since in my editor its displayed correctly:

Image

gkjohnson: It's a mix of tabs and spaces (you can tell by highlighting). I'm not sure what editor you use but [editor config](https://editorconfig.org/) has a plugin for most IDEs (I think?) and will automatically configure "tab" behavior based on [the config within a project](https://github.com/gkjohnson/three-gpu-pathtracer/blob/main/.editorconfig) to help enforce consistency.

src/webgpu/materials/GltfCompliantMaterial.js
gkjohnson: > Yeah, ior tracking is not really a viable option since we can enter a transmissive body, never hit a back facing triangle and still "exit" the body because of alpha transparency/non-watertight meshes.

I think it's okay. If users are marking non-watertight meshes as transmissive it will have to be expected that things may behave a bit oddly. This is a probably for another time, though, I think.

> If we think about it as an ior of medium that interfaces with iridescence layer, then its 1.0 ( other value in case there is a clearcoat layer? ) when triangle is hit from the front and ior of the object itself when hit from the back.

I'm not sure I'm completely following this. Perhaps we're talking about different things, but Ignoring clearcoat for a moment (which I believe is modeled as above the iridescence layer) what I'm intended to say is: we're tracking a ray as it enters the iridescence layer, is color shifted based on the ior difference, and then bounced back out of it. So that "outside ior" depends on the medium that the ray had last entered - a ray entering a body of water and then bouncing off of an iridescent object, for example. The "outside ior" should have to be ~1.33 in that case and will impact the color shift from the model.

Not that we have to get to this level of granularity. I just want to know where the lines are at the moment and keep things in mind for later.

TheBlek: > If users are marking non-watertight meshes as transmissive it will have to be expected that things may behave a bit oddly. This is a probably for another time, though, I think.

Hm, I wonder if some kind of cheap check can be done on transmissive meshes for users that encounter weird artifacts. This ties into the idea of checking user's mesh attributes for valid tangent / normal combination mentioned in https://github.com/gkjohnson/three-gpu-pathtracer/pull/758#issuecomment-4318277199. 

> a ray entering a body of water and then bouncing off of an iridescent object, for example

Okay, yes, I overlooked that case. Ior would need to be tracked for correct interactions of light with objects inside of other objects. What I meant to say is that material has a layered model:

ray from back -> | base | iridescence | clearcoat | <- ray from front

So if the ray is entering from the back, it interacts with base and then transmits to iridescence layer which makes baseIor the outside ior for this interaction.

And if its entering from the front, then maybe clearcoatIor should be used as outsideIor for the interaction if the layer exists.

At least this is my understanding of how it should be modeled taking layers into account. I don't recall seeing this anywhere so should be double-checked.

Tracking current medium's ior does make me think about how one would determine ior in the camera point.

I agree, this is an issue for a more detailed transmission implementation.

gkjohnson: > Tracking current medium's ior does make me think about how one would determine ior in the camera point.

In the WebGL version the ray origin is tested for being within any relevant volume using single raycast. If a backside is hit that means we're "inside" the volume. This was done for fog volumes, at least - I hadn't added transmissive volume support for this but the same should apply.

> ray from back -> | base | iridescence | clearcoat | <- ray from front

This is my impression, as well, which gets a bit complicated. Though iridescent + clearcoat models may be less common.

TheBlek:
I've gone ahead and implemented NEE with directional lights and env map sampling in separate branch (and a bunch more, but thats for later). And indeed, noise on the left suzanne is much lower:

| webgl | webgpu + NEE |
| ------- | ---------------- |
| khronos-IridescenceSuzanne | khronos-IridescenceSuzanne  |

So now I'm pretty sure about the implementation here. Let me know if you want to hold out on merging this or get more information.

gkjohnson:
This is amazing, thanks! Sorry for the delays on getting this stuff merged - I've had too many other things to keep up with.

PR #763, opened 2026-04-29
WebGPUPathtracer: Screenshot capture

TheBlek:
@gkjohnson
Adds ability to capture screenshots of webgpu pathtracing results in update-screenshots.js via an additional flag. 

I used this to capture latest screenshots. Or is there an easier way that I missed?

When launching with current puppeteer, there was a weird error in the console, that disappeared after update. It launched a _very_ old chromium version. I understand this is a big change (9 major versions). What is your policy on upgrading libraries like that? When to update and which version to choose?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "pixelmatch": "^5.3.0",
         "pngjs": "^6.0.0",
         "process": "^0.11.10",
-        "puppeteer": "^15.4.0",
+        "puppeteer": "^24.42.0",
         "rollup": "^2.70.0",
         "simple-git": "^3.10.0",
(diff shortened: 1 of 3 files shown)


gkjohnson:
This looks great - I didn't know puppeteer supported WebGPU well enough, yet.

> I understand this is a big change (9 major versions). What is your policy on upgrading libraries like that? When to update and which version to choose?

Generally it's best to keep the dependencies as up to date as possible but it's a chore to do so consistently. Some projects use [renovate bot](https://github.com/renovatebot/renovate), I believe, to keep things up to date but I'm unfamiliar with it. It might be worth looking in to.

PR #764, opened 2026-05-03
WebGPUPathtracer: Fixes rare NaN issues

TheBlek:
@gkjohnson 

On my machine there were few rare NaN pixels that were appearing over time:
image

This was due to `cos` function in `schlickFresnel` receiving a negative value because `NdotL` was bigger than 1. This is fixed by renormalizing a vector after reflect function. Or maybe we should just clamp the `NdotL` calculation?

### DIFF
--- a/src/webgpu/materials/GltfCompliantMaterial.js
+++ b/src/webgpu/materials/GltfCompliantMaterial.js
@@ -157,7 +157,7 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 				} else if ( r <= cdf.y ) { // specular
 
 					wh = ${ ggxDirectionFunc }( wo, vec2( alpha ), ${ pcgRand2 }() );
-					wi = - reflect( wo, wh );
+					wi = - normalize( reflect( wo, wh ) );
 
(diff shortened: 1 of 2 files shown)


gkjohnson:
> This was due to cos function in schlickFresnel receiving a negative value because NdotL was bigger than 1. This is fixed by renormalizing a vector after reflect function. 

Do we know how it was the case that the reflected vector is > 1 magnitude after reflection? Is it floating point error? Or are `wo` or `wh` non-normalized? I'm wondering if it would be better to just ensure those two vectors are normalized to start with, instead?

```wgsl
wi = - normalize( reflect( wo, wh ) );
```

TheBlek:
> Is it floating point error?

I believe so, because both those vectors are explicitly normalized beforehand:

wh:
```
	fn ggxDirection( incidentDir: vec3f, alpha: vec2f, uv: vec2f ) -> vec3f {
                ...
		N = normalize( vec3( alpha * N.xy, max( 0.0, N.z ) ) );

		return N;

	}

```

wo:
```
let wo = normalize( invBasis * worldWo );
```

gkjohnson:
Okay how fun 😅

Thanks for investigating this!

PR #766, opened 2026-06-16
WebGPU: Add support for a texture atlas

gkjohnson:
Related to #713 

cc @theBlek

Adds support for a texture atlas using ["skyline" packing](https://jvernay.fr/blog/skyline-2d-packer/implementation/) - algorithm implementation mostly by Claude. There are some improvements that could be made but this is good enough until we run into some problems.

The atlas lets us be more texture-memory efficient while supporting higher texture resolutions than before. The width and height of the atlas are defined by the largest texture (up to 4096, which textures are sized to if needed) and if we overrun one "page" we create a new layer in the texture array to fit more. Previously every texture given it's own layer and scaled to 1024x1024.

From a technical side of things - since we're running up to storage buffer attribute limits the texture bounds are uploaded as an array uniform. And since we can't pass the full uniform array as a pointer down as function arguments some functions have been adjusted to use a "factory" pattern to pass the specific uniform reference and the PathTracerBVHComputeData has been made the owner of the atlas to simplify it.

You can also see the atlas layers in the "viewerTest" example through an added dropdown.

### MaterialsVariantsShoe

For the "MaterialsVariantsShoe" this is how the texture looks - 3x 2048 textures sampled at their native resolution:

| | After | Before | Optimal |
|---|---|---|---|
| sampled textures | 3 * 2048 | 3 * 1024 | 3 * 2048 |
| texture array | 3 * 2048 | 3 * 1024 | 3 * 2048 |
| memory | 50.33MB | 12.6MB | 50.33MB |

### ToyCar

And for this "ToyCar" example:

| | After | Before | Optimal|
|---|---|---|---|
| sampled textures | 5 * 1024 + 2 * 512 | 7 * 1024 | 5 * 1024 + 2 * 512 |
| texture array | 6 * 1024 | 7 * 1024 | 5 * 1024 + 2 * 512 |
| memory | 25.2MB | 29.3MB | 23MB |

--

And a peak at the texture quality difference on the shoe:

| Before | After |
|---|---|
| image | image |
| image | image |

This of course becomes more beneficial as scenes have more varied texture sizes - or using a lot of small textures in particular.


src/webgpu/AtlasTexture.js
```diff
@@ -0,0 +1,446 @@
...
+
+}
+
+// Patch the skyline after placing a rect spanning [ x, x + w ] with top edge y.
+// The left edge x always aligns to an existing segment boundary, so only the
+// right edge can split a segment. Mutates nodes in place.
+function addSkylineNode( nodes, x, y, w ) {
+
+	const right = x + w;
+	const result = [];
```
TheBlek, line 73: There are some inefficiencies in this skyline algorithm implementation that should be fine until we get to really complex scenes. For example an array creation can be avoided here.

src/webgpu/AtlasTexture.js
```diff
@@ -0,0 +1,446 @@
...
+	_hashesMatch( textures ) {
+
+		const hashes = this.hashes;
+		if ( hashes.length !== textures.length ) {
+
+			return false;
+
+		}
+
+		for ( let i = 0, l = textures.length; i < l; i ++ ) {
```
TheBlek, line 283: Seems like this can be reduced to `hashes.every( ( hash, i ) => hash === getTextureHash( textures[ i ] ) )`

src/webgpu/nodes/PathtracerBVHComputeData.js
```diff
@@ -42,24 +44,30 @@ export class PathtracerBVHComputeData extends BVHComputeData {
 		this.storage.materials = null;
 		this.materials = [];
 		this.bvhMap = new Map();
+		this.textureAtlas = new AtlasTexture();
```
TheBlek, line 47: Why did you opt into keeping AtlasTexture in PathtracerBVHComputeData? It seems to only be used in useTransparencyRaycastFn as was RenderTarget2DArray previously.

TheBlek (approved):
Great stuff! I think pathtracer can go far with this texture packing.

src/webgpu/nodes/PathtracerBVHComputeData.js
gkjohnson: The array texture needed this separate "textureInfoUniform" which couldn't be easily strung down through the function parameters, so I opted for a generated function that references them directly. We coudl manage that function separately and assign it to the path tracer data class but I'm not sure if I feel one is better than the other?

PR #767, opened 2026-06-17
WebGPU: Add support for multiple UV channels

gkjohnson:
Depends on #766

Adds support for multiple uv channels by adding any available geometry uv channels in the scene to the struct, and adding a function mapping a texture uv channel to the relevant uv. Also makes the path tracer resilient to no-uvs being present.

In the future it might be good to pack the uv channels into an array of vec2s so we can more easily sample the appropriate UV but that can be added later.

Tested with the "khronos-MultiUVTest" model in "viewerTest":

| Before | After |
|---|---|
| image | image |


gkjohnson:
I have some time to work on this at the moment so I'm going to merge this. I'll work on getting the color attribute conditionally included next to save memory.

PR #768, opened 2026-06-18
WebGPUPathtracer: Migrate examples

gkjohnson:
Related to #713 

Migrate all examples to use WebGPUPathtracer, removing all WebGLPathtracer references - this should make it easier to see which examples are and aren't working and work towards getting the remaining features in. I've also fixed path tracer so it works without calling "useMegakernel( bool )" and defaulted it to use the wavefront mode.

A number of the examples work quite well but I'm seeing some big performance drops in particular with the M2020 rover model in the "viewer" example (slower than the WebGL version). It's not clear to me why, yet.

**Other**
- Add webgpu-version BlurredEnvMapGenerator
- Add support for "stableNoise", "pause" fields
- Fix SkinnedMesh bounds
- Adjust the default to use the wave front pathtracer
- Fix support for Orthographic cameras

**TODO**
- Fix the remaining examples (or decide to punt)
  - Fog (punt)
  - Graphing (punt)
  - AO Render (punt)
  - LKG (delete / punt)
  - Render Video (delete / punt)
  - webgpu_primitives (delete?)

**Later**
- TLAS / BVH strategy for performance and adjusted getShapecastFn API
- "Stalenhag Winter", Rover models are slow.
  - Likely due to large TLAS overlap.
- Lego models are slower when _not_ merging them.
  - Visuals of the cockpit window change when merging the models for some reason?
- Simplify redundancy between "getBVH" and BVH generation logic in WebGPUPathTracer.


gkjohnson:
I've added a bounds debug view to understand what's going on performance-wise:

image

Some timing for the model:

| | timing |
|---|---|
| original | ~10fps |
| white standard mat | ~60fps |
| merged white mat | 120fps |

So it looks like the large TLAS overlaps may be the fundamental issue 🤔

gkjohnson:
Some possible future plans for improving the traversal performance:

### Shared BLAS

- MultiMeshBVH: Build one unified BVH over the triangles of multiple geometries - adding a dedicated "object index" per triangle. Or a "MixedBVH" that supports any kind of primitive / leaf?
- Objects with shared geometry do _not_ go in the shared BLAS. Nor do specially marked "dynamic" or "skinned" meshes.
  - Alternatively objects with shared geometry _could_ be placed into a single, shared multi-mesh BVH to speed things up at the cost of higher BVH node memory usage. Since each triangle index will come with an associated id we can share the underlying geometry data.

### Shared Data Model
- Keep vertices in object space (don't bake to world space) - preserves instancing memory, normals stay object-space.
- Build the unified BVH over world-space triangle bounds (transform at build time to compute bounds only).

### Dynamic Adjustments
- While moving: refit each frame
- On settle: trigger a full rebuild to restore tree quality.

### Build Performance

- Eventually use full GPU-based BVH construction and refit.
- Wide BVH (BVH4/BVH8) — separate traversal-efficiency win, deferred for now.

gkjohnson:
cc @TheBlek - is it okay to merge this? It enables a lot of new models for performance evaluation and will help with checking some of the improvements I plan to make after #771

TheBlek:
Hi @gkjohnson, sorry for long wait here. Could you wait for me to take a look at this later tomorrow? I am curious to see how bvh bounds visualization is implemented.

From cursory glance:
Do we want to migrate all examples to webgpu only? I expect there will be a period of adoption of webgpu where webgl will still be relevant. Or is this a temporary change to enable faster development?

gkjohnson:
> Do we want to migrate all examples to webgpu only? I expect there will be a period of adoption of webgpu where webgl will still be relevant. Or is this a temporary change to enable faster development?

From my perspective the WebGLPathTracer is legacy and won't be maintained or updated anymore. It's simply too brittle, slow, and inconsistent to maintain and that limits the practical use, as well. Looking forward to when the WebGPU path tracer is merged to main: we can keep the WebGLPathTracer around for a few releases but I anticipate deleting it entirely with a v0.1.0 release. If users want to keep using the WebGL version they can continue to use whatever the last release is that included it. I'm open to other opinions but my plan is to keep these examples WebGPU only from here and if we need to compare to the WebGL version for feature parity etc we can look at the main branch.

TheBlek (approved):

PR #774, opened 2026-07-04
WebGPU: Convert material reference to uint rather than struct

gkjohnson:
@TheBlek - I think this is what we'd discussed in #770: converting the global scratch material reference to a uint rather than copying the full material to improve register use. Maybe this is where better GPU-side timing instrumentation would be useful but I'm not seeing any meaningful difference in performance or anything when testing the rover model. Are there improvements on your end?


TheBlek:
Profiling on my computer, there is meaningful difference in register allocation:

| Mode | webgpu-pathtracer | this patch |
| ------- | ------- | ----- |
| Megakernel | RadeonGPUProfiler_ucZf6MWoUO | RadeonGPUProfiler_tsQqTRxNaP |
| Wavefront (trace ray kernel) | RadeonGPUProfiler_IryX6uG12E | RadeonGPUProfiler_a0af2rNTx8 |

And performance reflects that. With this patch I'm seeing ~10% improvement. (Well, if looking at the captured traces. Perf tool is still WIP, hope to finish later this week)

Problem with those kind of optimisations is that they heavily depend on the gpu: can it increase occupancy even more?; is reduction in register usage enough to increase occupancy? will the change trigger some compiler's heuristic that makes him keep some data in registers now that were previously refetched? etc. With this in mind, it would be very useful to see more details on register usage, occupancy limiters, general bottleneck. Unfortunately, I don't have access to an Apple device so can't investigate myself.

gkjohnson:
Thanks for the details and explanation - I think it makes sense to merge this and it intuitively makes sense that there could be some benefit.

CodyJasonBennett:
Is that NSight (NVIDIA) or Radeon (AMD)? I was just about to link https://developer.playcanvas.com/user-manual/optimization/gpu-profiling for Apple.

TheBlek:
> Is that NSight (NVIDIA) or Radeon (AMD)?

Screenshots are from Radeon dev tools on windows. [[1]](https://frguthmann.github.io/posts/profiling_webgpu/) [[2]](https://theblek.online/posts/webgpu-profiling/)

PR #775, opened 2026-07-04
WebGPU: Use global workgroup stack

gkjohnson:
Related to #774, #770

This PR adjusts the node traversal to use a global "workgroup" space buffer as a set of stacks for use across the threads, as suggested in #770. There are technically two changes here: the TLAS and BLAS now share a common stack for traversal, and the stack is packed packed into a workgroup buffer. The stack is still 60 elements deep, accommodating both TLAS and BLAS.

I've also tried just moving the staack to a module level variable (commented out here) and I'm actually seeing that that's _faster_ than the workgroup version. Here is some timing using the rover model on my machine (performance changes based on whether it's plugged in, which is why the baseline may be different from before):

| | baseline | workgroup space | private space |
|---|---|---|---|
| timing | ~53-60 fps | ~40-43 fps | ~55-63 fps |

So I _am_ seeing some improvement but it's only when using a private space variable. And so far it's not clear if it's due to sharing a single stack (eg less memory allocated) or if it's due to moving the stack to a module level variable. It will probably be better to test after #773 is merged. If we adjust the stack depth to 40 I'm seeing that workgroup space and private space start run at a more comparable framerate.

@TheBlek is this lining up with what you're seeing? Or have I missed something in the implementation?

### DIFF
--- a/src/webgpu/compute/PathTracerMegaKernel.js
+++ b/src/webgpu/compute/PathTracerMegaKernel.js
@@ -1,9 +1,9 @@
 import { DataTexture, Matrix3, Matrix4, Vector2, StorageTexture } from 'three/webgpu';
 import { ComputeKernel } from './ComputeKernel.js';
-import { texture, sampler, uniform, globalId, textureStore } from 'three/tsl';
+import { texture, sampler, uniform, globalId, textureStore, localId, vec3 } from 'three/tsl';
 import { rngInit, rngNextBounce, rand2, RNG_INDEX_RAY_JITTER, RNG_INDEX_ENVIRONMENT_SAMPLE } from '../nodes/random.wgsl.js';
 import { sampleEnvironmentFn, weightedAlphaBlendFn } from '../nodes/sampling.wgsl.js';
-import { ndcToCameraRay, proxy, proxyFn, wgslTagFn } from '../lib/three-mesh-bvh/index.js';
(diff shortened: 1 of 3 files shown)


gkjohnson:
After merging #773 and updating this PR here is the fps I'm seeing with wavefront kernel and the rover model at a stack size of 60:

| Workgroup space (shared stack) | Private space | Function space |
|---|---|---|
| ~55-63fps | ~79-84fps | ~79-84fps |

TheBlek:
With current unified stack, moving it all to workgroup memory is a bad idea. Because it will occupy the whole LDS and limit occupancy.

See register allocation of baseline:
image

And with whole stack in workgroup memory:
image

Note how occupancy is lower in second case.

The original optimisation idea was to move just the *half* of the stack so as to allow 50% occupancy. (moving more to the LDS will lower occupancy further which is undesirable):
image

However, it seems that moving a half of the unified stack is not cheap. I tested with an `if` and a `select` for access: both variants were slower than the baseline. So, it seems to me like a dead end. Perhaps LDS can be used in another way.

gkjohnson:
Oh interesting - thanks for the investigation. What tool are you using to do the GPU memory inspection?

I'll close this for now then

TheBlek:
> What tool are you using to do the GPU memory inspection?

This is Radeon GPU Profiler. Its vendor specific, so it probably won't help you much. XCode Metal debugger could have similar statistics: https://toji.dev/webgpu-profiling/xcode.html

PR #776, opened 2026-07-05
WebGPU: Add support for equirect camera, depth of field

gkjohnson:
Adds support for "EquirectCamera" and "PhysicalCamera" (w/ bokeh etc) to WebGPUPathTracer. The path tracer retrieves a "camera ray generation" function node from a `getCameraRayFn` function on the camera, which is used to slot into the path tracer kernels. The original ortho and perspective camera-compatible function is used by default. An "update" function is also called on pathtracer reset to ensure that the camera-owned uniforms required for the function are updated.

This gets both the "interior" demo and "depthOfField" demo both now demonstrate both equirect and bokeh rendering respectively:

image

image

This also lets users specify their own custom ray generation function to make stuff. I'm thinking a similar pattern could be used for custom materials in the future, as well:

<details>
<summary>Code Snippet</summary>

```js
camera.getCameraRayFn = function getCameraRayFn() {

	const invViewProjectionMatrix = uniform( new Matrix4() );
	const swirl = uniform( 0 );

	const fn = wgslTagFn/* wgsl */`
		fn getCameraRay( uv: vec2f, resolution: vec2f ) -> ${ rayStruct } {

			let ndc = uv * 2.0 - vec2f( 1.0 );

			// rotate the sample point about the center, more the further out it is
			let aspect = resolution.x / resolution.y;
			var p = ndc * vec2f( aspect, 1.0 );
			let angle = ${ swirl } * length( p );
			let s = sin( angle );
			let c = cos( angle );
			p = vec2f( p.x * c - p.y * s, p.x * s + p.y * c );
			p /= vec2f( aspect, 1.0 );

			return ${ ndcToCameraRay }( p, ${ invViewProjectionMatrix } );

		}
	`;

	const update = () => {

		this.coordinateSystem = WebGPUCoordinateSystem;
		this.updateMatrixWorld();
		this.updateProjectionMatrix();

		invViewProjectionMatrix.value.multiplyMatrices( this.matrixWorld, this.projectionMatrixInverse );
		swirl.value = options.swirl;

	};

	return { fn, update };

};
```
</details>

image


src/webgpu/nodes/random.wgsl.js
```diff
@@ -6,7 +6,8 @@ export const RNG_INDEX_RAY_JITTER = 0;
 export const RNG_INDEX_ENVIRONMENT_SAMPLE = 1;
 export const RNG_INDEX_SCATTER_TYPE = 2;
 export const RNG_INDEX_SCATTER_DIRECTION = 3;
-export const RNG_INDEX_ALPHA_TEST = 4;
+export const RNG_INDEX_APERTURE_SAMPLE = 4;
+export const RNG_INDEX_ALPHA_TEST = 5;
```
gkjohnson: I'm wondering if we should include a set of "user-space" indices so the random functions can be used without conflicting with internal ones. "RNG_INDEX_ALPHA_TEST" could start at 30 or something so 4 to 29 would be free to use

TheBlek: Yeah, that sounds good. Especially since custom functions for everything is what we a moving towards. We could name the first custom index and guarantee a certain number of them.

TheBlek (approved):
Looks amazing!

src/webgpu/nodes/random.wgsl.js
gkjohnson: I've adjusted the alpha test index to start at 1000, which seems like a good amount. We can adjust or add more named indices as we go

PR #778, opened 2026-07-06
WebGPU: Add support for anisotropic materials

gkjohnson:
- Adds support for anisotropic settings from MeshPhysicalMaterial based on the glTF model.
- Adjusts the BXDFContext to store "H", "L", and "V" directly instead of individual dot products. Guards are now apply in-place, instead.
- Adjust the viewer test page to default to 1-to-1 pixel scale by default
- Fix WebGPUPathTracer displaying stale content when disabling and enabling it
- Tried to update some of the comments and references in the ggx file

| before | after | three.js |
|---|---|---|
| image | image | image |

### DIFF
--- a/src/webgpu/nodes/ggx.wgsl.js
+++ b/src/webgpu/nodes/ggx.wgsl.js
@@ -85,7 +85,8 @@ export const ggxLambdaFunc = wgslFn( /* wgsl */ `
 		let alphaT = alpha.x;
 		let alphaB = alpha.y;
 
-		let cos2 = V.z * V.z;
+		let NdotV = max( V.z, MIN_INCIDENT_COS );
+		let cos2 = NdotV * NdotV;
 		let t = ( alphaT * alphaT * V.x * V.x + alphaB * alphaB * V.y * V.y ) / cos2;
 		let numerator = - 1.0 + sqrt( 1.0 + t );
 		return numerator / 2.0;
@@ -114,8 +115,8 @@ export const ggxSmithVisibilityFunc = wgslFn( /* wgsl */ `
 		let alphaT = alpha.x;
 		let alphaB = alpha.y;
 
-		let NdotV = V.z;
-		let NdotL = L.z;
+		let NdotV = max( V.z, MIN_INCIDENT_COS );
+		let NdotL = max( L.z, MIN_INCIDENT_COS );
 
 		let TdotV = V.x;
 		let TdotL = L.x;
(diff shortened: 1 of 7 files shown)


src/webgpu/nodes/structs.wgsl.js
```diff
@@ -206,13 +216,15 @@ export const lobeWeightsStruct = new StructTypeNode( {
...
-	VdotNc: 'float',
-	LdotNc: 'float',
-	HdotNc: 'float',
+	V: 'vec3f', // view dir
+	L: 'vec3f', // light dir
+	H: 'vec3f', // half dir
+
+	Vc: 'vec3f', // clearcoat view dir
+	Lc: 'vec3f', // clearcoat light dir
+	Hc: 'vec3f', // clearcoat half dir
```
TheBlek, line 226: Why did you decide to pass around vectors instead of floats? When writing bxdf context struct my thinking was that we can do clamping once and reuse consistently across all computations.

gkjohnson: The anisotropic BRDF needs the view and light dot product with the tangent, bitangent, and normal, so it basically would have amounted to adding every dimension for these vectors as floats. So I just collapsed them back to the original vec3s to avoid adding 9 arguments to some of these functions.

We could clamp individual dimensions of these vectors by the epsilons before passing them into the functions but it also made sense to me for every function to be responsible for its own epsilons. I'm open to changing this, though.

src/webgpu/nodes/ggx.wgsl.js
```diff
@@ -55,82 +72,97 @@ export const ggxDirectionFunc = wgslFn( /* wgsl */ `
...
 export const ggxShadowMaskG1Func = wgslFn( /* wgsl */ `
 
-	fn ggxShadowMaskG1( cosTheta: f32, alpha: f32 ) -> f32 {
+	fn ggxShadowMaskG1( V: vec3f, alpha: vec2f ) -> f32 {
 
-		let a2 = alpha * alpha;
-		let cosTheta2 = cosTheta * cosTheta;
-		let denom = cosTheta + sqrt( cosTheta2 * ( 1 - a2 ) + a2 );
-		return 2.0 * cosTheta / denom;
+		return 1.0 / ( 1.0 + ggxLambda( V, alpha ) );
```
TheBlek, line 100: I believe this can be more efficiently written in the old form, just replacing denom with following:
```
let denom = cosTheta + sqrt( cosTheta2 + alphaT * alphaT * V.x * V.x + alphaB * alphaB * V.y * V.y );
```

I can write up the transitions from ggxLambda to the form of ggxShadowMaskG1 if you want.

src/webgpu/nodes/structs.wgsl.js
TheBlek: > it also made sense to me for every function to be responsible for its own epsilons

This way it could be easier to read I guess. I'm ok with this change but it will require a bit more attention when hardening against edge cases

TheBlek (approved):

src/webgpu/nodes/ggx.wgsl.js
gkjohnson: I changed it to this form since it just made it easier for me the trace it when comparing it to the references I was reading. I'll add a comment noting that we can adjust this later but I'm thinking it might be good to keep the verbose forms at least until things like energy conservation is handled fully.

PR #781, opened 2026-07-09
WebGPU: Add support for dynamic features

gkjohnson:
Add an "updateMaterials" function to WebGPUPathTracer that enables quickly updating material content only for scene edits. Texture atlas is regenerated if new texture content is present. "updateTransforms" has also been added in three-mesh-bvh to support adjusting matrices.

Fast dynamic mesh updates like skinning and other geometry will come later.


gkjohnson:
Colors and material properties can now be adjusted in real time without reconstructing the full scene:

https://github.com/user-attachments/assets/bb37a51c-a59e-442b-8be2-146d2b77ca68

PR #786, opened 2026-07-14
WebGPU: Add new random sampling strategies

gkjohnson:
This PR adds some new random sampling strategies using spatiotemporal blue noise (from [here](https://github.com/NVIDIA-RTX/STBN)) and one based on the "stratified" sampling strategy from WebGLPathTracer. The "stratified" strategy more or less uses a single sequence of  scrambled sobol numbers across the whole screen at once but uses screen-space blue noise offset per pixel to adjust it, which results in a pixels using common sample sampling directions tiled across the screen.

The STBN strategy uses the raw blue noise textures up until the last temporal slice (64) and then falls back to sobol. This isn't fully implemented and requires loading larger 3d textures so I'm inclined to not include this one. But I wanted to check if was a significant improvement but it seems it is not compared to sobol and adds some complexity and memory overhead.

It also:
- Adjusts the wavefront ray generation kernel to use the same seed as the megakernel (use sample count as the seed.
- Adjust the traversal algorithm to increment the random-sample dimension only when encountering a transparent surface rather than for every triangle. This will let us keep the required dimensionality down significantly.
- Move sobol + pcg to separate files for readability (random.wgsl.js now just reexports from these files)

Here are some screenshots at different sample counts:

| | PCG | Sobol | STBN | Blue Dither |
|---|---|---|---|---|
| 1 sample | image | image | image | image |
| 6 samples | image | image | image | image |
| 32 samples | image | image | image | image |
| 500 samples | image | image | image | image |

You can see that the "blue noise dither" approach looks smoother _very_ quickly but it will result in some "off" looking results due to the consistency in samples across the screen (you can see by flip booking between "blue dither" and one of the others). The above uses a 64^2 blue noise offset texture. I'll have to see if scrambling some of the other intermediate dimension values gives us a better effect. But for simple scenes the appearance of resolving so quickly could be useful for some use cases, though I don't think we should use this as the default or for our testing.

[This Arnold paper on blue noise dithering](https://blogs.autodesk.com/media-and-entertainment/wp-content/uploads/sites/162/dither_abstract.pdf) is relevant.


gkjohnson:
I've removed the STBN randomness since I didn't think it added a whole lot - and I've adjusted the default random sampler back to sobol. I think the bluedither random sampling will have some value especially at low sample counts so maybe its some thing we can add as a user-option in the future.

gkjohnson:
There's also the ZSobol sampler from PBRT which may be worth looking into at some point:

https://pbr-book.org/4ed/Sampling_and_Reconstruction/Sobol_Samplers#BlueNoiseSobolrsquoSampler

gkjohnson:
Merging since this doesn't change significant functionality aside reorganization & raytraversal transparency increments. Sobol will continue to be the default random function.

Note this requires a three-mesh-bvh update.

Issue #777, opened 2026-07-05
WebGPUPathTracer: Remaining features


gkjohnson:
Migrating the list of remaining work for WebGPUPathTracer from https://github.com/gkjohnson/three-gpu-pathtracer/pull/713 to a dedicated issue. cc @TheBlek feel free to edit and anything you think should be worked on / mark if you're planning to / ark working on something.

### Features
- Clamp based on the camera far plane, aligned with three.js rasterization (primary ray only)

### Performance
- Allow for tweaking budget mid-render (adjust requirement to preprocess?) (also use for low res mode, up max to to full frame size)

### Validation
- Inconsistency in triangle handling (missing triangles in PT only) - see the stormtrooper model. Possibly related to generation with no uvs?
  - Lone monk models also missing geometry
  - Due to shared index attribute
- Leaf cluster size overruns (high LoD rover model)
- Fix handling of ior = 1.0 or reflection, improve bookkeeping of f90 on the surface record and stop assuming 1.0
  - Setting ior to 1.0 can seemingly cause NaN errors
- Soap Bubble material fails (See material orb)
  - NaN errors
- Generate new viewer test screenshots
- GLTF Fidelity Models with issues:
  - x-DispersionTest (KHR_materials_dispersion, NaNs)
  - x-IORTestGrid (NaN)
  - x-SunglassesKhronos (NaN)
  - x-TransmissionThinwallTestGrid (NaN)
  - x-MeshOptCubeTest (three.js failure?)

### Demos
- Remove scaled setting
- Material Orb
  - soap bubble ior + iridesence handling seems different compard to blener
- Clean up demos
  - Remove as many options as possible in most demos
  - Improve scene / camera position quality
  - Remove unneeded HDRI
  - Find a "Hero" model
  - Denoiser on more demos
- Demos to update
  - LKG (use a single model, get it working with LKG)
  - Fog (get it working, new model)
  - RenderVideo (update model? Change background? Improve look)
  - SkinnedMesh (remove?)
- Denoise / Upscaler
  - Upscale support for transparent bg
    - Requires changes to three.js FSR1Node (changeable texture) or pmndrs/upscaler (pass through alpha)
- Update example renders
- Fix / remove remaining demos
  - Fog (depends on new feature)

### Other
- JSDoc Documentation
- Ensure WebGPUPathTracer handles "resets" consistently and correctly.
- Reorganize PathTracingMaterial
- New README screenshots
[...]

PR #787, opened 2026-07-14
Sample UI: Throttle sampling UI updates

Ryan-G7:
This PR introduces a throttling mechanism for the HTML sampling UI updates and makes the throttle interval fully configurable (defaulting to `100ms`).

Currently, updating the DOM on every single frame can cause unnecessary CPU overhead and potential layout thrashing, especially during high-frame-rate rendering or on low-end devices. 

By throttling these UI updates, we can significantly reduce DOM manipulation frequency without compromising the user experience. This aligns with standard industry practices; for instance, Blender employs a similar throttled/interval-based approach for updating its viewport sample counters.

I'm not sure if you consider this completely necessary...

### DIFF
--- a/example/utils/LoaderElement.js
+++ b/example/utils/LoaderElement.js
@@ -104,6 +104,11 @@ export class LoaderElement {
 		this._credits = creditsEl;
 		this._samples = samplesEl;
 		this._container = container;
+		this._lastSampleUpdateTime = - Infinity;
+
+		// Keep path tracing independent from DOM updates. This can be tuned by hosts
+		// that need a more or less responsive progress display.
+		this.sampleUpdateInterval = 100;
 
 		this.setPercentage( 0 );
 
@@ -144,6 +149,15 @@ export class LoaderElement {
 
 	setSamples( count, detailedSamples = null ) {
 
+		const now = performance.now();
+
+		// Rendering can run every animation frame, but updating text at that rate
+		if ( now - this._lastSampleUpdateTime < this.sampleUpdateInterval ) {
+
+			return;
+
+		}
+
 		if ( detailedSamples !== null ) {
 
 			const {
@@ -167,6 +181,8 @@ export class LoaderElement {
 			this._samples.innerText = `~${ Math.floor( count ) } samples`;
 
 		}
+
+		this._lastSampleUpdateTime = now;
 
 	}
 


gkjohnson:
Hi @Ryan-G7 - thanks for the contribution. Do you have any examples of the problematic overhead? I've never seen modification of a single div entry be a performance bottleneck - especially in such a simple page.

Ryan-G7:
Thanks for the quick response!@gkjohnson

You are absolutely right. In standard scenarios (e.g., lower frame rates or heavy scenes where path tracing itself is the major bottleneck), the DOM overhead is indeed negligible.

However, I observed this issue in a high-performance environment:

On a high-end machine (RTX 5090) with an animation loop running at around 180 FPS, the absolute frequency of DOM updates becomes quite high. According to the Chrome Performance Profiler, the constant layout recalculation and reflow/repaint of these rapid DOM changes do introduce measurable layout overhead (apologies as I cannot share the screenshots from my internal intranet environment).

Even setting the performance aspect aside, having the numbers flashing and jumping 180 times per second is visually distracting and hard for the human eye to read.

Let me know what you think!

gkjohnson:
> DOM changes do introduce measurable layout overhead (apologies as I cannot share the screenshots from my internal intranet environment).

I need to ask for something concrete. None of these examples are CPU-bound at all since the real work is all GPU side and all of my measurements show the layout is minimal. Unless there's some demonstrable impact I'm not inclined to add micro optimizations to the example.

> having the numbers flashing and jumping 180 times per second is visually distracting and hard for the human eye to read.

Which example are you referring to? The sample counts are whole numbers and are not flickering or raising rapidly on any machine in my experience.

I appreciate you interest in the project but if you'd like to contribute I ask that you be specific and concrete otherwise I have to spend my time guessing at what issues you're actually trying to solve.

Ryan-G7:
Apologies for the lack of clarity earlier!

I observed this in the viewer example with a simple scene. the sample count climbs and flashes so rapidly (as can be seen in the video).  

However, I completely agree with your point, in simple scenes, performance is already too high to cause lag, and in complex scenes, the GPU bottleneck makes this DOM overhead negligible.

Since this is indeed a micro-optimization, please feel free to close the PR if you feel it's unnecessary. Thanks again for your time and for this great project!

https://github.com/user-attachments/assets/506befd0-a8bc-44d5-82a8-5008bc2c491e

gkjohnson:
Thanks for the video - I think it's useful to have the rate which numbers are climbing especially during development, so I don't think this is a change worth making at the moment.

If you'd like to help update some of the other examples, though, that would be appreciated. The following exmaples listed in #777 could use some attention:

- Render Video example currently does not work with WebGPUPathTracer
- Rename the webgpu_primitives.js example to just "primitives"
- The "basic.js" example is not currently rendering

Issue #784, opened 2026-07-12
Energy Conservation Plan


gkjohnson:
I've started to think through the energy conservation issue for the furnace test and I think I've started to wrap my head around things again after being in the BRDF code a bit. My understanding is that we need a few dedicated `attribute by view angle` energy integral tables per lobe / term that has has energy loss due to lack of multi-scatter modeling as well as for fresnel. 

As far as I understand that will be a dedicated table for the following cases:

- GGX multiscatter energy loss
- Diffuse lobe energy loss (if the diffuse term is not energy conserving)
- Possibly sheen energy loss table if using a dedicated lobe? (as [OpenPBR-BSDF repo](https://github.com/adobe/openpbr-bsdf/blob/main/impl/openpbr_fuzz_lobe.h#L457) does)
- Baked fresnel term based on the GGX scatter to use for scaling the energy used in the specular and diffuse terms in opaque dielectrics (possibly mergeable wit the transmissive fresnel tables below)
- Fresnel term for every transmissive ior within a reasonable range

So then the energy loss terms can be used to scale the energy of any lobes that lose energy due to unmodeled factors like multiscatter. And the fresnel terms can be used to query the energy scattered by the specular and diffuse / transmissive layers.

--

From a technical perspective I'm thinking we should shift from pre-baked textures (like the turquin one right now) to a dynamically generated one that lives in 3D Texture or Array Texture, which will ensure the results are consistent as we change BRDFs etc. We can include as many layers here as we want, generating a block of layer for transmissive lobes over some reasonable ior range to sample in addition to the others.

I'm also thinking we switch to a basic cosine lambertian diffuse lobe for now rather than the Disney diffuse term since it's not energy conserving.

@TheBlek How does this sound to you? Am I thinking about this right? It's been awhile since I've worked on this quite so in depth.

TheBlek:
Generally, yes, this sounds about right.

> - Baked fresnel term based on the GGX scatter to use for scaling the energy used in the specular and diffuse terms in opaque dielectrics (possibly mergeable wit the transmissive fresnel tables below)
> - Fresnel term for every transmissive ior within a reasonable range

[...]

PR #788, opened 2026-07-17
WebGPU: Improve energy conservation, use lambertian diffuse

gkjohnson:
Related to #777, #784 

- Adjusts the "furnace_test" example to expose options for adjusting axes to `'metalness', 'roughness', 'iridescence', 'clearcoat', 'transmission', 'opacity', 'none'` for testing conservation of different BRDFs.
- Switch the diffuse BRDF to use a Lambertian cosine BRDF for the sake of evaluating energy conservation of the existing layers (we can replace this with an EON diffuse BRDF later)
- Pulls the energy compensation boost from the turquin texture to the GLTFCompliantMaterial out of the conductor fresnel function (its now used for iridescent conservation, as well).
- Improve energy conservation for dielectric specular BRDF, iridescent BRDF, and clearcoat BRDF.
- Small lint & spacing adjustments

| | Before | After |
|---|---|---|
| roughness x metalness |  |  |
| roughness x iridescence (dielectric) |  |  |
| roughness x clearcoat roughness |  |  |

--

You can see in the dielectric cases that there is a brightening around the rim of the low-roughness dielectric case. My assumption that this is related to the fresnel and the cosine approximation not lining up with the GGX microfacet distribution somewhere:

image


src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -94,29 +97,39 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
...
 				let NdotL = ctx.L.z;
 
+				// account for multi scatter energy loss for specular
+				let energySs = max( textureSampleLevel( ${ this.turquinNode }, ${ this.turquinSampler }, vec2f( NdotV, surf.roughness ), 0 ).r, 1e-5 );
 				let specular = ${ this.specularBrdf }( ctx.V, ctx.L, ctx.H, alpha );
 				let diffuse = ${ this.diffuseBrdf }( NdotV, NdotL, ctx.VdotH, surf );
-				let dielectricBase = ${ this.fresnelMix }( ctx.VdotH, surf.ior, diffuse, specular );
+
+				// Turquin multiscatter energy compensation term: 1 + f0 * ( 1 - E ) / E
+				// This simplifies to 1.0 / E for dielectric (f0 = 1.0)
```
TheBlek: Why is f0 = 1.0 here?

In the paper I'm seeing this formula `1.0 / E` for LUT with baked in fresnel calculation. Shouldn't we reuse formula for the metals until we get to transmission and baking 3D LUT?

gkjohnson: The idea was boost specular to reflect 100% energy and then blend with the energy conserving diffuse layer using fresnel - but I agree it doesn't feel right. I've made some further changes that move it in the direction of the paper for transmissive energy conservation - but I have some questions maybe you'll have a better sense for (outlined below)

gkjohnson:
I've made some changes that align my impression of the Turquin approach, though the paper doesn't seem as explicit about handling transmissive nor diffuse surfaces:

I've moved the texture generation to a dedicated "TurquinTexture" instance that includes 1 layer for conductor energy compensation & 10 layers for dielectric ior 1-2.5, which includes the fresnel in the value. This should be the total energy reflected toward a specific direction from the specular lobe.

I had some trouble understanding how to apply the approach from the paper to handle diffuse so I've effectively multiplied the the diffuse contribution by the the complement of the scattered specular energy so that `spec_energy + diff_energy = 1`, similar to what I alluded to in #784, which gets the furnace test passing (clearcoat still needs to be updated):

image

--

So it's passing but it doesn't feel complete and is missing any compensation for specular. I'm thinking it might be best to boost the specular using the same conductor formula (`1 + f0 * ( 1 - Ess ) / Ess`) and then account for that in the diffuse attenuation? Any thoughts @TheBlek?

gkjohnson:
Okay I made some of the changes I alluded to above because I couldn't leave it alone 😅 Dielectric specular is now boosted to account for multiscatter using the same formula as that used for metals and an f0 derived from a 1.5 ior everywhere (as defined by the gltf spec). Clearcoat has been updated to use the same pattern and is also passing the furnace test.

I'm going to put this down until it can be reviewed and assuming this looks good I'm thinking next steps would be:
- Transmission support
- Delete Disney-related functions, replace lambertian diffuse with EON
- Organize the lobe evaluation so it's more trace-able as a set of "layers".

gkjohnson:
@TheBlek - I'm going to merge this for now since the results seem to be good. I have some work I'd like to use this project for so I'm planning to try to get this to a useable first version and we can make more improvements from there. So apologies for moving quickly on these things.

Regarding disney BRDF vs lambertian - I'm okay switching to lambertian for now since this is what glTF declares and the energy preservation is something I want to be able measure more cleanly. And we can add EON support later which should be a better model anyway.

My plans for some upcoming PRs related to energy:
- Delete disney-diffuse related lobe functions
- Organize BRDF composition into layers so energy is easier to trace
- Handle transmission energy compensation

PR #789, opened 2026-07-18
WebGPU: Update the turquin texture generation

gkjohnson:
Related to #784 

Updates turquin texture generation since I saw it as broken after some recent changes:
- Reduce samples from `1024**2` to `64**2`
- Use a grid-sampling pattern rather than random samples
- Fix construction of texture by removing random calls, which had broken with the switch to sobol from PCG random sampling
- Checking the GPU timing - the time-to-generate is much lower on my machine, now (a few milliseconds). So we should be able to switch to on-the-fly generation comfortably.
- Discard samples below the surface.

Note: changes are easier to read with [white space changes disabled](https://github.com/gkjohnson/three-gpu-pathtracer/pull/789/changes?w=1).


src/webgpu/nodes/material.wgsl.js
```diff
@@ -555,50 +554,64 @@ export const fresnelCoatFunc = wgslFn( /* wgsl */ `
...
 
-		const INTEGRATION_SAMPLES = ( 1 << 20 );
-		let pixelIndex = ( globalId.x << 16 ) | globalId.y;
-		${ rngInit }( pixelIndex, 0, 0 );
+		// sample the brdf directions in a grid pattern
+		const GRID_SIZE = 64u;
 
+		// TODO: this sampling means that energy at 0.0 & 1.0 roughness (and 0 and 90deg cos) are never
+		// written to the texture due to the half texel inset, resulting in small, though possibly noticeable,
+		// error in common cases.
```
gkjohnson, line 569: @TheBlek one small potential issue I noticed is that due to the half texel inset we never actually write the energy value for roughness = 0.0 or 1.0. This amounts to an error of ~0.016 at a resolution of 32x32, which may be negligible but something to bear in mind if we see some issues at these common values,

src/webgpu/nodes/material.wgsl.js
```diff
@@ -555,50 +554,64 @@ export const fresnelCoatFunc = wgslFn( /* wgsl */ `
...
 		texture: texture_storage_2d<r16float, write>,
-
 		globalId: vec3u,
 	) -> void {
 
-		const INTEGRATION_SAMPLES = ( 1 << 20 );
-		let pixelIndex = ( globalId.x << 16 ) | globalId.y;
-		${ rngInit }( pixelIndex, 0, 0 );
+		// sample the brdf directions in a grid pattern
+		const GRID_SIZE = 64u;
```
TheBlek, line 565: Should this be 642? PR description implies this.

Is this enough to integrate to a similar precision?

gkjohnson: Sorry that was a markdown mistake - I intended to write `1024**2 to 64**2` which rendered as "1024**2 to 64**2", bolding the characters between the exponent notation.

gkjohnson:
I just merged the main webgpu dev branch in and re-added a `vec2u` value cast that I'd forgotten to commit previously - testing the furnace page now with the dynamically generated texture looks good:

| Before | After |
|---|---|
| image | image |

TheBlek (approved):

PR #791, opened 2026-07-19
Update webgpu examples

Ryan-G7:
### Description

In response to the requirements in #787, I am glad to contribute to this project. This PR updates the example files to support the WebGPU path tracer and resolves some issues regarding video rendering.

#### Basic
* Updated the pipeline to utilize the current synchronous `setScene()` API to maintain consistency in scene loading.

#### Primitives
* Removed the original `primitives` file and renamed `webgpu_primitives`.

#### Video Rendering (renderVideo)
* Removed the WebGL-specific `preserveDrawingBuffer` option, which is unsupported in WebGPU.
* Created bounding boxes for `SkinnedMesh` before the BVH update to avoid errors when `target.makeEmpty()` is executed in the `getPreciseBounds` method.
* Fixed a division-by-zero exception that occurred when `fadeDuration` was set to 0.
* Added a 100ms render delay during preview mode, and automatically disabled this delay during active recording.

### Questions
* I noticed an issue where the preview does not update during camera interaction when `renderDelay` is set to 0. I would like to know if, under ideal conditions, the preview should maintain full-resolution rendering during camera interaction?

Currently, all of these examples are working properly. Please let me know if you have any feedback or suggestions regarding these modifications.


Ryan-G7:
Additionally, I implemented direct and indirect light clamping, as well as `filterGlossy`, in an attempt to resolve the firefly artifact issues. However, I found that the noise was not caused by this. If you are interested, I can open a separate PR for these changes.

gkjohnson:
Thanks! This is great - a few comments:

- I see that the "samplesPerFrame" option in "renderVideo" currently isn't doing anything. We should adjsut this setting so it calls "renderSamples" a that amount of times.
- Settings "samples" to "1" in the render video example results in no frame other than the first preview frame being rendered.
- When rendering the video I'm seeing that it keeps switch back to the original low res preview frame before showing the current frame (likely related to your question) - setting "minSamples" to "0" fixes this
- Lets keep the fade in the render view scene when generation is not happening (we can just remove the line where the duration is set to 0)

To your questions:

> I noticed an issue where the preview does not update during camera interaction when renderDelay is set to 0. I would like to know if, under ideal conditions, the preview should maintain full-resolution rendering during camera interaction?

This is a good question - I guess the problem is that when "renderDelay" is set to zero that the preview frame doesn't update and therefore becomes stale? I guess we should at least make sure the low res preview frame is up to date if we're going to display it. That means if `minSamples !== 0 || renderDelay !== 0` then we need to update the preview frame.

> Additionally, I implemented direct and indirect light clamping, as well as filterGlossy, in an attempt to resolve the firefly artifact issues. However, I found that the noise was not caused by this.

The general noise won't be addressed by the filter glossy factor - this is specifically designed to address a specific type of ray that hits a diffuse, then specular surface, then bright light source. You can see blender's write up [here](https://docs.blender.org/manual/en/latest/render/cycles/optimizations/reducing_noise.html#caustics-and-filter-glossy). This will probably be more noticeable once NEE is supported but I think we'll need a demonstration scene to evaluate how well any given implementation is working.

Ryan-G7:
Thanks for the feedback! I've made the following adjustments to address those issues:

> * I see that the "samplesPerFrame" option in "renderVideo" currently isn't doing anything. We should adjsut this setting so it calls "renderSamples" a that amount of times.

samplesPerFrame now controls the number of times renderSample() is called within each RAF.

> * Settings "samples" to "1" in the render video example results in no frame other than the first preview frame being rendered.

After fixing the preview update issue when samples = 1, subsequent animation frames now update correctly. However, I noticed that it still only records the preview frame. Keeping minSamples at 0 during recording resolves this.

To fix the preview not updating when renderDelay = 0, I am now using `this._resetTime === -1` to determine if it is the first frame. After each reset, if `this._resetTime === -1` and `minSamples !== 0`, it ensures that the low-resolution preview updates at least once.

> * Lets keep the fade in the render view scene when generation is not happening (we can just remove the line where the duration is set to 0)

Removed fadeDuration = 0.

gkjohnson:
Thanks! I'll merge this - if there's anything else you were interested in working on let me know. A few other things that I've noticed that could use some attention:

- Toggling to the "orthographic" rendering mode in the main "index.js" example
- Trying to render the "khronos-RecursiveSkeletons" demo file in the "viewerTest" example seems to cause the gpu device to crash but it's not exactly clear to me why (see `http://localhost:5173/viewerTest.html#khronos-RecursiveSkeletons`)
- In the "renderVideo" example I'm seeing a bunch of errors related to LZWEncoder on load, which would be good to figure out. I've also noticed that the time to write a frame is very slow - not sure if there are any other more modern options for doing WebGPU recording like this now nowadays.

PR #794, opened 2026-07-21
WebGPU: Fix package-lock to use registry three-mesh-bvh

TheBlek:
Point lockfile's three-mesh-bvh at the npm registry (`0.9.13`) and normalize the dep to `^0.9.13`
Otherwise it does not work without local three-mesh-bvh installed


gkjohnson:
Oops 😅 Thanks!

PR #795, opened 2026-07-23
Fix webgpu examples

Ryan-G7:
Hi! I have resolved some of the issues with the examples. Next, I will look into the remaining example issues in #777, and after that, I will probably take a look at the self-intersection and firefly noise issues.

**Fix orthographic camera switching issue:**
* reusing  `inverse view-projection` uniform in the pipeline to ensure that the matrix data correctly synchronizes and updates to the pipeline after a camera switch.

**Fix pipeline validation error in `RecursiveSkeletons`:**
* In the `RecursiveSkeletons` model, `COLOR_0` and `JOINTS_0` share interleaved `Uint8Array` data. When the Three.js WebGPU backend processes unnormalized joint indices, it expands them into a `Uint32Array`. This inadvertently causes the color format to be misidentified as `unorm32x4`, triggering a validation crash in `createRenderPipeline`. I was also able to reproduce this issue by loading the model with a standard `GLTFLoader`.
* For now, I have split these attributes apart. I plan to address this issue in three.js later.

**Migrate from `canvas-capture` to `Mediabunny`:**
* I was unable to reproduce the error you mentioned on my machine, but I agree that replacing this library is a wise choice. I have completely removed `canvas-capture` from `renderVideo`, replaced it with the modern WebCodecs-based `Mediabunny` library, and updated `package.json` accordingly.

Added the `getMaxDeviceLimits` method to request a `GPUDevice` based on the maximum limits supported by the GPU, and integrated it into the example `index`.


example/index.js
```diff
@@ -30,6 +30,7 @@ import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
 import { getScaledSettings } from './utils/getScaledSettings.js';
 import { LoaderElement } from './utils/LoaderElement.js';
 import { LDrawConditionalLineMaterial } from 'three/addons/materials/LDrawConditionalLineMaterial.js';
+import { getMaxDeviceLimits } from '../src/webgpu/index.js';
```
gkjohnson: Lets move this function to the examples folder. I don't think it's generally great practice to blindly request the maximum buffer limits so I'm hesitant to include this at all but we can add it with a comment noting that it should be improved later.

example/viewerTest.js
```diff
@@ -453,9 +454,22 @@ async function updateModel() {
...
+		// and non-normalized attributes share an interleaved buffer then expanding one changes
+		// the format of the other, so separate them before the raster and path tracing renderers
+		// access the geometry.
+		if ( c.geometry && ! geometries.has( c.geometry ) ) {
+
+			geometries.add( c.geometry );
+			deinterleaveGeometry( c.geometry );
+
+		}
+
```
gkjohnson: If this is a three.js / WebGPURenderer issue then I think it's okay to just submit a three.js issue and assume it will be fixed later. I don't want to add a bunch of workarounds for three.js issues here that we'll have to remove later if we can avoid it.

gkjohnson (commented):
Thanks for the investigation! I've added a few comments above.

A couple other things I've noticed - I'm seeing this UV issue in the log of the record video example. I don't think it's new necessarily? But would be good to know where it's from if you have any ideas:

Image

I'm also seeing that the video quality using the mediabunny encoder is significantly lower than it was previously. I'm wondering if there's any way to improve the quality further? I'm not sure if this quality is something we want to use to showcase the project capabilities

Image

example/viewerTest.js
Ryan-G7: I will address these issues today.

Ryan-G7:
Hi, I have resolved the following issues:

- Fixed the missing UV warning caused by `turquinTexture`.
- Improved the quality of videos recorded with Mediabunny.
- Moved the device limit requesting method to the example, and now only request the configured options.
- Regarding the `RecursiveSkeletons` issue, I fixed it in Three.js yesterday, which should be resolved once r186 is released.

Ryan-G7:
Related to #777 

> WebGPU Primitives logs errors when toggling mega kernel, then toggling enabled

I couldn't reproduce it. The missing uv warning when switching to the mega kernel is fixed in this PR. Could you check if it's still happening on your end?"

example/index.js
```diff
@@ -134,6 +145,43 @@ async function waitFrame() {
...
+ */
+function getRequiredDeviceLimits( adapter ) {
+
+	if ( ! adapter || ! adapter.limits ) {
+
+		throw new TypeError(
+			'getRequiredDeviceLimits: A valid GPUAdapter is required.'
+		);
+
+	}
```
gkjohnson: Lets remove this - I don't think it's necessary to check types everywhere and I think it's best to keep thing as simple as possible here

example/index.js
```diff
@@ -134,6 +145,43 @@ async function waitFrame() {
 
 }
 
+/**
+ * Returns required GPU limits according to DEVICE_LIMITS_REQUESTED.
+ *
+ * Only limits explicitly listed in DEVICE_LIMITS_REQUESTED will be requested.
+ * This avoids requesting all adapter limits and improves compatibility.
```
gkjohnson, line 152: Lets add a note indicating that limits should to be set based on the requirements of the application rather than just requesting the maximum values.

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -48,7 +48,7 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 		this.turquinTexture.minFilter = LinearFilter;
 		this.turquinTexture.magFilter = LinearFilter;
 
-		const turquinNode = texture( this.turquinTexture ).setName( 'turquinTexture' );
+		const turquinNode = texture( this.turquinTexture, vec2( 0.0 ) ).setName( 'turquinTexture' );
```
gkjohnson, line 51: So this change is required to remove the "uv" console logs? Is this an issue with three.js / TSL? It seems like it should be valid to instantiate a texture node without a uv.

gkjohnson (commented):
It looks great now, thanks! I've added a few final comments.

Also just to reference it - it looks like this is the PR that fixed the skinned mesh issue: https://github.com/mrdoob/three.js/pull/34098

src/webgpu/materials/GltfCompliantMaterial.js
Ryan-G7: Yes, you're right. This was done to eliminate the UV console log, and I actually think it is a current limitation of TSL.

When instantiating a texture node, TSL implicitly binds a default uv() node to it. While this works perfectly fine in a standard rasterization pipeline, it fails and throws warnings in contexts where UVs need to be calculated manually, such as in Compute Shaders or path tracing.

I feel that, currently, TSL somewhat conflates resource declaration with sampling behavior when handling textureNode.

Ryan-G7: Do you think we should fix this in Three.js?

gkjohnson: > I feel that, currently, TSL somewhat conflates resource declaration with sampling behavior when handling textureNode.

I agree. I think this is happening in a lot of places across TSL at the moment.

> Do you think we should fix this in Three.js?

I think it's worth at least making an issue

gkjohnson:
Thanks!

> I couldn't reproduce it. The missing uv warning when switching to the mega kernel is fixed in this PR. Could you check if it's still happening on your end?"

I wasn't able to reproduce it so I think this is addressed.

PR #796, opened 2026-07-24
WebGPU: Add support for NEE, MIS

gkjohnson:
Related to #777, #770

- Add support for MIS, NEE to both backends based on the WebGL version. 
- renames "object" variable to "objectInfo" to avoid conflict with built-in three.js uniform group naming.
- Moves background sampling, variables to a dedicated class.

**TODO**
- Review


gkjohnson:
Before / After of a single sample:

| Before | After |
|---|---|
| image | image |

This _has_ made this stochastic opacity artifact apparent, which I hadn't noticed before. This seems to be occuring because the random dimension "effect" being used is based on the intersection order of the triangles, since that's was causing the effect to increment and the intersection order isn't consistent between rays. I'll add this to the list to figure out.

image

--

cc @TheBlek - just FYI that I'm trying to get all the basic features ported over from the WebGL version, which will make it easier to then focus on performance I think. I'll plan to look into getting lighting working next.

I am noticing rendering becoming a bit less responsive (as you might expect). I think that can be addressed by how we iterate over the wave front work but from #770 think you have a better handle on that than me.

PR #797, opened 2026-07-27
Add dynamic scene editing support for WebGPU

Ryan-G7:
Related to #777 

This PR adds support for editing a WebGPU path-traced scene at runtime and introduces a `dynamicScene` example demonstrating the complete workflow.

The example allows users to create, select, transform, delete, and edit the materials of asset instances while keeping the path-traced result synchronized.

## Changes

- Add `WebGPUPathTracer.updateTransforms()` for lightweight transform-only updates.
- Complete resource cleanup in `WebGPUPathTracer.dispose()`, including BVH data, texture atlases, and environment/background caches.
- Add a new `dynamicScene` WebGPU example featuring:
  - model asset instancing with shared geometry and textures;
  - translate, rotate, and scale controls;
  - object deletion;
  - interactive editing of common physical material properties;

Transform and material changes are coalesced to at most one update per animation frame. 

## Controls

- Click Add/Delete to manage an instance.
- Click an object to select it.
- Press `T`, `R`, or `S` to switch between translate, rotate, and scale modes.

### DIFF
--- a/src/webgpu/WebGPUPathTracer.js
+++ b/src/webgpu/WebGPUPathTracer.js
@@ -279,6 +279,13 @@ export class WebGPUPathTracer {
 		bvhData.update();
 		bvhData.textureAtlas.setTextures( this._renderer, bvhData.textures );
 
+		if ( this._bvhData ) {
+
+			this._bvhData.dispose();
+			this._bvhData.textureAtlas.dispose();
+
+		}
+
 		this.scene = scene;
 		this._bvhData = bvhData;
 		this._pathTracer.setBVHData( bvhData );
@@ -360,6 +367,14 @@ export class WebGPUPathTracer {
 		const { _bvhData, _renderer } = this;
 		_bvhData.updateMaterials();
 		_bvhData.textureAtlas.setTextures( _renderer, _bvhData.textures );
+		this.reset();
+
+	}
+
+	updateTransforms() {
+
+		this.scene.updateMatrixWorld( true );
+		this._bvhData.updateTransforms();
 		this.reset();
 
 	}
@@ -660,10 +675,12 @@ export class WebGPUPathTracer {
 	dispose() {
 
 		this._pathTracer.dispose();
+		this._bvhData.dispose();
+		this._bvhData.textureAtlas.dispose();
+		this._environmentCache.dispose();
+		this._backgroundCache.dispose();
 		this._blitQuad.dispose();
 		this._lowResTarget.dispose();
-		this._envColorTexture.dispose();
-		this._backgroundColorTexture.dispose();
 		this._atlasDebugQuad?.dispose();
 
 		if ( this._debugBoundsQuad !== undefined ) {
--- a/example/dynamicScene.html
+++ b/example/dynamicScene.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 	<head>
-		<title>Dynamic Scene Editing Example</title>
+		<title>Dynamic Scene Editing</title>
 		<meta name="viewport" content="width=device-width, initial-scale=1">
 		<style>
 			body { margin: 0; overflow: hidden; background: #222; }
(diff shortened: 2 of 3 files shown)


gkjohnson:
Thanks for the looking into this! I think I'm hesitant to add anything that may turn into an app with custom UI, etc that has to be maintained and that users may wind up requesting features for. In my mind the example for this is just something simple with multiple objects that involves:

- clicking to change the active object
- widget for moving the object
- simple color adjustment

To that end the example serves the purpose of showing & testing that dynamic adjustment works. And then a more complete editor can be started as a dedicated project.

src/webgpu/WebGPUPathTracer.js
```diff
@@ -274,11 +274,20 @@ export class WebGPUPathTracer {
 
 		} );
 
-		// Build TLAS and compute functions
-		const bvhData = new PathtracerBVHComputeData( scene );
+		// Reuse the data object for topology changes within the same scene. This preserves
+		// the texture atlas cache and avoids repacking unchanged textures.
+		const sceneChanged = this.scene !== scene;
+		const bvhData = sceneChanged ? new PathtracerBVHComputeData( scene ) : this._bvhData;
```
gkjohnson: This strategy will break are cases where users _have_ modified material textures or added objects into the scene, or modified geometries. Just because the root object is the same doesn't mean the content has changed or isn't completely different. 

It would be good to make this more smart, though. Having some method of hashing the objects in the scene to detect changes and only update the scene when necessary would be good. This would involve creating a hash per object involving geometry uuid, and attribute versions per geometry, at least. We'd also need to force transform updates, etc.

It might be nice to enable this at some point but I think it also becomes fairly complex. Exposing dedicated functions like we have for updating transforms so users can call the necessary function is probably our best bet for now.

src/webgpu/WebGPUPathTracer.js
```diff
@@ -274,11 +274,20 @@ export class WebGPUPathTracer {
 
 		} );
 
-		// Build TLAS and compute functions
-		const bvhData = new PathtracerBVHComputeData( scene );
+		// Reuse the data object for topology changes within the same scene. This preserves
+		// the texture atlas cache and avoids repacking unchanged textures.
```
gkjohnson: The "AtlasTexture" class will already keep a hash of all the texture content and avoid repacking the textures if the set is already the same.

Ryan-G7:
Thanks for the feedback.

 I have simplified  the dynamic scene example and I agree that reusing `_bvhData` based only on the scene root is unsafe.

I removed that reuse path: `setScene()` now rebuilds the scene data when called, while in-place changes use the explicit `updateTransforms()` and `updateMaterials()` methods.

gkjohnson:
Thanks! Looks good - I think last thing is to add some kind of really brief description at the top with instructions so it's more clear how to change transform tools, etc.

Let me know if there's anything else you're interested in taking on - but some other options that come to mind:

- adding support for ArrayCamera I think would be a cool feature (outlined [here](https://github.com/gkjohnson/three-gpu-pathtracer/pull/799#issuecomment-5149931772))
- investigating a good way to implement nice post effects like bloom or lens flares (see https://github.com/gkjohnson/three-gpu-pathtracer/issues/777#issuecomment-5150605758) would also be cool to see.

Ryan-G7:
I've added a description for this example.

> * adding support for ArrayCamera I think would be a cool feature (outlined [here](https://github.com/gkjohnson/three-gpu-pathtracer/pull/799#issuecomment-5149931772))

I think this is a good idea, and I'll look into it over the next few days.

> * investigating a good way to implement nice post effects like bloom or lens flares (see [WebGPUPathTracer: Remaining features #777 (comment)](https://github.com/gkjohnson/three-gpu-pathtracer/issues/777#issuecomment-5150605758)) would also be cool to see.

I took a look at those real Bloom scenarios, and they look amazing. incorporating these into the project (e.g., for vehicle lights and general lighting) makes a lot of sense. I'll take some time to investigate this.

gkjohnson:
> I took a look at those real Bloom scenarios, and they look amazing. incorporating these into the project (e.g., for vehicle lights and general lighting) makes a lot of sense. I'll take some time to investigate this.

It's probably worth looking around to see what other renderers like Blender use for these kinds of effects, as well. We can decide once things are further along but if the effects can run efficiently in real time then it might be worth considering adding them to the three.js repo. If not we can definitely host them here.

PR #798, opened 2026-07-30
Fix WebGPU GGX graphing example

Ryan-G7:
## Summary

- Fix the graphing example runtime issue, update example to use the current vector-based GGX APIs, and simplify the fixed view direction.
- Make GGX WGSL helper functions declare their own constant dependencies.
- Load the Turquin LUT without using top-level await.
- Upgrade Three.js to version 0.185.1 and synchronize the package-lock.json file.
- Prevent unrelated WebCodecs type packages from polluting declaration checks.


tsconfig.json
```diff
@@ -5,6 +5,7 @@
 		"strictNullChecks":true,
 		"noImplicitAny": true,
 		"lib": [ "dom", "es2017" ],
+		"types": [],
```
gkjohnson, line 8: > Prevent unrelated WebCodecs type packages from polluting declaration checks.

Can you explain what was going on here? Were there some conflicting types or something?

gkjohnson:
Thanks!

PR #799, opened 2026-07-31
Fix sample UI updates

Ryan-G7:
### Current changes
- `overlay.js`: Added `await renderer.init()` to prevent uninitialized renderer errors.
- Updated the example UI via `setPercentage( 1 )` to correctly display the currently accumulated sample count.

### Remaining Examples
I have tested the remaining examples, such as `fog`, `materialBall`, `areaLight`, `materialDatabase`, and `spotLights`, are currently not working properly. 

They can be restored to a working state once the WebGPU path tracer supports the following features:
- Collecting scene lights and uploading their data to GPU buffers.
- `updateLights()` and dynamic light updates.
- Direct-light sampling and shadow rays.
- NEE / MIS for area, spot, point, and directional lights.
- IES profile sampling.
- Volumetric media traversal and scattering for fog materials.

### lkg Example
The `lkg` example has not been migrated to WebGPU yet. It currently depends on:
- The WebGL-based `QuiltPathTracingRenderer`.
- GLSL fullscreen and quilt-preview materials.
- The Looking Glass WebXR WebGL layer integration.

I believe migrating this will be quite complex.

### DIFF
--- a/example/areaLight.js
+++ b/example/areaLight.js
@@ -133,6 +133,7 @@ async function init() {
 	// initialize scene
 	pathTracer.setScene( scene, camera );
 
+	loader.setPercentage( 1 );
 	loader.setCredits( CREDITS );
 
 	// gui
--- a/example/basic.js
+++ b/example/basic.js
@@ -97,6 +97,7 @@ async function init() {
 	// initialize the path tracer
 	pathTracer.setScene( scene, camera );
 
+	loader.setPercentage( 1 );
 	loader.setCredits( CREDITS );
 	loader.setDescription( DESCRIPTION );
 
--- a/example/depthOfField.js
+++ b/example/depthOfField.js
@@ -137,6 +137,7 @@ async function init() {
 	// update the scene
 	pathTracer.setScene( scene, camera );
 
+	loader.setPercentage( 1 );
 	loader.setCredits( CREDITS );
 	loader.setDescription( DESCRIPTION );
 	onParamsChange();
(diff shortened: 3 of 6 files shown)


gkjohnson:
Thanks!

> I have tested the remaining examples, such as fog, materialBall, areaLight, materialDatabase, and spotLights, are currently not working properly.

These all rely on punctual lights (MIS/NEE) or fog. #796 is in progres for MIS and lighting will come in a separate PR when I have a chance. These examples will get fixed as some of the new features are added.

> The lkg example has not been migrated to WebGPU yet. It currently depends on:

Regarding the LKG example - I think this could probably be rewritten entirely. If I recall correctly the Looking Glass API relies on some kind of polyfilled version of the WebXR API to report a multi-view quilted layout. And I believe three.js exposes this as an "ArrayCamera" via WebXRManager with different subviews in the the final render target. I'd have to check if WebGPURenderer supports WebXR but ultimately I think just making sure the path tracer supports ArrayCameras properly would be enough to get this working.

PR #776 added support for a custom "getCameraRayFn" function so you can imagine writing one (or modifying the built in one) to support generating a ray for the appropriate camera depending on which pixel is being rendered to. The brute force method would be to upload all subview boxes and projection matrices as array uniforms to the shader and iterate over every subview to find the bounds the current pixel lies in to generate the ray, but I'm sure there are some other smarter ways to do this, too.

A cool, though questionably useful, side effect of handling ArrayCameras generically is that it should let us render stereo images that can be live rendered and looked at in a VR headset 😅.

If adding this is interesting I can provide some more information if needed.

PR #808, opened 2026-08-07
WebGPUPathTracer: Add Transmission Support

gkjohnson:
Related to #770, #777

- Adds support for transmission and thin film transmission
- Removes "transmissiveBounces"
- Adds three options for transparent background handling with transmissive objects - display env map, overlay env map with average color used to drive alpha (so there's some color overlaid), and just transparent.
- Add support for thin film, mirroring Blender's method
- Rename "thinFilm" to "thinWall", matching blenders vocabulary

**TODO**
- Clean up
- Review

image 
image
image

And the new transmissive background handling options:

| Env Map | Overlay | Transparent | WebGLPathTracer Approach |
|---|---|---|---|
| image | image | image | image |

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -207,7 +207,6 @@ async function init() {
 	pathTracer = new WebGPUPathTracer( renderer );
 	pathTracer.tiles.set( params.tiles, params.tiles );
 	pathTracer.multipleImportanceSampling = params.multipleImportanceSampling;
-	pathTracer.transmissiveBounces = 10;
 	pathTracer.useMegakernel( params.useMegakernel );
 
 	// camera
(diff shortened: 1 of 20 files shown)


gkjohnson:
There are still some noteable differences compared to other pathtracers. The WebGLPathTracer does not have this issue.

| This PR | Blender | Stellar |
|---|---|---|
| image | image | image |

- [x] Normals have been fixed by transforming the tangent appropriately
- [x] Fix irridesence / ior / eta handling causing black points

gkjohnson:
Fixed the rendering of the glass dome by
- Fixing tangents not being provided in a consistent world frame
- Passing ior in with the correct order (accounting for which side of the surface the ray is incoming from)
- Handle TIR explicitly in the iridescence function

image

I think there's some logic that can be cleaned up here but hopefully this can be handled more organized with a "layering" model down the line.

PR #810, opened 2026-08-09
Support ArrayCamera and restore the LKG example

Ryan-G7:
## Summary

- Add `ArrayCamera` ray generation support to the WebGPU path tracer.
- Migrate the Looking Glass quilt example to WebGPU.
- Add a node-based quilt preview material for storage textures.

## Details

`ArrayCamera` sub-camera inverse view-projection matrices and viewports are
uploaded as uniform arrays. The camera ray function selects the appropriate
sub-camera for each output pixel.

Changing the camera ray function rebuilds the compute kernels through
`setBVHData`, while regular camera movement only updates uniforms and resets
accumulation.

The LKG example uses the megakernel backend to avoid the large fixed buffers
required by the wavefront backend. It also requests the available storage
buffer limits needed by larger LDraw scenes.

OrbitControls now use raster rendering during interaction. When interaction
ends, the quilt cameras are updated once and path tracing resumes. This avoids
showing a cleared quilt texture while the camera is moving.

### DIFF
--- a/example/lkg.js
+++ b/example/lkg.js
@@ -101,7 +101,7 @@ const params = {
 
 	enable: true,
 	model: modelName in MODELS ? modelName : 'Voltron - Voltron',
-	renderScale: 0.25,
+	renderScale: 1,
 	tiles: 1,
 
 	samplesPerFrame: 1,
@@ -167,7 +167,7 @@ async function init() {
 	const adapter = await navigator.gpu?.requestAdapter();
 	const requiredLimits = getRequiredDeviceLimits( adapter );
 	// init renderer
-	renderer = new WebGPURenderer( { antialias: true, preserveDrawingBuffer: true, requiredLimits } );
+	renderer = new WebGPURenderer( { antialias: true, requiredLimits } );
 	await renderer.init();
 	renderer.toneMapping = ACESFilmicToneMapping;
 	document.body.appendChild( renderer.domElement );
@@ -348,11 +348,14 @@ async function animate() {
 
 	}
 
+	let pathTracingFrame = false;
 	if ( ! params.enable || interacting ) {
 
 		renderer.render( scene, camera );
 
 	} else {
+
+		pathTracingFrame = true;
 
 		// set path tracer variables
 		pathTracer.pause = params.pause;
@@ -386,7 +389,7 @@ async function animate() {
 	distEl.innerText = `Distance: ${ camera.position.length().toFixed( 2 ) }`;
 
 	const queue = renderer.backend?.device?.queue;
-	if ( queue?.onSubmittedWorkDone ) {
+	if ( pathTracingFrame && queue?.onSubmittedWorkDone ) {
 
 		// Keep CPU-side progress in step with completed GPU work.
 		await queue.onSubmittedWorkDone();
@@ -426,7 +429,6 @@ function onLkgParamsChange() {
 	const { renderScale } = params;
 
 	lkgParams = getLkgParams( params.numViews );
-	const previousCameraCount = quiltCamera.cameras.length;
 	updateQuiltCamera();
 
 	const width = Math.max( 1, Math.floor( renderScale * lkgParams.quiltWidth ) );
@@ -435,20 +437,11 @@ function onLkgParamsChange() {
 	previewQuad.material.quiltMap = pathTracer.target;
 	previewQuad.material.quiltDimensions.set( lkgParams.quiltTilesX, lkgParams.quiltTilesY );
 
-	if ( previousCameraCount !== quiltCamera.cameras.length ) {
-
-		pathTracer.setCamera( quiltCamera );
-
-	} else {
-
-		pathTracer.updateCamera();
-
-	}
+	pathTracer.updateCamera();
 
 }
 
-// Rebuild the off-axis cameras and their quilt viewports. Extra cells in the
-// final row repeat the last view so every output pixel belongs to a camera.
+// Rebuild the off-axis cameras and their quilt viewports.
 function updateQuiltCamera() {
 
 	const {
@@ -461,7 +454,6 @@ function updateQuiltCamera() {
 	const { renderScale, viewCone, viewerDistance } = params;
 	const width = Math.max( 1, Math.floor( renderScale * quiltWidth ) );
 	const height = Math.max( 1, Math.floor( renderScale * quiltHeight ) );
-	const cameraCount = quiltTilesX * quiltTilesY;
 
 	const displayHalfHeight = DISPLAY_HEIGHT * 0.5;
 	const displayHalfWidth = DISPLAY_WIDTH * 0.5;
@@ -472,7 +464,7 @@ function updateQuiltCamera() {
 	camera.updateProjectionMatrix();
 	camera.updateMatrixWorld();
 
-	while ( quiltCamera.cameras.length < cameraCount ) {
+	while ( quiltCamera.cameras.length < numViews ) {
 
 		const subCamera = new PerspectiveCamera();
 		subCamera.coordinateSystem = WebGPUCoordinateSystem;
@@ -481,11 +473,10 @@ function updateQuiltCamera() {
 
 	}
 
-	quiltCamera.cameras.length = cameraCount;
-	for ( let i = 0; i < cameraCount; i ++ ) {
+	quiltCamera.cameras.length = numViews;
+	for ( let i = 0; i < numViews; i ++ ) {
 
-		const viewIndex = Math.min( i, numViews - 1 );
-		const viewAlpha = numViews === 1 ? 0.5 : viewIndex / ( numViews - 1 );
+		const viewAlpha = numViews === 1 ? 0.5 : i / ( numViews - 1 );
 		const offset = MathUtils.lerp( - halfViewWidth, halfViewWidth, viewAlpha );
 		const subCamera = quiltCamera.cameras[ i ];
 
--- a/example/lkg.js
+++ b/example/lkg.js
@@ -123,6 +123,8 @@ const params = {
 let loadingEl, samplesEl, distEl;
 let renderer, camera, quiltCamera;
 let pathTracer, previewQuad, controls, scene;
+let interacting = false;
+let cameraChanged = false;
 const _translation = new Matrix4();
 const _size = new Vector2();
 
@@ -198,10 +200,15 @@ async function init() {
 
 	// init controls
 	controls = new OrbitControls( camera, renderer.domElement );
-	controls.addEventListener( 'change', () => {
+	controls.addEventListener( 'start', () => {
 
-		updateQuiltCamera();
-		pathTracer.updateCamera();
+		interacting = true;
+
+	} );
+	controls.addEventListener( 'end', () => {
+
+		interacting = false;
+		cameraChanged = true;
 
 	} );
 
@@ -333,7 +340,15 @@ async function init() {
 
 async function animate() {
 
-	if ( ! params.enable ) {
+	if ( cameraChanged && ! interacting ) {
+
+		updateQuiltCamera();
+		pathTracer.updateCamera();
+		cameraChanged = false;
+
+	}
+
+	if ( ! params.enable || interacting ) {
 
 		renderer.render( scene, camera );
 
--- a/example/lkg.js
+++ b/example/lkg.js
@@ -70,21 +70,14 @@ const MATERIALS_URL = 'https://raw.githubusercontent.com/gkjohnson/ldraw-parts-l
 const PARTS_PATH = 'https://raw.githubusercontent.com/gkjohnson/ldraw-parts-library/master/complete/ldraw/';
 const MODELS = {
 	'X-Wing': 'https://raw.githubusercontent.com/mrdoob/three.js/r150/examples/models/ldraw/officialLibrary/models/7140-1-X-wingFighter.mpd_Packed.mpd',
-	'UCS AT-ST': 'https://raw.githubusercontent.com/mrdoob/three.js/r150/examples/models/ldraw/officialLibrary/models/10174-1-ImperialAT-ST-UCS.mpd_Packed.mpd',
-	'Death Star': 'https://raw.githubusercontent.com/gkjohnson/ldraw-parts-library/master/models/10143-1 - Death Star II.mpd',
 	'Lunar Vehicle': 'https://raw.githubusercontent.com/mrdoob/three.js/r150/examples/models/ldraw/officialLibrary/models/1621-1-LunarMPVVehicle.mpd_Packed.mpd',
 };
 
 [
 	'6814-1 - Ice Tunnelator.mpd',
-	'6861-2 - Super Model Building Instruction.mpd',
-	'75060 - Slave I.mpd',
 	'6983-1 - Ice Station Odyssey.mpd',
 	'6835-1 - Saucer Scout.mpd',
-	'21311 - Voltron - Voltron.mpd',
-	'21303 - WALLE.mpd',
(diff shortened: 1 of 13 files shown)


src/webgpu/WebGPUPathTracer.js
```diff
@@ -378,6 +376,7 @@ export class WebGPUPathTracer {
 		}
 
 		this._bvhData.fns.getCameraRay = this._cameraRayFnHandle.fn;
+		this._pathTracer.setBVHData( this._bvhData );
```
gkjohnson: Why is this being set here? The camera changing is unrelated to the BVH data.

src/webgpu/WebGPUPathTracer.js
```diff
@@ -308,7 +307,6 @@ export class WebGPUPathTracer {
 
 		this.scene = scene;
 		this._bvhData = bvhData;
-		this._pathTracer.setBVHData( bvhData );
```
gkjohnson, line 311: Likewise this shouldn't be removed.

src/webgpu/WebGPUPathTracer.js
```diff
@@ -182,7 +182,6 @@ export class WebGPUPathTracer {
 
 		this._pathTracer.dispose();
 		this._pathTracer = value ? new MegaKernelPathTracer( this._renderer ) : new WaveFrontPathTracer( this._renderer );
-		this._pathTracer.setBVHData( this._bvhData );
```
gkjohnson, line 185: Nor should this

src/webgpu/shims/ArrayCameraShim.js
```diff
@@ -0,0 +1,78 @@
...
+		}
+	`;
+
+	const update = () => {
+
+		if ( this.cameras.length !== cameraCount ) {
+
+			throw new Error( 'ArrayCamera: Call WebGPUPathTracer.setCamera() after changing the number of sub-cameras.' );
+
+		}
```
gkjohnson: Is this actually needed? The goal of this design is to be able to return would be for WebGPUPathTracer to call this "update" function, which would then update any uniforms needed - like expanding the array uniform size in this case. So a user should really just have to call `WebGPUPathTracer.updateCamera` when the number of cameras change

src/webgpu/shims/ArrayCameraShim.js
```diff
@@ -0,0 +1,78 @@
...
+					cameraIndex = i;
+					cameraUv = ( pixel - viewport.xy ) / viewport.zw;
+					break;
+
+				}
+
+			}
+
+			let ndc = cameraUv * 2.0 - vec2f( 1.0 );
+			return ${ ndcToCameraRay }( ndc, ${ invViewProjectionMatrices }[ cameraIndex ] );
```
gkjohnson: One thing I'm noticing is that when when there are _blank_ tile spaces (or pixels that do _not_ land within a camera bounds) then it simply using the the 0u camera. This means that a bunch of work will potentially be done for pixels that will never be seen in the given WebXR device. I see that you're implicitly upping the number of array cameras to the nearest square number to work around this in the demo but this isn't a practical solution for real use cases since the WebXRManager (or end user) should ultimately be providing the array camera.

I'm thinking it would be best if this function returned a boolean value to indicate whether ray should be skipped or not. And then ray can then be output using a pointer parameter.

example/lkg.js
```diff
@@ -77,7 +101,7 @@ const params = {
 
 	enable: true,
 	model: modelName in MODELS ? modelName : 'Voltron - Voltron',
-	renderScale: 1,
+	renderScale: 0.25,
```
gkjohnson: We should be demonstrating that the project works in a functional in realistic and high quality scenarios, so we shouldn't be lowering the render scale. If there are issues you're trying to address we should discuss those instead of trying to work around them.

example/lkg.js
```diff
@@ -92,24 +116,45 @@ const params = {
...
+
+			limits[ limit ] = value;
+
+		}
+
+	}
+
+	return limits;
+
+}
```
gkjohnson: I this needed in this example? I suppose it might be needed for the AT-ST model or something?

example/lkg.js
```diff
@@ -119,59 +164,58 @@ async function init() {
 	loadingEl = document.getElementById( 'loading' );
 	samplesEl = document.getElementById( 'samples' );
 
+	const adapter = await navigator.gpu?.requestAdapter();
+	const requiredLimits = getRequiredDeviceLimits( adapter );
 	// init renderer
-	renderer = new WebGLRenderer( { antialias: true } );
+	renderer = new WebGPURenderer( { antialias: true, preserveDrawingBuffer: true, requiredLimits } );
```
gkjohnson: What is `preserveDrawingBuffer` for? This doesn't seem to be a valid WebGPU option.

example/lkg.js
```diff
@@ -119,59 +164,58 @@ async function init() {
...
-
-	camera.fov = ptRenderer.viewFoV * MathUtils.RAD2DEG;
-	camera.updateProjectionMatrix();
+	quiltCamera = new ArrayCamera();
+
+	// initialize the path tracer. The quilt has its own fixed resolution, independent
+	// of the preview canvas, so automatic and low-resolution resizing must be disabled.
+	pathTracer = new WebGPUPathTracer( renderer );
+	// Avoid the large fixed ray queues used by the wavefront backend.
+	pathTracer.useMegakernel( true );
```
gkjohnson: Using megakernel makes the page nearly unuseable performance-wise - especially with just a single rendering tile. We're using wavefront by default in all other examples for this reason so we should be doing the same thing here.

example/lkg.js
```diff
@@ -444,16 +423,98 @@ function getLkgParams( numViews ) {
...
+		numViews,
+		quiltTilesX,
+		quiltTilesY,
+		quiltWidth,
+		quiltHeight,
+	} = lkgParams;
+	const { renderScale, viewCone, viewerDistance } = params;
+	const width = Math.max( 1, Math.floor( renderScale * quiltWidth ) );
+	const height = Math.max( 1, Math.floor( renderScale * quiltHeight ) );
+	const cameraCount = quiltTilesX * quiltTilesY;
```
gkjohnson: As mentioned above we shouldn't be doing something different than what we're implying in the UI - we should make sure the renderer can handle an arbitrary set of cameras in an array camera

gkjohnson (commented):
Thanks this is great! It seems like the right direction. To some of your points:

> Changing the camera ray function rebuilds the compute kernels through
setBVHData, while regular camera movement only updates uniforms and resets
accumulation.

Why is this necessary?

> The LKG example uses the megakernel backend to avoid the large fixed buffers
required by the wavefront backend. It also requests the available storage
buffer limits needed by larger LDraw scenes.

Why are the large buffers specifically an issue in this example? Wavefront is being used in all other examples.

> OrbitControls now use raster rendering during interaction. When interaction
ends, the quilt cameras are updated once and path tracing resumes. This avoids
showing a cleared quilt texture while the camera is moving.

I think the interaction feels a bit odd at the moment and I'm wondering whether we should get rid of the controls entirely to just focus on ArrayCamera support. But we can deal with this once some of the more technical issues are handled.

src/webgpu/WebGPUPathTracer.js
Ryan-G7: The call in setCamera() is currently necessary because getCameraRay is compiled into the ray-generation kernel through proxyFn. Removing setBVHData() from setCamera() would therefore leave the previous camera-ray shader in use when the function changes.

example/lkg.js
Ryan-G7: The choice of megakernel is intentional, as the wavefront backend simply cannot run with the current settings.

Due to the 128 MiB ray queue capacity limit, it cannot accommodate all rays under a single-tile setup, resulting in a black output. If we want to use wavefront later, this queue capacity and tiling limitation must be resolved first.

example/lkg.js
Ryan-G7: Yes, it's needed here. Without requesting higher limits, loading complex models can easily fail and crash.

example/lkg.js
Ryan-G7: Makes sense! I had temporarily lowered it just to keep frame rates higher during testing, but I agree that we should showcase the highest quality output by default.

I've bumped renderScale back to 1.

src/webgpu/shims/ArrayCameraShim.js
Ryan-G7: I've pushed an updated version, but to be honest, I'm a bit confused about this part of the logic, so I'm not entirely sure if my changes are completely correct. :thinking: :thinking:

src/webgpu/WebGPUPathTracer.js
gkjohnson: Why does it need to be recompiled, though? Is it because the uniform arrays need to be fixed length? Either way we shouldn't move "setBVHData" calls around just to take advantage of it's side effect. If we need to we should add a "rebuild" function to the PathTracerBackend that can be called when we know we need to rebuild the shaders.

Then the "update" function can return a boolean value that indicates whether the "update" implies a necessary shader recompile.

example/lkg.js
gkjohnson: Do we know which models? I'm more inclined to change or remove some of the complicated models from the demos than include this.

example/lkg.js
gkjohnson: > Due to the 128 MiB ray queue capacity limit, it cannot accommodate all rays under a single-tile setup,

I don't think there's any reason we need to use a single tile set up. We should adjust that to 3x3 tiles or larger if needed and disallow lower values. The demo is unusable and gives a very poor impression with the current setup due to the performance.

example/lkg.js
```diff
@@ -386,7 +389,7 @@ async function animate() {
 	distEl.innerText = `Distance: ${ camera.position.length().toFixed( 2 ) }`;
 
 	const queue = renderer.backend?.device?.queue;
-	if ( queue?.onSubmittedWorkDone ) {
+	if ( pathTracingFrame && queue?.onSubmittedWorkDone ) {
```
gkjohnson: What was this change for?

src/webgpu/WebGPUPathTracer.js
```diff
@@ -366,10 +364,13 @@ export class WebGPUPathTracer {
...
 				},
 				fn: wgslTagFn/* wgsl */`
-					fn getCameraRay( uv: vec2f, resolution: vec2f ) -> ${ rayStruct } {
+					fn getCameraRay( uv: vec2f, resolution: vec2f, ray: ptr<function, ${ rayStruct }> ) -> bool {
 
 						let ndc = uv * 2.0 - vec2f( 1.0 );
-						return ${ ndcToCameraRay }( ndc, ${ invViewProjectionMatrix } );
+						let cameraRay = ${ ndcToCameraRay }( ndc, ${ invViewProjectionMatrix } );
+						ray.origin = cameraRay.origin;
+						ray.direction = cameraRay.direction;
```
gkjohnson: We should able to make this one line like so:

```wgsl
*ray = ${ ndcToCameraRay }( ndc, ${ invViewProjectionMatrix } );
```

src/webgpu/shims/ArrayCameraShim.js
```diff
@@ -0,0 +1,90 @@
...
+			rebuild();
+
+		}
+
+		for ( let i = 0; i < cameraCount; i ++ ) {
+
+			const camera = this.cameras[ i ];
+			const viewport = camera.viewport;
+
+			camera.updateMatrixWorld();
```
gkjohnson, line 60: We need to set the camera coordinate frame to WebGPUCoordinateSystem and update the projection matices just like we do in the PhysicalCameraShim

src/webgpu/shims/ArrayCameraShim.js
gkjohnson: The new function signature looks great - that's what I was thinking. Thanks!

src/webgpu/WebGPUPathTracer.js
Ryan-G7: I have resolved this and added the rebuild method to PathTracerBackend.

example/lkg.js
Ryan-G7: Yes, I removed that part of the code, tested all models, and removed the ones that failed to load correctly.

example/lkg.js
Ryan-G7: I have updated it.

example/lkg.js
Ryan-G7: This was originally meant to avoid queue conflicts between raster interaction and path tracing, but I have removed both the raster interaction and the orbitController. This example will now focus strictly on showing ArrayCamera.

src/webgpu/shims/ArrayCameraShim.js
```diff
@@ -0,0 +1,96 @@
...
+						ray.direction = cameraRay.direction;
+						return true;
+
+					}
+
+				}
+
+				return false;
+
+			}
```
gkjohnson: We shouldn't need to rebuild or modify the wgslTagFn instance here. The template tag functions such that it retains references to the uniforms. So we just need to modify the uniform nodes in place and new sizes will be accounted for when rebuilding the shader. I'll make the change in a commit to this branch so it's more clear

example/lkg.js
```diff
@@ -1,37 +1,60 @@
...
-import { VRButton } from 'three/examples/jsm/webxr/VRButton.js';
+		name = name.replace( /\\/g, '/' );
+		name = name.startsWith( 's/' ) ? `parts/${ name }` : name.startsWith( '48/' ) ? `p/${ name }` : name;
+		return `0 FILE ${ name }`;
+
+	} );
+
+}
+
+class PackedLDrawLoader extends LDrawLoader {
```
gkjohnson: What is this for? There shouldn't be any reason to write or wrap a custom loader for the LDraw models.

src/webgpu/shims/ArrayCameraShim.js
gkjohnson: I've just updated the branch to modify the uniforms in place - take a look and let me know what you think.

example/lkg.js
Ryan-G7: PackedLDrawLoader was originally added as a workaround for older packed MPDs whose embedded subfile names used backslash paths such as 0 FILE s\..., which did not match the paths expected by LDrawLoader. The affected models are no longer in the demo's model list, so the wrapper is unnecessary now. I've removed it and restored the standard LDrawLoader.

src/webgpu/shims/ArrayCameraShim.js
Ryan-G7: Thanks, that makes sense. and I made a few changes for you to check out, I removed the initial result.update() call and the redundant getCameraRay reassignment in updateCamera(). On the other hand, setCamera() explicitly rebuilds once because it replaces the camera-ray function itself. I also separated rebuild from reset to prevent the backend from being cleared twice.

gkjohnson: > I also separated rebuild from reset to prevent the backend from being cleared twice.

I think I'd prefer to keep calling "reset" in rebuild. If we're seeing that calling "reset" multiple times is causing any kind of hitches due to sequential calls we can revisit (eg the clearing can happen at the top of the render generator function) - otherwise this feels like a micro optimization and I'd rather keep things traceable with side effects fired where required.

Ryan-G7: I've restored the reset calls within rebuild. We can revisit deferring the clears if profiling shows that sequential resets cause noticeable hitches.

gkjohnson:
Thanks! I'll have to test and get it working with my looking glass display at some point.

PR #812, opened 2026-08-10
Add scale-independent ray origin offsets

Ryan-G7:
## Summary

Adds scale-independent ray origin offsets and removes the fixed local-space triangle intersection threshold.

## Changes

- Add the Ray Tracing Gems bit-level ray origin offset used by [Cycles.](https://github.com/blender/blender/blob/main/intern/cycles/kernel/bvh/util.h)
- Apply the offset to secondary and shadow rays in both MegaKernel and
  Wavefront paths.
- Orient the offset using the outgoing direction and geometric normal.
- Change the triangle intersection threshold from `1e-5` to `0.0`.

The previous `1e-5` threshold was evaluated in object-local space, making its effective size dependent on object scale. It could either fail to prevent self-intersections or skip valid nearby geometry.

### DIFF
--- a/src/webgpu/nodes/sampling.wgsl.js
+++ b/src/webgpu/nodes/sampling.wgsl.js
@@ -166,7 +166,7 @@ export const equirectDirectionPdfFn = wgslFn( /* wgsl */ `
 		let uv = equirectDirectionToUv( direction );
 		let theta = uv.y * PI;
 		let sinTheta = sin( theta );
-		if ( sinTheta == 0.0 ) {
+		if ( sinTheta <= EPSILON ) {
 
 			return 0.0;
 
--- a/src/webgpu/nodes/sampling.wgsl.js
+++ b/src/webgpu/nodes/sampling.wgsl.js
@@ -166,7 +166,7 @@ export const equirectDirectionPdfFn = wgslFn( /* wgsl */ `
 		let uv = equirectDirectionToUv( direction );
 		let theta = uv.y * PI;
 		let sinTheta = sin( theta );
-		if ( sinTheta <= EPSILON ) {
+		if ( sinTheta == 0.0 ) {
 
 			return 0.0;
 
--- a/src/webgpu/nodes/utils.wgsl.js
+++ b/src/webgpu/nodes/utils.wgsl.js
@@ -172,8 +172,11 @@ export const applyWrapFunc = wgslFn( /* wgsl */ `
 
 ` );
 
-// Bit-level origin offset from Ray Tracing Gems, also used by Cycles. It keeps
-// secondary and shadow rays off the source triangle without a scene-scale bias.
+// Bit-level ray origin offset based on Section 6.2.2 of "A Fast and Robust
+// Method for Avoiding Self-Intersection" in Ray Tracing Gems:
+// https://github.com/Apress/ray-tracing-gems/blob/master/Ch_06_A_Fast_and_Robust_Method_for_Avoiding_Self-Intersection/offset_ray.cu
+// The original implementation expects a normal oriented for the outgoing ray;
+// this version orients the geometric normal internally.
 export const offsetRayOriginFunc = wgslFn( /* wgsl */ `
 
 	fn offsetRayOrigin( point: vec3f, direction: vec3f, geometricNormal: vec3f ) -> vec3f {
--- a/src/webgpu/nodes/utils.wgsl.js
+++ b/src/webgpu/nodes/utils.wgsl.js
@@ -172,6 +172,26 @@ export const applyWrapFunc = wgslFn( /* wgsl */ `
 
 ` );
 
+// Bit-level origin offset from Ray Tracing Gems, also used by Cycles. It keeps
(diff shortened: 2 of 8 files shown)


gkjohnson:
Can you provide some screenshots of the issues this fixes and how you're testing this?

Ryan-G7:
> Can you provide some screenshots of the issues this fixes and how you're testing this?

Thanks. I added a focused repro and compared the parent branch with this PR.

The repro renders three world-space-identical spheres. Their local geometry and object scales are reciprocal, using scales of 1e-4, 1, and 1e4. I rendered both branches with the same camera, resolution, Sobol sampler, and 256 spp. 

On the parent branch, the 1e-4 object shows a incorrect contact shadow caused by self-intersections. On this branch, that artifact is removed and the contact shadows are consistent across all three scales.

| webgpu/mis | this PR |
| :---: | :---: |
| 3aef60645795280cf607af0aa306ac33 | d091d73fd0f6361a716a71f57e0461d2 |

src/webgpu/nodes/sampling.wgsl.js
```diff
@@ -166,7 +166,7 @@ export const equirectDirectionPdfFn = wgslFn( /* wgsl */ `
 		let uv = equirectDirectionToUv( direction );
 		let theta = uv.y * PI;
 		let sinTheta = sin( theta );
-		if ( sinTheta == 0.0 ) {
+		if ( sinTheta <= EPSILON ) {
```
gkjohnson: What is this change for?

Ryan-G7: Thiss designed to guard against floating errors near the poles that cause invalid PDFs, but this change is no longer needed since #813 switched to using equirectLuminancePdf.

gkjohnson: lets remove it with #813 merged, then - then I'll take a deeper look at the rest of the PR

Ryan-G7: I've completed this. And, when do you think we can make webgpu-pathtracer the main branch and remove the WebGL version?

src/webgpu/nodes/utils.wgsl.js
```diff
@@ -172,6 +172,26 @@ export const applyWrapFunc = wgslFn( /* wgsl */ `
 
 ` );
 
+// Bit-level origin offset from Ray Tracing Gems, also used by Cycles. It keeps
+// secondary and shadow rays off the source triangle without a scene-scale bias.
+export const offsetRayOriginFunc = wgslFn( /* wgsl */ `
```
gkjohnson: Lets add a direct reference to the chapter and section that this function is provided by in Ray Tracing Gems - it looks like chapter 6.2?

src/webgpu/nodes/utils.wgsl.js
```diff
@@ -172,6 +172,26 @@ export const applyWrapFunc = wgslFn( /* wgsl */ `
...
+		let intScale = 256.0;
+		let integerOffset = vec3i( intScale * normal );
+		let pointBits = bitcast<vec3i>( point );
+		let signedOffset = select( integerOffset, -integerOffset, point < vec3f( 0.0 ) );
+		let offsetPoint = bitcast<vec3f>( pointBits + signedOffset );
+		let origin = 1.0 / 32.0;
+		let floatScale = 1.0 / 65536.0;
+		return select( offsetPoint, point + floatScale * normal, abs( point ) < vec3f( origin ) );
+
+	}
```
gkjohnson, line 253: It looks like this function differs bit from the RTGems function. Specifically it takes three parameters and internally chooses the offset. Lets also add a note that the original implementation assumes an already-scatter-aligned geometric normal for the sake of clarity:

Image

gkjohnson (commented):
Thanks - I didn't know about this approach.

> when do you think we can make webgpu-pathtracer the main branch and remove the WebGL version?

I don't have an exact timeline but overall I want to get to parity with the WebGL version, address energy preservation for transmissive surfaces, and do some organization cleanup, docs, etc. Maybe a few weeks?

- Fix transmission energy conservation (in progress)
- Review + merge NEE/MIS, lighting branches (in progress)
- Tune, review, merge the wavefront improvements from #770 (in progress)
- Fix some bugs, polish + test further
- Fog (might punt to after an initial release)
- Matte materials (might punt until custom surfaces / layers are supported)
- Docs via JSDoc

And some features that I'd really like to see. I don't consider these blocking for an initial release but I think they'd be some nice new features:

- Subsurface scattering
- Spectral dispersion
- EON support
- Investigate MNEE support for caustics
- High quality bloom, lens flares

I've tried to update the list in #777 to reflect what's left in the short vs long term. If you're interested in picking any of these up let me know which ones.

Ryan-G7:
Thanks, I’ve updated the comment .

> I've tried to update the list in #777 to reflect what's left in the short vs long term. If you're interested in picking any of these up let me know which ones.

I'll look into the implementation of EON and spectral dispersion next. Since I've already researched papers on EON before, I might prioritize this step first, and then move on to spectral dispersion.

PR #813, opened 2026-08-10
Fix webgpu environment MIS radiance

Ryan-G7:
The WebGPU BSDF miss path now evaluates the environment using the exact ray direction instead of applying an additional stochastic direction offset. This ensures that the BSDF and NEE strategies evaluate the same environment
radiance.

The equirectangular CDF now uses `luminance * sin(theta)` so each texel's spherical solid angle is included in its sampling weight. This follows Cycles' `background_cdf` construction: [light.cpp](https://github.com/blender/blender/blob/main/intern/cycles/scene/light.cpp).

| Before | After |
| :---: | :---: |
| 95e431a893855faa8cd40ee819cad165 | 90b166e117c0b20dc9220edca246aa83 |
| f4c551085e4e8f97a379b93956395fcf | 1f0164cbadf8320cec3794054cdcdc7b |
| 9afea8e0c36e7941fee3d16fe3199c12 | 0d784cd0577a60e3c4151d96b565bf07 |

As for the firefly noise occurring at glossy boundaries and transmission areas on the model, the exact cause has not yet been identified.

### DIFF
--- a/src/uniforms/EquirectHdrInfoUniform.js
+++ b/src/uniforms/EquirectHdrInfoUniform.js
@@ -178,6 +178,12 @@ export class EquirectHdrInfoUniform {
 
 	}
 
+	getPixelWeight( r, g, b ) {
+
+		return colorToLuminance( r, g, b );
+
+	}
+
 	dispose() {
 
 		this.marginalWeights.dispose();
@@ -218,10 +224,7 @@ export class EquirectHdrInfoUniform {
 				const g = DataUtils.fromHalfFloat( data[ 4 * i + 1 ] );
 				const b = DataUtils.fromHalfFloat( data[ 4 * i + 2 ] );
 
-				// the probability of the pixel being selected in this row is the
-				// scale of the luminance relative to the rest of the pixels.
-				// TODO: this should also account for the solid angle of the pixel when sampling
-				const weight = colorToLuminance( r, g, b );
+				const weight = this.getPixelWeight( r, g, b, y, height );
 				cumulativeRowWeight += weight;
 				totalSumValue += weight;
 
--- a/src/uniforms/EquirectHdrInfoUniform.js
+++ b/src/uniforms/EquirectHdrInfoUniform.js
@@ -224,6 +224,8 @@ export class EquirectHdrInfoUniform {
 				const g = DataUtils.fromHalfFloat( data[ 4 * i + 1 ] );
 				const b = DataUtils.fromHalfFloat( data[ 4 * i + 2 ] );
 
+				// the probability of the pixel being selected in its row is proportional
+				// to its luminance relative to the other pixels.
 				const weight = this.getPixelWeight( r, g, b, y, height );
 				cumulativeRowWeight += weight;
 				totalSumValue += weight;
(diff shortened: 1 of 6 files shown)


src/uniforms/EquirectHdrInfoUniform.js
```diff
@@ -218,10 +224,7 @@ export class EquirectHdrInfoUniform {
 				const g = DataUtils.fromHalfFloat( data[ 4 * i + 1 ] );
 				const b = DataUtils.fromHalfFloat( data[ 4 * i + 2 ] );
 
-				// the probability of the pixel being selected in this row is the
-				// scale of the luminance relative to the rest of the pixels.
```
gkjohnson, line 222: Can we move this comment up to getPixelWeight instead of deleting it

src/webgpu/compute/wavefront/RayIntersectionKernel.js
gkjohnson: This is only used in the background branch, now, so we can move that down into the background sample code block. We should also rename this variable to something like `RNG_INDEX_BACKGROUND_SAMPLE`.

src/webgpu/EquirectHdrInfoNode.js
```diff
@@ -28,6 +28,13 @@ export class EquirectHdrInfoNode extends EquirectHdrInfoUniform {
 
 	}
 
+	getPixelWeight( r, g, b, row, height ) {
+
+		const theta = Math.PI * ( row + 0.5 ) / height;
+		return super.getPixelWeight( r, g, b ) * Math.sin( theta );
+
```
gkjohnson, line 47: Can we add a comment here indicating that this is weighting the pixel contribution by the solid angle.

src/webgpu/EquirectHdrInfoNode.js
```diff
@@ -68,10 +75,9 @@ export class EquirectHdrInfoNode extends EquirectHdrInfoUniform {
 		} = this;
 
 		this.sampleColor = wgslTagFn/* wgsl */`
-			fn sampleEnv( direction: vec3f, uv: vec2f ) -> vec4f {
+			fn sampleEnv( direction: vec3f ) -> vec4f {
 
-				let offsetDir = ${ sampleHemisphereFn }( direction, uv ) * 0.5;
-				let sampleDir = normalize( ${ rotationNode } * direction + offsetDir );
+				let sampleDir = ${ rotationNode } * direction;
```
gkjohnson, line 92: Thanks looks like this was left over after the env.blur field was removed.

src/webgpu/nodes/sampling.wgsl.js
```diff
@@ -178,6 +178,17 @@ export const equirectDirectionPdfFn = wgslFn( /* wgsl */ `
...
+export const equirectLuminancePdfFn = wgslFn( /* wgsl */ `
+
+	fn equirectLuminancePdf( luminance: f32, totalWeight: f32, resolution: vec2u ) -> f32 {
+
+		return f32( resolution.x * resolution.y ) * luminance / ( 2.0 * PI * PI * totalWeight );
+
+	}
+
+`, [ constants ] );
+
```
gkjohnson: This can just be bookkept in the EnvironmentHdrInfoNode file.

Ryan-G7:
I have resolved these.

gkjohnson:
Great, thanks!

gkjohnson:
> As for the firefly noise occurring at glossy boundaries and transmission areas on the model, the exact cause has not yet been identified.

This branch doesn't yet have filter glossy implementation, yet, so it's possible that will help address some of these fireflies if they're from diffuse-specular rays.

PR #816, opened 2026-08-13
WebGPU: Reorganize lobes into categorized as layers

gkjohnson:
Related to #777

I've asked Claude to organize the lobes so the logic per-lobe is more self contained so it's easier to understand, energy is maybe easier to track, and we can possibly expand to more layers in the future.

- Separate lobes into dedicated blocks
- Adjust transmission lobe sample and eval to handle reflection & transmission
- Remove pre-calculate clearcoat values and compute them as-needed
- Calculate iridescence in the specular / transmission branch

**NEXT**
- Remove unused disney functions
- Consider splitting up specular and metallic lobes into separate branches


gkjohnson:
cc @ryan-G7 - just FYI that the changes in this branch may make dispersion or eon contributions conflict.

PR #822, opened 2026-08-17
Add EON rough diffuse BRDF

Ryan-G7:
Adds EON, an energy-preserving rough diffuse BRDF.

### Implemented

- Add the FON single-scattering diffuse BRDF.
- Add the energy-compensated EON diffuse BRDF.
- Added `material.diffuseRoughness`, stored in the existing material buffer.
- Add CLTC importance sampling for EON.
- Add the corresponding EON mixture PDF.

97bc55be782443196bffe2556d0cdc99

<br>

| Lambert | FON | EON |
| :---: | :---: | :---: |
| [Lambert](https://github.com/user-attachments/assets/5e0da9ee-2ac9-433e-8148-d391c83a02e5) | [FON](https://github.com/user-attachments/assets/6b3e3683-aa41-490c-89ce-681134e12251) | [EON](https://github.com/user-attachments/assets/1c2f45e5-dd82-4e19-b17b-068771fa7d7e) |

### Furnace Test
| FON | EON |
| :--- | :--- |
| FON | EON |

### References

- [EON: A Practical Energy-Preserving Rough Diffuse BRDF](https://jcgt.org/published/0014/01/06/)
- [Blender Cycles Oren-Nayar Implementation](https://github.com/blender/blender/blob/main/intern/cycles/kernel/closure/bsdf_oren_nayar.h)
- [OpenPBR Specification](https://academysoftwarefoundation.github.io/OpenPBR/)

### Shading-normal Correction

Reflecting a sampled direction back above the geometric surface changes the effective sampling distribution. The existing PDF still described the original direction and did not include the folded probability mass or its transform.

The half vector was also computed before the direction adjustment, so subsequent BSDF and PDF evaluation would use a half vector that no longer matched the sampled direction.


gkjohnson:
Awesome! Looks great - I'll take a deeper look this weekend.

Would you be able to put together a set of 4-5 screenshots of a reasonably complex model with a surface color so the color and intensity of the reflection as the variable changes are clear? As far as I understand setting "diffuseRoughness" to 0 means the EON model will match lambertian exactly right? I'm thinking about proposing adding this model to three.js' MeshPhysicalMaterial

Ryan-G7:
I've updated the PR description to show the comparison between different diffuse roughness values and Lambert. 

>  I'm thinking about proposing adding this model to three.js' MeshPhysicalMaterial

I also agree with proposing this model to Three.js so that users have the flexibility to adjust diffuse roughness.

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -368,7 +370,8 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 				} else if ( lobeSample <= cdfTotal ) {
 
 					// diffuse
-					wi = ${ diffuseDirectionFunc }( wo, directionUV );
+					// all diffuse BRDF variants share the EON sampling proposal
+					wi = ${ eonDirectionFunc }( wo, surf.diffuseRoughness, directionUV );
```
gkjohnson, line 374: This direction function needs to match the diffuseBrdf function - so I think we either need to lean in on the user-defineable functions and make sure everything necessary can be defined symmetrically or defer on that. I'm thinking we defer that until we start to look into work for things like custom layers / BRDFs down the line. Same for the PDF line below

We don't have to make any changes here in that case - just something that came to mind.

src/webgpu/nodes/material.wgsl.js
```diff
@@ -387,6 +387,7 @@ export const getSurfaceRecordFunc = ( sampleTexel, getUvFromChannel, getColor )
 		surf.normal = normal;
 
 		surf.metalness = metalness;
+		surf.diffuseRoughness = saturate( material.diffuseRoughness );
```
gkjohnson, line 390: Why call saturate here? We're not doing so for any other fields

gkjohnson (commented):
Thanks! A couple small comments - but looks good. Would you like to make an issue in three.js to discuss adding EON support to three.js' MeshPhysicalMaterial? I'd been talking about mrdoob about it earlier in the week

src/webgpu/nodes/material.wgsl.js
Ryan-G7: specularRoughness uses clamp to prevent divide-by-zero singularities in the GGX denominators, whereas diffuseRoughness is mathematically safe even when it is strictly 0.0 in the diffuse BRDF.

Therefore, only need to ensure it stays within the valid physical range [0, 1].

Ryan-G7:
> Would you like to make an issue in three.js to discuss adding EON support to three.js' MeshPhysicalMaterial? I'd been talking about mrdoob about it earlier in the week

Sounds good, I'll take care of it this weekend.

src/webgpu/nodes/material.wgsl.js
gkjohnson: > Therefore, only need to ensure it stays within the valid physical range [0, 1].

I understand the intended range for the model is [0, 1] - but that' the case for a lot of these fields. Metalness, transmission, etc will all fall apart in a way if we use values outside 0, 1 and we're not bounding. So in part I'm asking "why here and not there". Unless there's a different reason I'll remove these bounds for consistency

Ryan-G7: I think it's reasonable to remove these boundaries to maintain consistency across the codebase.

PR #825, opened 2026-08-20
Add configurable direct and indirect path-contribution clamping

Ryan-G7:
## Description

I retested the firefly noise issue using the mis branch, which includes the merged filterGlossy code, and found that the noise was not eliminated. I then added clamping for both direct and indirect light; however, I noticed that this only reduced the brightness of the noise rather than removing it completely.

I submitted these changes as a mitigation for the firefly noise, and I will leave it to your discretion whether to merge this PR.

## Changes

- Add `clampDirect` and `clampIndirect` settings.

- Propagate the settings through both WaveFront and MegaKernel backends.

- Clamp contributions using `abs(R) + abs(G) + abs(B)`.

- Normalize limits by `3.0`, matching Cycles' user-facing threshold semantics.

- Preserve chromaticity by uniformly scaling RGB contributions.

- Apply clamping to environment lighting, BSDF environment misses, and emissive path contributions.

- Keep clamping disabled by default.

- Enable `clampIndirect = 10` in the WebGPU viewer example.

## Trade-offs

Contribution clamping introduces bias and may suppress very bright caustics or legitimate high-energy indirect lighting. Set the limits to `0` to disable clamping.

image

### DIFF
--- a/example/viewer.js
+++ b/example/viewer.js
@@ -45,8 +45,6 @@ async function init() {
 
 	// path tracer
 	pathTracer = new WebGPUPathTracer( renderer );
-	pathTracer.filterGlossyFactor = 0.5;
-	pathTracer.clampIndirect = 10;
 	pathTracer.renderScale = renderScale;
 	pathTracer.tiles.set( tiles, tiles );
 
--- a/example/viewer.js
+++ b/example/viewer.js
@@ -46,6 +46,7 @@ async function init() {
 	// path tracer
 	pathTracer = new WebGPUPathTracer( renderer );
 	pathTracer.filterGlossyFactor = 0.5;
+	pathTracer.clampIndirect = 10;
 	pathTracer.renderScale = renderScale;
 	pathTracer.tiles.set( tiles, tiles );
 
--- a/src/webgpu/compute/wavefront/LightConnectionKernel.js
+++ b/src/webgpu/compute/wavefront/LightConnectionKernel.js
@@ -4,7 +4,7 @@ import { uniform, storage, globalId, texture, sampler } from 'three/tsl';
 import { hitQueueStruct } from './structs.js';
 import { proxy, proxyFn, wgslTagFn, rayStruct } from 'three-mesh-bvh/webgpu';
 import { misHeuristicFn } from '../../nodes/sampling.wgsl.js';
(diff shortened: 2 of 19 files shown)


example/viewer.js
```diff
@@ -46,6 +46,7 @@ async function init() {
 	// path tracer
 	pathTracer = new WebGPUPathTracer( renderer );
 	pathTracer.filterGlossyFactor = 0.5;
+	pathTracer.clampIndirect = 10;
```
gkjohnson: Lets leave these at the default values defined in the renderer (you can change the filter glossy factor, too). Unless there's a reason to adjust them?

src/webgpu/compute/wavefront/LightConnectionKernel.js
```diff
@@ -20,6 +20,8 @@ export class LightConnectionKernel extends ComputeKernel {
 			// settings
 			misEnabled: uniform( 1, 'uint' ),
 			filterGlossy: uniform( 1 ),
+			clampDirect: uniform( 0 ),
+			clampIndirect: uniform( 0 ),
```
gkjohnson: Does Blender differentiate between these two?

gkjohnson (commented):
I don't love clamping the samples but it looks like cycles does this, too 😅 I assume blenders defaults are also 0 here?

Ryan-G7:
> I assume blenders defaults are also 0 here?

Actually, Blender's defaults are not both 0. Cycles ships with `clamp_direct = 0` but `clamp_indirect = 10` to suppress fireflies from indirect paths.

src/webgpu/compute/wavefront/LightConnectionKernel.js
Ryan-G7: Yes. Cycles exposes two separate settings.

gkjohnson:
> Cycles ships with clamp_direct = 0 but clamp_indirect = 10 to suppress fireflies from indirect paths.

I'm wondering if we should just match Blender's defaults, then? What do you think?

Ryan-G7:
> I'm wondering if we should just match Blender's defaults, then? What do you think?

I agree, matching Blender's defaults makes the most sense. I've already updated the PR to align with them. Let me know if you need any other adjustments!

Issue #826, opened 2026-08-21
WebGPU: texture atlas repacks on every `updateMaterials()`, even when no texture changed


MichaMaGit:
## Describe the bug

`AtlasTexture` compares per-texture hashes (`source.uuid + ':' + source.version`, **positionally**)
to decide whether to repack. During live editing both halves of that key change on their own, so the
check fails almost every call and the atlas repacks everything. With 9 maps that is **760–805 ms per
`updateMaterials()`**, all inside `setTextures`.

**1. The atlas dirties the shared source.** `_renderTextures()` clones each texture and sets
`needsUpdate = true` on the clone. `clone()` shares the `source`, and `needsUpdate` bumps
`source.version` — the value the hash is built from. So packing invalidates the atlas's own hashes:
pack → hash changed → pack again. Measured: `source.version` +2 per tick, no image changed.

**2. Material order is random.** `updateMaterialsMap()` sorts with
`materials.sort( ( a, b ) => a.uuid < b.uuid ? 1 : - 1 )`. Textures are collected in material order,
so an app that assigns a *new* material instance per change (what our editor does) gets a new uuid,
the material jumps to another slot, and the positional comparison fails on an identical set.

Cause 1 hides cause 2 — fixing only the version keeps the repacks, just less often.

### How to reproduce

Not reduced to a standalone example yet; measured in an application. Minimal shape: two meshes with
**two different textured materials** (one is not enough — with a single one the order cannot change),
then in a loop assign a new material with the same maps to one mesh and call `updateMaterials()`.
Wrapping `textureAtlas.setTextures` shows most calls repacking; with both fixes, none do.

Measured over 40 ticks (2 materials, 6 maps): **7 repacks** with the sort, **0** without, and the
rendered image is identical (mean luminance 91.63 vs 91.64, red bias −5.16 vs −5.16). Every repack
we classified was "same set, different order", never a different set.

### Possible fix

**1 — keep the shared source clean across the blit** in `_renderTextures()`:

```js
const prevVersions = textures.map( ( t ) => t.source.version );

// ...existing clone / needsUpdate / blit / dispose loop, unchanged...

for ( let i = 0, l = textures.length; i < l; i ++ ) {

	textures[ i ].source.version = prevVersions[ i ];

}
```

The clone is still uploaded — that upload is gated on the clone's own `texture.version`, not on
`source.version`. Only `getTextureHash()` reads `source.version`.

[...]

PR #830, opened 2026-08-23
WebGPU: stop repacking the texture atlas when only the order changed

m-w-marker:
Fixes the repacking reported in #826.

**Stable pack order.** `AtlasTexture` compares the texture hashes by position, and the order it
gets them in comes from material order, which changes by itself: `updateMaterialsMap` sorts the
materials by uuid, and our editor assigns a new material instance per edit, so the list gets
reshuffled. The atlas then repacks an identical set of textures, 760 to 805 ms with 9 maps. This
sorts the textures by `source.uuid` before packing and maps the placements back to the caller's
indices, so `textureInfo` stays indexed the way the shader expects. I put the sort inside
`AtlasTexture` rather than at the call site, because by then the texture indices are already
packed into the material buffer and sorting there would mean remapping all of them.

**One redundant line.** `_renderTextures` sets `needsUpdate` on the clone. The clone shares its
source, so this bumps `source.version` on a texture the atlas only reads. The blit uploads the
clone either way. It is a separate commit, so feel free to drop it if you want to keep the flag.

Checked on a real GPU (Edge, WebGPU, r185): three textures, `setTextures` called five times, each
time in a different order, reading the packed pixels back through `textureInfo`. Before: 5 repacks
out of 5 calls. After: 1. Pixels correct in both. A sixth call with an added texture still repacks.

### DIFF
--- a/src/webgpu/AtlasTexture.js
+++ b/src/webgpu/AtlasTexture.js
@@ -218,6 +218,7 @@ export class AtlasTexture {
 		this.textureInfo = [ new Vector4() ];
 
 		this.hashes = [];
+		this.placements = [];
 		this.quadMesh = new QuadMesh( new MeshBasicMaterial() );
 		this.quadMesh.material.blending = NoBlending;
 
@@ -225,9 +226,17 @@ export class AtlasTexture {
 
 	setTextures( renderer, textures ) {
 
+		// The order the textures arrive in follows material order, which can change without the
+		// set of textures changing. Pack in a uuid-stable order instead, and map the placements
+		// back to the caller's indices afterwards.
+		const order = textures.map( ( t, i ) => i );
+		order.sort( ( a, b ) => textures[ a ].source.uuid < textures[ b ].source.uuid ? - 1 : 1 );
+		textures = order.map( i => textures[ i ] );
+
 		// nothing to do if the set of textures is unchanged
 		if ( this._hashesMatch( textures ) ) {
 
+			this._buildCallerTextureInfo( order );
 			return;
 
 		}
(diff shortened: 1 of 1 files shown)


gkjohnson:
Great, thanks! I'm thinking that to keep things simple we can just remove the call to "needsUpdate" and call `textures.sort( ( a, b ) => a.uuid < b.uuid ? 1 : - 1 )` in WebGPUPathTracer before calling "setTextures". Then we can avoid remapping everything in the AtlasTexture.

MichaMaGit:
Dropping the `needsUpdate` line is fine.

Sorting at the call site breaks the material buffer though. `getTexture()` in
`writeMaterialsBuffer` hands out texture indices in discovery order and packs them into the low 23
bits of the material word, and `_bvhData.textures` is the array those indices point into.
`updateMaterials()` builds that buffer before it calls `setTextures` (same order in `setScene`), so
reordering the array in between makes materials sample the wrong map.

`uuid` is also the wrong key. Dedup is by `getTextureHash` = `${source.uuid}:${colorSpace}`, and
`texture.uuid` changes whenever a new Texture wraps the same source, which is exactly the case that
causes the repack here.

If you want the atlas free of remapping, the sort can move ahead of the indices: collect the
textures in `writeMaterialsBuffer`, sort by `source.uuid`, then assign indices from that order. I
can redo the PR that way if you prefer.

Worth noting if you take that route: the indices are handed out lazily across 18 `getTexture()`
calls, so a pre-pass needs its own list of the map slot names. That list silently goes stale when
a new slot is added, and the failure mode is invisible (correct image, repacking is back).

gkjohnson:
Thanks you're right - that makes sense.

PR #844, opened 2026-09-03
Add dispersion support to the WebGPU path tracer

Ryan-G7:
## Description

This PR adds `KHR_materials_dispersion` support to the WebGPU path tracer for both the Megakernel and Wavefront implementations. Also fixes NaN issues related to IOR.

- Packs the material `dispersion` property into the WebGPU material buffer.
- Samples one RGB hero channel at the first dispersive interaction.
- Preserves the selected channel across subsequent bounces.
- Applies the same dispersed surface parameters during BSDF sampling and NEE.
- Supports both front-face and back-face dielectric transitions.
- Handles matched IOR boundaries as straight-through delta transmission, preventing undefined rough-BTDF operations and NaNs.
- Preserves the BSDF PDF used for transmissive environment MIS.

This is an RGB spectral approximation rather than continuous spectral transport.

## Cauchy IOR approximation

`KHR_materials_dispersion` stores dispersion as `20 / Vd`. The channel IORs are evaluated at the Fraunhofer C, d, and F wavelengths:

- Red: 656.27 nm
- Green: 587.56 nm
- Blue: 486.13 nm

The implementation uses:

```text
n(lambda) = max(
  nd + (nd - 1) * dispersion / 20
    * (523655 / lambda^2 - 1.5168),
  1
)
```

The green channel remains at nd, matching the reference slang-pbr implementation used by dspbr-pt.

## Sampling

A path selects one RGB hero channel uniformly when it first reaches a dispersive dielectric. Its throughput is multiplied by three to account for the 1/3 selection probability. The selected channel is then carried through the rest of the path, reducing inconsistent wavelength changes across internal reflections and transmission events.

## Matched IOR

When the incident and transmitted IORs match, the rough dielectric BTDF half-vector and Jacobian become undefined. These boundaries are handled as unit-energy straight-through delta transmission, following PBRT-v4.

## References

- [KHR_materials_dispersion](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_dispersion)
- [slang-pbr Cauchy implementation](https://github.com/bsdorra/slang-pbr/blob/c12046ec3b9628bcfc384ed2c275c92d007e5d7e/src/models/enterprise/components/transmission.slang#L20-L27)
- [PBRT-v4 dielectric material](https://pbr-book.org/4ed/Textures_and_Materials/Material_Interface_and_Implementations.html#sec:dielectric-material)
- [Hero Wavelength Spectral Sampling](https://doi.org/10.1111/cgf.12419)

| This PR | Dspbr-pt |
| :---: | :---: |
| This PR | Dspbr-pt |
| **BabylonJS** | **Adobe Substance 3D Stager** |
| BabylonJS | Adobe Substance 3D Stager |


gkjohnson:
Thanks! This is great - I'll try to take a deeper look at this next week, but one thing I'm wondering is whether we can use a more fluid spectrum value for the dispersion channel rather than just RGB? Then we could avoid this type of dispersion banding and match something closer to what the adobe version has:

image

Is it significantly more code or complexity?

Also FYI the"viewerTest" example include some new test models for dispersion if you'd like to test use those for testing too:

- x-DragonDispersion
- x-CompareDispersion
- x-DispersionTest

Ryan-G7:
Thanks for the suggestion! I've updated the implementation to sample a continuous hero wavelength uniformly over 360–830 nm instead of selecting one of three RGB channels. Since Adobe is closed-source, I have no way of knowing its exact implementation, but I suspect it is similar to our current approach.

* Evaluated $n(\lambda)$ directly via the Cauchy equation, rather than relying on three discrete R/G/B IOR values.
* Converted the wavelength into linear RGB reconstruction weights using the Wyman et al. analytic CIE 1931 fit.
* Normalized the weights per channel to preserve the original RGB renderer's expected throughput and energy balance.
* Replaced the `uint` channel in the wavefront queue with an identically sized `float` wavelength, maintaining the existing buffer stride.

This remains an RGB-compatible spectral approximation rather than a fully spectral renderer, material, texture, light, and attenuation data remain in RGB format. However, it eliminates the fixed three-band behavior and produces smooth, continuous hue transitions.

Implementing a full spectral renderer would require:

* Spectral up-sampling to convert RGB textures into spectral reflectances.
* Spectral representations for light sources and environment maps.
* Wavelength-dependent absorption, emission, and transmission across the entire BSDF/media pipeline.
* Spectral consistency across MIS, Russian roulette, and denoising passes.
* Late sensor reconstruction into CIE XYZ / target RGB color space.

This would require substantial architectural changes across the rendering engine.

## Updated Results

| `x-DragonDispersion` | `x-DispersionTest` | `x-CompareDispersion` |
| :---: | :---: | :---: |
| x-DragonDispersion | x-DispersionTest | x-CompareDispersion |

src/webgpu/compute/wavefront/ProcessHitsKernel.js
```diff
@@ -119,13 +119,29 @@ export class ProcessHitsKernel extends ComputeKernel {
...
-				let scatterRec = ${ bsdfSampleFn }( input.view, surface );
-
 				var throughputColor = input.throughputColor;
+				var dispersionWavelength = input.dispersionWavelength;
+
+				let isDispersive = materialInfo.dispersion > 0.0 && surface.ior > 1.0 && surface.transmission > 0.0 && ! surface.thinWall;
+				if ( isDispersive ) {
+
+					let wavelength = abs( dispersionWavelength );
+					surface = ${ applyDispersionFunc }( surface, materialInfo.dispersion, wavelength );
```
gkjohnson, line 134: Just a thought - does it make sense to inline the "applyDispersionFunc" in the "getSurfaceRecordFn"? It would mean passing the wavelength in to the function but I'm not sure if it's a better approach.

Either way - lets change this "applyDispersionFunc" to take a reference via pointer to the "surface" object so it can modify it in place rather than creating new full copies of the structure.

src/webgpu/nodes/utils.wgsl.js
```diff
@@ -197,7 +197,8 @@ export const isTerminatingScatterFunc = wgslFn( /* wgsl */ `
 
 	fn isTerminatingScatter( scatterRec: ScatterRecord ) -> bool {
 
-		return scatterRec.pdf <= 0;
+		// A comparison with NaN is false, so this also rejects invalid PDFs.
+		return ! ( scatterRec.pdf > 0 );
```
gkjohnson, line 201: Can you explain what this was for? Is this a general change? Or something necessitated by the changes in this PR? Are we seeing NaNs? I'm somewhat worried that this will be masking broken code paths rather than fixing any issues we're seeing.

src/webgpu/materials/GltfCompliantMaterial.js
```diff
@@ -105,6 +105,8 @@ export class GltfCompliantMaterial extends PathtracingMaterial {
 				// handles specular reflection and transmission
 				if ( surf.transmission > 0.0 ) {
 
+					let matchedIor = ! surf.thinWall && ${ isMatchedIorFunc }( surf.eta );
```
gkjohnson, line 108: Is it right that we don't need to guard thinWall here? It seems like it should be susceptible to the same issues. This seems to somewhat fix some of the remaining viewerTest comparison models, but not completely. Here's what I'm seeing specifically. Do you mind looking at these models and figuring out what's going on?

| model | three.js | path tracer |
|---|---|---|
| x-TransmissionThinwallTestGrid | Image | Image |
| x-SunglassesKhronos | Image | Image |
| x-IORTestGrid | Image | Image |

I think we should at least get the "TransmissionThinwallTestGrid" case fixed here. The last two I think may be a bit more complicated and possibly related to how three.js is handling specular at ior = 1.0, which is not physically correct? So I think there's a question of do we want to do what's "correct" or match three.js / glTF more exactly. That's a question for a different PR, I suppose, but I'm interested to hear what you think.

gkjohnson (commented):
This is looking great! I made a few comments above - also note I made a fix for transparent backgroudn rendering in the view test demo on the main webgpu branch

src/webgpu/materials/GltfCompliantMaterial.js
gkjohnson: I recall that the "soap bubble material" from the pbr database is exhibiting some differences, as well, possibly related to the same thing we're seeing with the sunglasses above. This is a difference with Blender, though:

| path tracer | pbr database reference |
|---|---|
| image | image |

gkjohnson:
Just realized there were a bunch of conflicts now from the wavefront branch 🤦 Sorry about that. I can help resolve them once the PR is all ready if you'd like.

src/webgpu/compute/wavefront/ProcessHitsKernel.js
Ryan-G7: Good point. I've changed `applyDispersion` to take a `ptr<function, SurfaceRecord>`

src/webgpu/nodes/utils.wgsl.js
Ryan-G7: Agreed. This was added as a defensive check while I was tracking down the matched-IOR NaN. But, I noticed that after reverting this code, a few NaN black pixels appeared in x-CompareDispersion. This seems to happen because floating-point rounding causes the reflection direction wi to become exactly equal to -wo.

src/webgpu/materials/GltfCompliantMaterial.js
Ryan-G7: Exactly. When eta == 1, the roughness evaluates to 0, which hits 0 / 0 in GGX and makes the PDF NaN.

Since matched-IOR means no refractive boundary, thin-wall transmission should just collapse to straight-through delta transmission. Sampling now prioritizes matched-IOR over the thin-wall GGX path, returning 0 for the solid-angle density along that delta direction.

Ryan-G7: > I think we should at least get the "TransmissionThinwallTestGrid" case fixed here. The last two I think may be a bit more complicated and possibly related to how three.js is handling specular at ior = 1.0, which is not physically correct? So I think there's a question of do we want to do what's "correct" or match three.js / glTF more exactly. That's a question for a different PR, I suppose, but I'm interested to hear what you think.

 I checked them separately after fixing the matched-IOR thin-wall NaN.

For `x-SunglassesKhronos`, the outer lens uses IOR 1 and transmission 1, but it also uses `KHR_materials_iridescence` with a film IOR of 2 and a thickness of 300.01 nm. Although the substrate media are IOR-matched, the thin film can still reflect light. The current matched-IOR delta path bypasses the iridescent Fresnel term, which explains why the colored lenses disappear. Fixing that correctly requires iridescence-aware reflection/transmission sampling probabilities and matching PDF/MIS handling, so I agree that this should be handled in a follow-up PR.

For `x-IORTestGrid`, some IOR 1 materials also use `KHR_materials_specular`. Three.js produces an F0 of zero for those materials but still uses a nonzero Schlick F90, so they can retain grazing-angle reflections even at IOR 1. Fully transmissive, matched-IOR cells should still be straight-through delta events.

gkjohnson:
I just resolved the merge conflicts - I'll try to take a look tomorrow. Some of the images rendered for testing:

image

image

image

src/webgpu/nodes/utils.wgsl.js
gkjohnson: So to be clear - are the NaN pixels addressed now in the compare dispersion demo? I think if we do explicit NaN checks we should do it in a single spot before blend / write so the check is in one spot and allow it to be toggled for the sake of testing.

src/webgpu/materials/GltfCompliantMaterial.js
gkjohnson: > I agree that this should be handled in a follow-up PR.

If you're able to take a look at this after merging that would be great.

> For x-IORTestGrid, some IOR 1 materials also use KHR_materials_specular. Three.js produces an F0 of zero for those materials but still uses a nonzero Schlick F90, so they can retain grazing-angle reflections even at IOR 1. Fully transmissive, matched-IOR cells should still be straight-through delta events.

Yeah - it seems three.js (and glTF) aren't using a physically correct model here. I'm inclined to say we should just use the model we have right now and decide what to do if someone complains.

gkjohnson:
Thanks! This is a great addition

PR #846, opened 2026-09-10
Support iridescence on matched-IOR transmission

Ryan-G7:
## Description

Related to #844. Fixes the disappearing colored lenses on x-SunglassesKhronos: the outer lens uses IOR 1 with transmission 1 plus KHR_materials_iridescence (film IOR 2, thickness ~300 nm). The substrate media are IOR-matched, but the thin film can still reflect light - and the matched-IOR delta path bypassed the iridescent Fresnel term entirely.

This PR makes the transmissive dielectric layer iridescence-aware with matching reflection/transmission sampling probabilities and PDF/MIS handling:

- Adds a shared `transmissionFresnel` helper returning colored reflectance / transmittance coefficients plus a scalar branch probability derived from their mean energies, following Cycles' glass closure.
- Carries the thin-film transmittance on the matched-IOR delta transmission event, weighted by the matching branch probability, so iridescent lenses keep their color instead of vanishing.
- Represents the delta event as a singular event with BSDF and PDF both scaled by 1e6 (Cycles convention): the ratio preserves throughput while forward MIS correctly assigns full weight to the event.
- Uses the same coefficients consistently across BSDF sampling, evaluation, and the PDF so the light-sampling path can never disagree with the BSDF path.
- Keeps index-matched media fully transparent when no film is present, avoiding the non-physical Schlick F90 reflection that KHR_materials_specular would otherwise introduce on a substrate with no IOR boundary.

## References

- PBRT-v4 dielectric material (eta == 1 delta path)
- Blender Cycles glass closure (energy-based reflection probability, singular events)
- Belcour / Barla thin-film interference model

| Before | This PR |
| --- | --- |
| image | image |

### DIFF
--- a/src/webgpu/nodes/material.wgsl.js
+++ b/src/webgpu/nodes/material.wgsl.js
@@ -20,7 +20,7 @@ import {
 	ggxRefractionAdjustedPDFFunc,
 	ggxShadowMaskG1Func,
 } from './ggx.wgsl.js';
-import { constants, surfaceRecordStruct } from './structs.wgsl.js';
+import { constants, surfaceRecordStruct, transmissionFresnelStruct } from './structs.wgsl.js';
 import { wgslTagFn } from 'three-mesh-bvh/webgpu';
 
 // Correct the shading normal to prevent scattering rays below the geometry surface by bending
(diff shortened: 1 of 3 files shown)


gkjohnson:
I still need to do a full review but this is great - for completeness, here are the remaining screenshots that had issues in #844:

| Soap Bubble | Sun Glasses | TransmissionThinwallTestGrid |
|---|---|---|
|  |  |  |

A couple things I'm seeing - in the soap bubble case I'm seeing that the shadows are less strong (or non existent) compared to the cycles version. Not sure if you have any thoughts on this? It might be a an energy bookkeeping issue? If it's complicated to fix (I expect it might be) we can leave it but it would be good to understand if possible.

| Path Traced (this PR) | PBRDB Reference |
|---|---|
|  |  |

The last thing I'm seeing is some bright artifacts of the metal clips showing through the lenses in the sunglasses scene. This may also just be an unavoidable fireflies artifact?

image

AntonPalmqvist:
I take it that you gathered the Cycles reference from [Physically Based](https://physicallybased.info/)?
It's either my render settings, or Cycles just not being entirely accurate, but I'd say the Cycles shadow is too strong.
PBRDB looks more correct to me.

gkjohnson:
> I take it that you gathered the Cycles reference from [Physically Based](https://physicallybased.info/)?
> It's either my render settings, or Cycles just not being entirely accurate, but I'd say the Cycles shadow is too strong.
PBRDB looks more correct to me.

Sorry - I had the labels swapped. The PBRDB Reference is the one with the stronger shadows. I assumed that was Cycles? But maybe it's not. So this path tracer is the one with LESS of a shadow.

I also thought the PBRDB shadows looked a bit strong but didn't know if it was more correct or not.

AntonPalmqvist:
No, I'm sorry, my brain didn't register the swapped labels at all!
Just to clarify, here's my Cycles render from Physically Based, where the shadow is too strong:


The render for this PR does indeed seem more correct!

gkjohnson:
@Ryan-G7 - I was going to take a look at this tomorrow to review. Is there a reason that the last commit that changed the irridesence model was merged into this branch instead of being a separate PR? I think it would be easier to review as a separate change. You can make a "PR Stack" so you can avoid being blocked in the mean time.

Sorry for taking a bit to get to the review

src/webgpu/nodes/material.wgsl.js
```diff
@@ -791,6 +791,73 @@ export const iridescentFresnelFunc = wgslFn( /* wgsl */ `
...
+				vec3f( dielectricF ),
+				${ schlickFresnelVecFunc }( cosine, dielectricF0, vec3f( 1.0 ) ),
+				airIncident,
+			);
+			reflectance = surf.specularIntensity * dielectricFr;
+
+		}
+
+		reflectance = clamp( reflectance, vec3f( 0.0 ), vec3f( 1.0 ) );
+		var transmittance = vec3f( 1.0 - dielectricF );
```
gkjohnson, line 824: The one question I have is whether this line is right - should it be `1.0 - reflectance`? Or `1.0 - dielectricFr` (the one in the `! matchedIor` branch separate from `dielectricF`)? Relatedly, the `dielectricF` and `dielectricFr` naming is a bit unclear - I'm wondering if it's worth renaming for clarity?

Ryan-G7: Yes — this should be 1.0 - reflectance.

dielectricF is only the scalar Fresnel term derived from the IOR, so it does not include the KHR_materials_specular color or intensity adjustments. dielectricFr includes the directional dielectric reflectance, but is still before specularIntensity is applied. Therefore neither represents the complete coefficient used by the reflection branch.

Using the complement of the final reflectance keeps reflection and transmission consistent. I also renamed dielectricF, dielectricFr, and dielectricF0 to exactDielectricFresnel, dielectricReflectance, and specularF0 respectively to clarify their roles.

gkjohnson: Great, thanks!

PR #853, opened 2026-09-17
Improve iridescence model with exact polarized Fresnel

Ryan-G7:
Related #846 

Thanks for the heads-up!

The reason for adding this PR was that while comparing our results against Blender Cycles, I noticed our iridescence effect was indeed noticeably weaker and had color shifts. After digging into it, I found this was partly due to a few physical and mathematical issues in the current thin-film implementation:

* The original Schlick approximation underestimated the internal interface reflectance (especially when exiting from the thin film with IOR 1.4 back into air at glancing angles).
* The XYZ-to-RGB conversion matrix generated via TSL nodes was transposed, which skewed the chromaticity of the interference fringes.

Following an approach similar to Cycles, I updated it to use exact $s$ / $p$ polarized Fresnel calculations alongside a 3rd-order Airy interference model. 

**Left: Before | Right: After**

01-dark-opaque-iridescence-before-after

01-opaque-iridescence-before-after

02-matched-ior-soap-bubble-before-after

03-volume-glass-iridescence-before-after

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,14 +29,14 @@
         "rollup": "^2.70.0",
         "simple-git": "^3.10.0",
         "three": "^0.185.1",
-        "three-mesh-bvh": "^0.9.13",
+        "three-mesh-bvh": "^0.9.15",
         "typescript": "^5.9.2",
         "vite": "^8.2.2",
         "yargs": "^17.5.1"
       },
       "peerDependencies": {
         "three": ">=0.180.0",
-        "three-mesh-bvh": ">=0.9.9",
+        "three-mesh-bvh": ">=0.9.15",
         "xatlas-web": "^0.1.0"
       }
     },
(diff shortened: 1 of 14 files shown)


gkjohnson:
Thanks! I'll take a look at this once #846 is merged. Would you be able to put together some side-by-side comparison screenshots to show the impact of this change?

--

Also, once this PR is merged I'm going to try to do a final sweep of things to get the webgpu branch merged into main and an initial version published. In case you're interested in some other features or improvements to look into, here's what I think would be good additions (in addition to others in #777). Just suggestions, though - I'm open to anything you might be interested in:

- Subsurface scattering (and fog)
- Support for "compressed wide BVH" in three-mesh-bvh to improve tracing performance
- Support for compute-side BVH generation (there's an initial implementation in https://github.com/gkjohnson/three-mesh-bvh/pull/853 but it would be good to have some thing that aligns with the project patterns and memory layouts)

Ryan-G7:
> * Support for "compressed wide BVH" in three-mesh-bvh to improve tracing performance

Sounds great! I'm definitely interested in looking into the compressed wide BVH support after this PR lands.

